PR #3, opened 2012-01-22
Fix for VIM-167

marquiswang:
This commit fixes the bug described in: http://youtrack.jetbrains.net/issue/VIM-167?projectKey=VIM


AlexPl292:
Hi @marquiswang ! Could you please take a look at this discussion so we can use your change under the new MIT license? Thank you!

marquiswang:
Hi @AlexPl292, sorry for the delay, I have left a comment with an approval.

PR #8, opened 2012-09-20
Bug fix for gi behaviour

poxu:
gi shouldn't work like it does now (at least according to the documentation)
Bug description

Steps to reproduce
1. Open a file
2. Switch to insert mode
3. Insert any characters
4. Switch back to normal mode
5. Press gg (or move cursor to another position by any other means)
6. Press gi

Expected result:
Switch to insert mode with cursor after the last inserted character

Actual result:
Switch to insert mode with cursor before the first non space character in the line


AlexPl292:
Hi @poxu ! Could you please take a look at this discussion so we can use your change under the new MIT license? Thank you!

PR #12, opened 2013-01-08
VIM-92: Fix Vim keymap to work better in Mac OS

zolotov:
The problem is not completely solved. Please, review approach that I selected as a solution. 

In order to completely fix this I need Vim.xml generated by vim-plugin (with my changes) on Linux and Windows. Could you please attach it to this pull request (if it possible) or email me to alexander.zolotov@jetbrains.com? I want to compare it with old static Vim.xml. Most likely I will have to support custom shortcut-overrides for Linux and Windows default keymaps.

### DIFF
--- a/resources/Vim.xml
+++ b/resources/Vim.xml
@@ -50,5 +50,30 @@
     <keyboard-shortcut first-keystroke="shift RIGHT" />
     <keyboard-shortcut first-keystroke="shift UP" />
   </action>
+
+
+  <!--
+  Following actions would be added in result keymap if action corresponding will be completely override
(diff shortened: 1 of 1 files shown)


jucor:
Hi @zolotov , @vlasovskikh : I'm running in the same problem with "Ctrl-R" being hijacked by IdeaVim but vital in Mac OS X 10.5+ keymap. What's the current fix, please? I can't find it, as Ctrl-Alt-R discussed above is also used by the keymap.

vlasovskikh:
@jucor See Settings | Other | Vim Emulation.

jucor:
Thank you very much!! This is perfect indeed.

PR #25, opened 2014-03-09
 VIM-410 - Adding support for ':set rnu' and ':set nu'

jjsanchez:
Pull request to add a nice feature that I was missing on IdeaVim: relative line numbers.

By setting the relativenumber option, the editor shows the line numbers relative to the caret position, allowing users to quickly jump a few lines up or down to change some text or maybe yank it.

screen shot 2014-03-09 at 11 08 28

With this feature you can just go 4j to go 4 lines down or 3k to go 3 lines up at a glance :)


rohitkrishna094:
I am not sure where to ask this since there is no issue tab on this repo. For ideavim, is it possible to have both absolute line numbers and relative line numbers? I have this setup in vscode and was wondering if it was possible to do this for intellij as well.

I have added this property in .vimrc

`set rnu`

I can get relative line numbers to work, but is there a way to add another column in the line gutter where we also have absolute line numbers? Thanks

AlexPl292:
The question is answered here: https://stackoverflow.com/questions/61681596/how-to-get-both-absolute-and-relative-line-numbers-in-ideavim-for-intellij

PR #26, opened 2014-03-24
Run "launchctl stop com.apple.SystemUIServer.agent" instead of OS restart

riggs:
The command "launchctl stop com.apple.SystemUIServer.agent" is sufficient to make the changes to the ApplePressAndHoldEnabled setting take effect, rather than requiring a restart of OS X.

PS. I haven't touched Java in 10+ years, so I'm not sure what the best idiom is for two sequential 'process' calls.


vlasovskikh:
Thanks for your PR! I've applied the main part of your pull request on top of the master branch.

PR #32, opened 2014-07-12
Fixes http://youtrack.jetbrains.com/issue/VIM-287

breandan:
`ChangeGroup` looks like it could be refactored to use [`com.intellij.openapi.editor.actions`](https://github.com/JetBrains/intellij-community/tree/master/platform/platform-impl/src/com/intellij/openapi/editor/actions) as there are a few other surprises like this one. [`EditorActionUtil.moveCaretToLineEnd(...)`](https://github.com/JetBrains/intellij-community/blob/master/platform/platform-impl/src/com/intellij/openapi/editor/actions/EditorActionUtil.java#L497) seems to put us exactly where we need to be after the fold, and takes into account soft-wraps etc. - `ChangeGroup.insertNewLineBelow(...)` and `ChangeGroup.insertNewLineAbove(...)` will now behave as expected.

### DIFF
--- a/src/com/maddyhome/idea/vim/group/ChangeGroup.java
+++ b/src/com/maddyhome/idea/vim/group/ChangeGroup.java
@@ -194,7 +194,7 @@ public class ChangeGroup {
    * @param context The data context
    */
   public void insertNewLineBelow(@NotNull final Editor editor, @NotNull final DataContext context) {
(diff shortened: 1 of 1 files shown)


vlasovskikh:
@breandan Did your fix solve this problem for you? I don't remember if I tried / tested it. PR #43 definitely doesn't work in many cases. Why have you closed your PR?

PR #34, opened 2014-07-25
Fix "next method" IdeaVim shortcut ]m or [m for PyCharm.

ppalucki:
Using a ]m "next method" shortcut always ends with NullPointerException.

This merge request does two things:
1. Fixes of minor deprecation method calling (without editor) `createStructureViewModel` (fixes `AbstracMethodError`
2. Deals with the problem of `navigationOffsets` structure preparation caused by invalid Python PSI Classes hierarchy vs other languages PSI classes hierarchy

### DIFF
--- a/src/com/maddyhome/idea/vim/helper/PsiHelper.java
+++ b/src/com/maddyhome/idea/vim/helper/PsiHelper.java
@@ -30,6 +30,8 @@ import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.psi.PsiElement;
 import com.intellij.psi.PsiFile;
 import com.intellij.psi.PsiManager;
+import com.jetbrains.python.psi.PyElement;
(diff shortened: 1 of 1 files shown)


durandom:
@vlasovskikh I'm looking at this from a ruby perspective and have the same problem. All motions that depend on `PsiHelper` basically don't work in the ruby plugin / RubyMine.

Because we get `RubyStructureViewElement` in `addNavigationElements()` instead of `PsiTreeElementBase`. The common interface would be `StructureViewTreeElement`.

But the `getValue()` method only returns an `Object` and this cannot be cast to `PsiElement` 

`RubyStructureViewElement.getValue()` would return an `RPsiElement`

As you see, my Java skills are rudimentary. Is there anything we can do to make this code more generic? Or is it the python and ruby plugins that should change their interfaces?

PR #35, opened 2014-08-13
deselect visual selection range '<,'> when poping up ex entry field.

changwang:
When in visual selection mode, if ex mode is triggered, the visual selection range '<,'> text is always selected. This could be annoying, because I have to press right arrow key to deselect, then type the command.

This change basically deselect the text when the ExTextField gains the focus, set the caret at the end of text.


AlexPl292:
Hi @changwang ! Could you please take a look at this discussion so we can use your change under the new MIT license? Thank you!

PR #39, opened 2014-09-06
VIM-714 Set last column to caret column after delete

dathanb:
Addresses VIM-714 by explicitly setting the editor's last column after a visual delete.


vlasovskikh:
This PR breaks 4 tests in `ChangeActionTest`.

PR #42, opened 2014-10-13
VIM-652 Add support for executing and mapping arbitrary IDEA actions

smartbomb:
_no tests performed..!_

Example usage:
- `:nnoremap gi :action GotoImplementation<CR>`
- `:nnoremap gu :action FindUsages<CR>`
- `:nnoremap gcw :action RenameElement<CR>`

Action overview:
https://github.com/JetBrains/intellij-community/blob/6c6cb47c5a41025549157d64643a98a89481260d/platform/platform-resources-en/src/messages/ActionsBundle.properties

### DIFF
--- a/src/com/maddyhome/idea/vim/ex/handler/ExecuteActionByNameHandler.java
+++ b/src/com/maddyhome/idea/vim/ex/handler/ExecuteActionByNameHandler.java
@@ -49,7 +49,7 @@ public class ExecuteActionByNameHandler extends CommandHandler {
       return false;
     }
 
-    if(action.isInInjectedContext()) {
+    if(action.isInInjectedContext() || action.isEnabledInModalContext()) {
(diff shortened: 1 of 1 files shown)


hari-rangarajan:
This feature seems to be talking abot abbreviations. I was wondering if I could do aliases using this feature. For example,   :alias  myshortcmd  ":action IntelilijCommand"

This way, I can introduce new commands like :myshortcmd

Thanks!

brcolow:
The way to do this would be with vim's [command-line commands](http://vimdoc.sourceforge.net/htmldoc/usr_40.html#40.2). So, this would work like so (in your .ideavimrc):

````vimL
command ToggleWhitespace :action EditorToggleShowWhitespaces<cr>
````

However, executing `:ToggleWhitespace` results in error "Not an editor command: ToggleWhitespace" so this functionality (custom command-line commands) does not seem to be implemented in ideavim.

PR #43, opened 2014-10-19
VIM-287 Fix inserting new line before and after fold

rasendubi:
If you have your window like (notice the fold):

    class Main {
        public static void main(String[] args) {...}
    }<caret>

and then press `O`, you're expecting to get:

    class Main {
        public static void main(String[] args) {...}
        <caret>
    }

but actual result is:

    class Main {
        public static void main(String[] args) {
            <caret>
            // whatever in main
        }
    }

This pull request fixes the problem as well as adds test for it.


vlasovskikh:
I've fixed VIM-287 in a different way. The fix will be available soon via the Early Access Program, see https://github.com/JetBrains/ideavim#get-an-early-access for details.

PR #59, opened 2014-12-08
VIM-855 Fix regexp character class bugs

dezgeg:
Fixes https://youtrack.jetbrains.com/issue/VIM-855

Previously strchr and istrchr didn't consider NUL chars as end-of-string
terminators. This caused problems in regexps using character classes:
a regexp like "[^a]bc" would be effectively treated as "[^abc]bc" - i.e.
some literal characters from the rest of the pattern would accidentally be
included in the character class.


vlasovskikh:
@dezgeg Thanks for your pull request!

PR #60, opened 2014-12-09
VIM-210 Fix focus issues with the Ex panel

dezgeg:
Fixes https://youtrack.jetbrains.com/issue/VIM-210 and also https://youtrack.jetbrains.com/issue/VIM-178 which is a duplicate of VIM-210.

Previously, if the editor window had been splitted, under certain
conditions some actions related to the Ex panel would cause the editor
focus to change to a different split.

The required conditions for this bug to occur:
    - no docked windows (like the Project sidebar) are open
    - 'View -> Navigation bar' is disabled

At least these actions triggered the bug:
    - search with /
    - successfully executing an Ex command
    - dismissing the Ex window with Esc
    - Ex commands that opened the output panel (e.g. :!)

All the deleted lines of the form
    FileEditorManager.getInstance(project).openFile(vf, true);
seem to be very old, non-functional workarounds for the focus issue.

### DIFF
--- a/src/com/maddyhome/idea/vim/ex/ExOutputModel.java
+++ b/src/com/maddyhome/idea/vim/ex/ExOutputModel.java
@@ -56,7 +56,7 @@ public class ExOutputModel {
   public void clear() {
     myText = null;
     if (!ApplicationManager.getApplication().isUnitTestMode()) {
-      ExOutputPanel.getInstance(myEditor).deactivate();
+      ExOutputPanel.getInstance(myEditor).deactivate(false);
     }
   }
 
--- a/src/com/maddyhome/idea/vim/group/MotionGroup.java
+++ b/src/com/maddyhome/idea/vim/group/MotionGroup.java
@@ -140,7 +140,7 @@ public class MotionGroup {
    */
   private void processMouseClick(@NotNull Editor editor, @NotNull MouseEvent event) {
     if (ExEntryPanel.getInstance().isActive()) {
-      ExEntryPanel.getInstance().deactivate();
+      ExEntryPanel.getInstance().deactivate(false);
     }
 
     ExOutputModel.getInstance(editor).clear();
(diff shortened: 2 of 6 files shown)


vlasovskikh:
@dezgeg Thanks for your pull request! The workaround code was indeed obsolete.

khromalabs:
I've compiled and installed this plugin just because of this bug, but documentation was a bit incomplete for a noob like me, specially about how to deploy the plugin, [this guide helped](http://confluence.jetbrains.com/display/IDEADEV/Getting%20Started%20with%20Plugin%20Development) but it is also outdated ([plugin must be installed as this Stackoverflow's thread points out](https://stackoverflow.com/questions/24535241/how-can-i-install-an-intellij-plugin-from-source) ). Anyway thanks for the great plugin.

vlasovskikh:
@khromalabs Have you read https://github.com/JetBrains/ideavim#development-environment ? Are there any steps missing from your point of view?

khromalabs:
@vlasovskikh Yes the project's wiki was my first guide, following it, I found that Idea starts another instance of itself running Ideavim in a sandboxed enviroment (as I understood). To do the actual deployment I've followed this instructions from [my previously quoted Jetbrain's guide](http://confluence.jetbrains.com/display/IDEADEV/Getting%20Started%20with%20Plugin%20Development):
> Prepare your plugin for deployment. To do this, on the main menu, choose Build | Prepare Plugin Module <module name> for Deployment. This creates a .jar or .zip archive file to be used to install and publish your plugin. If the plugin module does not depend on libraries, the .jar archive will be created. Otherwise, a .zip archive will be created that will include all the plugin libraries specified in the project settings.

And then, unlike the previous guide says, I had to install the generated jar file using the Intellij's GUI provided as the linked SO thread says:

> Preferences > Plugins > Install plugin from disk...

If you want a pull request adding this to the documentation just say it :)

vlasovskikh:
@khromalabs It could add it by myself, but a pull request would be welcome as well :) I would add a small change to the process described above: I prepare builds of IdeaVim by running `ant dist` in the project directory instead of using the IDE action for that.

vlasovskikh:
Unfortunately, this PR introduces a regression on Oracle Java 6: the editor doesn't get focus after typing `<Enter>` in the Ex command entry field. See [VIM-930](https://youtrack.jetbrains.com/issue/VIM-930).

vlasovskikh:
Added a workaround for VIM-930.

PR #62, opened 2014-12-14
VIM-575 Don't change cursor position of other splits in visual mode

dezgeg:
Fixes https://youtrack.jetbrains.com/issue/VIM-575

If a file is opened in multiple splits, entering visual mode in one
split would change the cursor position in other splits as well.


vlasovskikh:
@dezgeg Thanks for your pull request! Not sure about the use case described in VIM-575, I failed to reproduce it, but the fix works for the problem you've described.

PR #63, opened 2014-12-14
Fix broken sorting when sorting with leading whitespace

selesse:
`StringUtil.splitByLines` incorrectly splits strings with leading whitespace. For example, ` a\n b\n c\n` is split into `[ a,b,c]` instead of `[ a, b, c]`.

This meant that if an IdeaVim user tried to sort lines with leading whitespace, the sorting was incorrect.

I changed the implementation from `StringUtil.splitByLines` to Guava's `Splitter` and added a unit test.


vlasovskikh:
@selesse Thanks for your pull request!

PR #64, opened 2014-12-24
Fix incorrect behavior in substitution with offset ranges

jdpopkin:
Substitution commands with offset ranges like .,+2s/a/b/g previously
did not work the way they do in Vim (replace a with b on the current
line and the next two lines). This change fixes that bug.


vlasovskikh:
@jdpopkin Thanks for your pull request!

PR #65, opened 2014-12-24
VIM-864 Fix visual marks gettinng changed during visual substitute

dezgeg:
Fixes https://youtrack.jetbrains.com/issue/VIM-864

Previously, running a substitute command in visual mode would
incorrectly set the visual mode end mark (>) to the location of the last
search match.


vlasovskikh:
@dezgeg Thanks for your pull request!

PR #68, opened 2015-01-20
VIM-856 Fix regex lookbehind bugs

dezgeg:
Fixes https://youtrack.jetbrains.com/issue/VIM-856

The regex code has the following bugs, mostly related to lookbehind:
    - Comparison of CharPointers with == is wrong, .equals should be
      used instead.

    - In the original C code, the behind_pos variable is declared as
      'static regsave_T behind_pos;', i.e. not a pointer, and the same
      for other variables of type regsave_T. So assignments to these
      types of variables need to make deep copies.


vlasovskikh:
Thanks for your pull request!

PR #70, opened 2015-01-30
VIM-306 When "set imecontrol" returns to Normal-Mode, IME is turned off.

masanobuimai:
experiment


masanobuimai:
It's a dirty hack, so this RP cancels it.

PR #71, opened 2015-01-30
replace % with current file for external commands

bomgar:


vlasovskikh:
I would prefer `:!gvim '%' &` or something like that to a custom command. This doesn't work at the moment, but maybe we should fix it. You can create a mapping for this command in your ~/.ideavimrc:

    nmap <Leader>v :!gvim '%' &<CR>

PR #72, opened 2015-02-21
VIM-700 Remapping '0' should still allow zero to be entered in count

dezgeg:
Fixes https://youtrack.jetbrains.com/issue/VIM-700

Previously, remapping '0' in command mode would prevent a number zero
from being entered in a command count.


vlasovskikh:
@dezgeg Thank you for your pull request!

PR #74, opened 2015-02-21
Support comments in brace matching

dezgeg:
In Vim, `%` can be used to jump between the `/*` and `*/` of block
comments. Support this functionality in a language-independent manner.


vlasovskikh:
Thanks for your pull request! Merged it with small changes.

PR #75, opened 2015-02-21
Visual block mode overhaul

dezgeg:
Fixes https://youtrack.jetbrains.com/issue/VIM-781 and https://youtrack.jetbrains.com/issue/VIM-845

The current visual block mode implementation has several bugs:

- The GUI never displays the rightmost characters as selected (blue)
- Due to some off-by-one, commands can't act on each line's last character
- It's possible for newlines to get deleted/changed into something else
- If the '$' mode is entered, the GUI doesn't always indicate this
- It's impossible to move to an empty line (VIM-781)

This commit fixes all of those problems. The only remaining problem are
the visible secondary carets, which makes seeing the real cursor a bit
difficult. This could be fixed later if support for per-caret visibility
would be added to IntelliJ core.

### DIFF
--- a/src/com/maddyhome/idea/vim/group/MotionGroup.java
+++ b/src/com/maddyhome/idea/vim/group/MotionGroup.java
@@ -1605,34 +1605,17 @@ public class MotionGroup {
 
   @NotNull
   public TextRange getVisualRange(@NotNull Editor editor) {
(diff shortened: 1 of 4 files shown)


vlasovskikh:
Thanks for your pull request!

PR #76, opened 2015-02-21
VIM-176 Fix arrow key navigation in console output

dezgeg:
Fixes https://youtrack.jetbrains.com/issue/VIM-176

Insert mode motion commands were marked as `Command.TYPE.INSERT`, which
caused them to be treated as buffer modifying commands and as such
couldn't be executed on read-only console output editors, like Run/Debug
tool windows. Make them `Command.TYPE.MOTION` instead.


vlasovskikh:
Thanks for your pull request! Even though it seemed to fix the most parts of the original problem, I went for another solution ed3f87e6a8b88036131441e3758b6f84b50e7704 that removed a legacy workaround instead.

PR #79, opened 2015-03-15
Add mapleader support.

psjay:


vlasovskikh:
Thanks for your pull request! I'll review it later during my next IdeaVim iteration.

This is a PR for [VIM-650](https://youtrack.jetbrains.com/issue/VIM-650).

vlasovskikh:
@psjay Thanks again for your PR! There are two issues with it however:

* It could be simpler since Vim uses the `mapleader` value during the execution of the `:map` command so there is no need for any notification events for variables
* Actually it would be better to extract the Vim expressions evaluator into a separate piece of code that could be reused for commands like `:echo`

I've started a new PR #84 that is based on these ideas and added the very basic support for `:let` and `:echo` that is enough to just use the value of the `mapleader` global variable for implementing VIM-650. See also `VimScriptParserTest`. I might do it by myself later on, but you are welcome to go ahead and help me a little bit with it.

PR #80, opened 2015-03-18
add tests for the block selection mode

jagd:
replace and delete the last colum in the block selection mode do not work as expected, here are the test cases.


dezgeg:
This is a subset of https://github.com/JetBrains/ideavim/pull/75 right?

jagd:
yes, I've noticed #75 after the pull request is created

vlasovskikh:
@jagd Thanks for your pull request. I've already accepted #75 that contains similar tests as well as fixes for some Visual Block mode bugs.

PR #84, opened 2015-04-05
Mapleader support based on the prototype of a Vim script expressions evaluator

vlasovskikh:
Currently only very simple `let` and `echo` commands are supported. Using the `mapleader` value in mappings is coming next.

### DIFF
--- a/src/com/maddyhome/idea/vim/VimPlugin.java
+++ b/src/com/maddyhome/idea/vim/VimPlugin.java
@@ -41,7 +41,7 @@ import com.intellij.openapi.util.SystemInfo;
 import com.intellij.openapi.wm.StatusBar;
 import com.intellij.openapi.wm.WindowManager;
 import com.maddyhome.idea.vim.ex.CommandParser;
-import com.maddyhome.idea.vim.ex.VimScriptParser;
+import com.maddyhome.idea.vim.ex.vimscript.VimScriptParser;
 import com.maddyhome.idea.vim.group.*;
 import com.maddyhome.idea.vim.helper.DocumentManager;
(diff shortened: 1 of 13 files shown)


hari-rangarajan:
I haven't done any plugin development before. I'm wondering if it would be trivial to add a rudimentary ":command" method that would allow to alias ":action intellijinternalcommands" to be abbreviated to a shorter name? Thanks!

tthanh:
I just update IdeaVim to version 4.1 and so happy with mapleader support. But went I use my old mapleader settings, I still can't use space like leader.
```
let mapleader = "\<Space>"
nnoremap <silent> <Leader>i :action ReformatCode<CR>
```

vlasovskikh:
@tthanh Please file a bug about it [here](https://youtrack.jetbrains.com/issues/VIM).

meonlol:
To use a space as a mapleader you can simply do:
```
let mapleader = " "
```

cloudrave:
I can't seem to get the backslash `mapleader` to work.  I have the following currently:
```vim
let mapleader="\\"
```

positron:
`let mapleader=" "` didn't work for me. I had to do `let mapleader=<Space>` for it to recognize space as my leader.

a-b:
What is up to date way to use space key as a leader?
Tried both 

`let mapleader=<Space>`
`let mapleader=" "`

On lastest IDEA and IdeaVim, and seems like both are not working...

plateauu:
Hi 
it is still `let mapleader = " "` 
You may try `noremap <SPACE> <Nop>` as well to disable any space actions

PR #85, opened 2015-04-19
VIM-523 Global mark remembers only the line number

tieTYT:
Bear with me as this is the first thing I've ever done with
Intellij-IDEA plugin development.

I tested this and I believe I fixed the problem.  The cause of the
problem was that the code was checking if the mark was in a different
file but if it was, it was using the editor of the current file to go to
the line of the mark.  I changed the code so that it opens the file of
the mark first, then jumps to its line number.

I need a VirtualFile so that I can get the Editor to open. The Mark stores the path of the file it's in, but the path is not enough information to get a VirtualFile.  I also need the "protocol" of the file.  So, I had to change the structure of the Mark and add a "protocol" field to it.

I had to update the save/load logic to use this new field, too.  This
way, when you restart Intellij-IDEA, the marks can still be used.  For
old marks, I defaulted the "protocol" to the value of "file".

One thing I couldn't figure out how to do is add a test case to check
that this is working correctly.  This is not due to a lack of effort.
In fact, I tried to write the test case first.  But I have been unable
to find any examples anywhere of testing multiple Editors, and that's
what's involved to fix this issue.  Because of this, this issue may be
best used as a reference to fix the problem instead of worthy of being
used in production.  Either way, I'd love to know how to write such a
test.


vlasovskikh:
Thanks for your pull request! I'll review it soon.

vlasovskikh:
Are there any specific files with the protocol other than file:// you want to navigate to? For example, for jump commands we just use `LocalFileSystem.findFileByPath()`, see `MotionGroup.moveCaretToJump()`.

tieTYT:
Sorry for the delay. I'm new to plug-in development so I don't know. But
the tests seem to use a "temp" protocol.

jemag:
@vlasovskikh is there any update / timeline on fixing the global Marks? Sorry but I couldn't find anything else about it other than this github pull request.

shmup:
@vlasovskikh what can I do to help? Are we still addressing:

> @tieTYT Are there any specific files with the protocol other than file:// you want to navigate to? For example, for jump commands we just use LocalFileSystem.findFileByPath(), see MotionGroup.moveCaretToJump().

letientai299:
Come here after looking for Global Mark implementation in ideavim. Any news on this PR?

maegek:
Did this ever get addressed? I am using ideavim and global marks only take me to line number of current file. This PR is the only thing I have found related to the issue.

deontologic:
Bump

yonglam:
Come here after looking for Global Mark implementation in ideavim.  Any solution? Thanks. @vlasovskikh

vlasovskikh:
@tieTYT Thank you for your pull request! It took me ages to merge it. It's finally in master and the EAP build 0.48.5 with the fix is already available. See [the instructions](https://github.com/JetBrains/ideavim#get-an-early-access) on how to get EAP builds of IdeaVim.

I've squashed your commits and rebased them onto master.

tieTYT:
Oh you ended up using mine? That's awesome thanks a lot for letting me
know. I hope it works well.

AlexPl292:
Hey, @DanKaplanSES! Could you please take a look at this discussion so we can use your change under the new MIT license? Thank you!

PR #86, opened 2015-04-24
Add `gc` verb that accepts a motion to comment

dhleong:
Character-wise motions use block comments, and line-wise ones
use line-wise comments, mostly because IntelliJ isn't good at
removing blockwise comments with the shortcut.

Also accepts `gcc` which does a linewise comment on current line

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -363,6 +363,10 @@
     <action id="VimRedo" class="com.maddyhome.idea.vim.action.change.RedoAction" text="Redo"/>
     <action id="VimUndo" class="com.maddyhome.idea.vim.action.change.UndoAction" text="Undo"/>
 
+    <!-- Comments -->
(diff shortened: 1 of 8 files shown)


dhleong:
@vlasovskikh thoughts on this? Would you prefer this be done as a plugin, like the vim-surround support?

dhleong:
This is also rebased off master

vlasovskikh:
@dhleong Yes, this PR would look better as an IdeaVim extension based on the infrastructure introduced in #103. Maybe we should put it on hold until #103 is merged in master.

dhleong:
Sounds good

dhleong:
Closing this in favor of https://github.com/dhleong/ideavim/tree/dhleong/CommentaryExtension, which requires #105 to be merged before it will work (I'll also rebase off of that and add `<Plug>` support to it if we get that merged)

PR #87, opened 2015-04-25
VIM-769 Support for vim-surround

dhleong:
This is an initial stab at defining semantics for replicating vim plugin behavior within IdeaVim, with some preliminary support for tpope's excellent [vim-surround](https://github.com/tpope/vim-surround) plugin.

I'm not sure if there's a more idiomatic way of doing stuff like this, but since I didn't hear back on my request for guidance I just went ahead and made something up.

I created a `Plugin` interface which contains a `Registrar` in which plugins register. Each plugin has an "option name" which is used to enable the plugin using the vim options syntax, eg:

```vim
  set surround
```

Plugins can be activated at runtime using this syntax, but I wasn't sure how to go about disabling them (I think we'd need a hook to register and de-register actions, or perhaps make a subclass of VimCommandAction that won't be automatically registered at init). I'm not sure that's a huge concern, though, as generally you set up your plugins once in your vimrc and just go.

Since plugin bindings are optional and are enabled after init via .ideavimrc, I slightly refactored `ParentNode` to be able to correctly handle adding new `BranchNode` onto a `BranchNode` that already has an `ArgumentNode`. This way, plugins can override built-in bindings as expected, as well as expand on them.

The included vim-surround port supports most of the core surroundings, but currently leaves xml/html tag support to future work, as well as (I'm sure) a few more obscure features that I wasn't aware of, but I wanted to get this PR up and get some feedback on the approach before I try to figure that one out.

Since the `ys` and `cs` motions take more than just a motion, I mimicked vim-surround's implementation and added a `getChar()` command to `KeyHandler`. While necessarily asynchronous, taking a callback, this does block other vim input processing, as expected. If there's a better way to deal with this sort of thing I'd love to hear it, but this seemed like it would prove helpful in porting other vim plugins in the future.

UPDATE: Repeat support via `.` for `cs` and `ys` is achieved by subclassing `Argument` and storing our extra arguments in it (ie, the `SurroundPair` object). I think this is the simplest way to support the more complex surrounds, such as that achieved by `ysiwtdiv>` to surround the inner word with `<div></div>` tags (which is not yet implemented, but I'm thinking ahead for when we figure that out).

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -363,6 +363,8 @@
     <action id="VimRedo" class="com.maddyhome.idea.vim.action.change.RedoAction" text="Redo"/>
     <action id="VimUndo" class="com.maddyhome.idea.vim.action.change.UndoAction" text="Undo"/>
 
+    <action id="VimSurroundMotion" class="com.maddyhome.idea.vim.action.plugin.surround.SurroundMotionAction" text="Surround" />
+
     <!-- Keys -->
     <action id="VimShortcutKeyAction" class="com.maddyhome.idea.vim.action.VimShortcutKeyAction" text="Vim Shortcuts"/>
   </actions>
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -364,6 +364,7 @@
     <action id="VimUndo" class="com.maddyhome.idea.vim.action.change.UndoAction" text="Undo"/>
 
     <action id="VimSurroundMotion" class="com.maddyhome.idea.vim.action.plugin.surround.SurroundMotionAction" text="Surround" />
+    <action id="VimChangeSurrounding" class="com.maddyhome.idea.vim.action.plugin.surround.ChangeSurroundingAction" text="Change Surround" />
     <action id="VimDeleteSurrounding" class="com.maddyhome.idea.vim.action.plugin.surround.DeleteSurroundingAction" text="Delete Surround" />
 
     <!-- Keys -->
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -364,6 +364,7 @@
     <action id="VimUndo" class="com.maddyhome.idea.vim.action.change.UndoAction" text="Undo"/>
 
     <action id="VimSurroundMotion" class="com.maddyhome.idea.vim.action.plugin.surround.SurroundMotionAction" text="Surround" />
+    <action id="VimDeleteSurrounding" class="com.maddyhome.idea.vim.action.plugin.surround.DeleteSurroundingAction" text="Delete Surround" />
 
     <!-- Keys -->
     <action id="VimShortcutKeyAction" class="com.maddyhome.idea.vim.action.VimShortcutKeyAction" text="Vim Shortcuts"/>
--- a/src/com/maddyhome/idea/vim/GetCharListener.java
+++ b/src/com/maddyhome/idea/vim/GetCharListener.java
@@ -1,0 +1,11 @@
+package com.maddyhome.idea.vim;
+
+import javax.swing.*;
+
+/**
+ * @author dhleong
+ */
+public interface GetCharListener {
+
+  void onCharTyped(KeyStroke key, char chKey);
+}
--- a/src/com/maddyhome/idea/vim/KeyHandler.java
+++ b/src/com/maddyhome/idea/vim/KeyHandler.java
@@ -606,6 +606,16 @@ public class KeyHandler {
   }
 
   /**
+   * Check if the given keystroke was a "cancel";
+   *  pressing escape or ctrl-c constitutes a "cancel."
+   */
+  public static boolean isCancelStroke(KeyStroke key) {
+    return key.getKeyCode() == KeyEvent.VK_ESCAPE
+           || ((key.getModifiers() | KeyEvent.CTRL_DOWN_MASK) == KeyEvent.CTRL_DOWN_MASK
+               && key.getKeyCode() == KeyEvent.VK_C);
+  }
+
+  /**
    * Partially resets the state of this handler. Resets the command count, clears the key list, resets the key tree
    * node to the root for the current mode we are in.
    *
--- a/src/com/maddyhome/idea/vim/KeyHandler.java
+++ b/src/com/maddyhome/idea/vim/KeyHandler.java
@@ -53,6 +53,7 @@ import java.util.Stack;
  * actions. This is a singleton.
  */
 public class KeyHandler {
+
   /**
    * Returns a reference to the singleton instance of this class
    *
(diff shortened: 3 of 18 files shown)


expez:
Hot damn.  This is the plugins I miss the most in ideavim.  Thanks for taking this on!

vlasovskikh:
@dhleong Thanks for your pull requests. I might be unavailable for a couple of weeks. I'll review your PRs later on.

dhleong:
Cheers :) I'll probably keep working on them as time allows

huyz:
awesome

brcolow:
In reference to your comment about a more idiomatic way to do plugin registration...have you considered using a <a href="https://docs.oracle.com/javase/tutorial/ext/basics/spi.html#define-the-service-provider-interface">Service Provider Interface (SPI)</a>? Considering Plugin is an interface, it seems to me this would fit very well, or at least provide you with an alternative to a static array of Plugin objects - which to me (and as you clearly state) is less then ideal.

dhleong:
I admit I'm unfamiliar with that API, but it seems like a good candidate if it plays well with IntelliJ. I'll wait to see if @vlasovskikh has any preferences before proceeding with anything though

brcolow:
Certainly. I am very excited by your work here, and want to give you kudos =)

dhleong:
Thanks :)

tlnagy:
Any progress on this? This would be great feature to have. I use vim-surround a lot.

dhleong:
@vlasovskikh Have you had a chance to take a look at this? Thoughts on how plugins should declare themselves?

vlasovskikh:
@dhleong Not yet, sorry. I think there is no need in creating a Vim plugin mechanism for IdeaVim. Currently it looks like there might be only a few plugins worth "porting" to IdeaVim, so it would be simpler to just include them as a part of the core IdeaVim. If this situation changes, we will switch to some plugin system for that kind of features.

dhleong:
Okay, sounds good, but I assume we want to be able to enable and disable
plugins in the .ideavimrc? I believe my initial implementation just uses a
map or an array or something, but @brcolow proposed using SPI, and I guess
IntelliJ has its own XML mechanism for providing feature implementations.
Since, as you said, there are only a few plugins that might get ported, an
array may be sufficient, but I'm open to something fancier. Just wanted
your input on a preferred method.

portOdin:
What are the commands for using your surround implementation? I tried cs" and ys" but it ignores c and s just switches to input mode.

dhleong:
`ds`, `cs`, and `ys` should all be supported as they are in [vim-surround](https://github.com/tpope/vim-surround) (insofar as the relevant motion is also supported in ideavim). See the [unit tests](https://github.com/dhleong/ideavim/blob/dhleong/Surround/test/org/jetbrains/plugins/ideavim/action/plugin/surround/SurroundPluginTest.java) for examples.

portOdin:
Nevermind, I forgot to `set surround`. Great work, by the way.

dhleong:
@vlasovskikh I think the biggest reason for requiring plugins be enable/disable-able is to be able to handle such plugins that override default bindings. vim-surround is innocuous enough, but something like [vim-sneak](https://github.com/justinmk/vim-sneak), for example, steals your `s`-key by default. I suppose for such plugins we could instead add support for `<Plug>FUNCTION` things and expose plugin functionality that way, but that might be a bit much to deal with. The `set` syntax, at least, mimics that of plugin managers like vim-plug. 

This doesn't address how plugins are declared in code, but perhaps we might prefer supporting the syntax from Vundle or vim-plug with the same github artifacts? That way any plugins a user might have in their .vimrc (if sourced as mentioned in the README) will be automatically enabled. Doesn't help pathogen users, of course, and there might be confusion if there's any difference between the vim plugin and the ideavim plugin (which I suspect there would be).... Just a thought.

nsh87:
Merge please, this is a sought-after feature. It's listed as a ["Major" priority](https://youtrack.jetbrains.com/issue/VIM-769) and seems like @dhleong has done the work already.

seanknox:
@dhleong @vlasovskikh What's needed to accept this PR? This is a widely used feature; [3000+ stars](https://github.com/tpope/vim-surround) on the original vim plugin repo alone.

dhleong:
Waiting on @vlasovskikh to decide how he wants plugins to be provided/specified. However we do it here will inform how other such plugins are made available in IdeaVim....

portOdin:
I switched systems, and can't remember how I set this up. Did you have an IdeaVim jar with this implemented, @dhleong ?

dhleong:
I don't, but you should be able to make one by running `ant dist`

shockone:
A workaround, add to `~/.ideavimrc`:
```
" Change surroundings.
nmap cs'" F'r"f'r"
nmap cs"' F"r'f"r'
nmap cs[( F[r(f]r)
```

dhleong:
@vlasovskikh I've rebased on this off of master so it can clean merge; the tests all pass. Do you have any thoughts on this?

tlnagy:
Great work @dhleong, I look forward to this feature.

vlasovskikh:
@dhleong Going to review this PR very soon.

vlasovskikh:
@dhleong I've done reviewing your PR and I've come up with a few ideas about the infrastructure for plugins for IdeaVim, see my PR #103. I believe with this framework in place it should be easier to port Vim plugins. My PR is a work in progress and I hope you'll join me. I would like to refactor your code so it fits into this plugin framework.

Here are key points of the review and ideas for the plugin infrastructure:

1. Lifecycle of plugins (better call them "IdeaVim extensions" since IdeaVim is a plugin itself)
  * There should be init() and dispose() methods (though it's not possible to disable the original Vim plugins as far as I know)
  * IntelliJ platform has a comprehensive extension mechanism, we can utilize it
2. Implementing IdeaVim extensions / porting Vim plugins
  * We should stay closer to the way the original plugins are written. In particular, the IdeaVim API for plugins should be as similar to the original Vim as possible
  * Registration of the extension functions should be done via `:map` commands, the original key handling loop should not be modified
  * There should be an API facade for the extension writers
3. Synchronous input functions
  * Async versions of `getchar()` necessary due to the async nature of the GUI events are very hard to follow and error-prone
  * It's possible to emulate sync input functions of Vim like `getchar()` and `input()` with modal Swing dialogs that look like the Vim command line
4. The repeat `.` command for the extension actions
  * Let's worry about the repeat command later, it doesn't work for many actions in the original Vim plugins

I'll describe the concrete steps I've taken in my PR #103.

dhleong:
@vlasovskikh Thanks for looking into this in such depth! See my comments on your PR. I think we'll be closing this PR in favor of yours sometime soon. Would you also like my other PR #86 to be implemented as an extension?

dhleong:
Going to go ahead and close this PR in favor of the work going on in PR #103

PR #88, opened 2015-05-09
VIM-944 - Navigation with keypad arrows 

dos65:
Added keypad arrows supporting for VimMotions


AlexPl292:
Hi @dos65!
Than you for your PR and such a long wait!

PR #95, opened 2015-07-29
small question if ctrl-p was replaced with ctrl-shift-p

tomer-ben-david:
and in scala already using ctrl-shift-p for implicit params then with ideavim i see implicit params  with ctrl-shift-p how do I then show params?


vlasovskikh:
@tomer-ben-david Is it just a question about shortcuts or are you proposing some changes in your pull request as well?

PR #97, opened 2015-08-28
Adds zC and zO folding commands

pbu88:
Hi!

I love the folding methods and I was in need of zC and zO (specially zO). So I saw that it was very straightforward to add them and went ahead and do it. I tested this on a dev environment. Please let me know your thoughts.

Cheers!


jbyler:
I have no objection to adding this to the source code, and I think this makes sense given the commands are there and the bindings exist in Vim.  But as a workaround in the meantime, you can just add to your `~/.ideavimrc`:

    nnoremap zC :action CollapseRegionRecursively<CR>
    nnoremap zO :action ExpandRegionRecursively<CR>

pbu88:
It's been in my `.ideavimrc` for a while now :) I use it all the time.

vlasovskikh:
@pbu88 Thanks for your pull request!

PR #98, opened 2015-10-29
Fix too sensitive entering visual mode on mouse click.

cho45:
Sometimes, unexpected visual mode entering by just clicking editor window.

### DIFF
--- a/src/com/maddyhome/idea/vim/group/MotionGroup.java
+++ b/src/com/maddyhome/idea/vim/group/MotionGroup.java
@@ -246,6 +246,7 @@ public class MotionGroup {
 
     int start = editor.getSelectionModel().getSelectionStart();
     int end = editor.getSelectionModel().getSelectionEnd();
+    if (start == end) return;
 
     if (mode == CommandState.SubMode.VISUAL_LINE) {
       end--;


montague:
:+1: for this--happens to me regularly. i use an apple track pad.

jschaf:
I've run into this a couple times.  I would like to stay in normal mode if I mouse-drag in just a single letter.

scott-hand:
I would love for this to be merged. For some reason, this isn't an issue on Linux, but it makes IdeaVim next to useless for me on Windows.

vlasovskikh:
@scott-hand Does it affect MacOS X? I'll check in on Windows.

montague:
@vlasovskikh i use OS X and it affects me.

hopperath:
This happens to me all the time and is very frustrating. I agree that mouse drag within a char is too sensitive activate selection.

vlasovskikh:
I still cannot reproduce the problem and I cannot verify that the suggested PR really fixes it. From what I see it doesn't change the behavior of single-character selection when you move your mouse over less than one character.

Could anyone please record a screencast showing the problem?

ayzenquwe:
@vlasovskikh, if I understand the problem correctly, it's quite easy to reproduce it (on mac): just click anywhere and slightly move a mouse cursor within a clicked character (without releasing a mouse button). The result: visual mode is activated. Original vim with mouse support doesn't activate visual mode if you move a cursor within a single character. Does it make sense?

scott-hand:
Here's an image showing the problem:

Pycharm Sadness

vlasovskikh:
@ayzenquwe @scott-hand Thank you! Now I get it. The related issue is [VIM-1098|https://youtrack.jetbrains.com/issue/VIM-1098].

vlasovskikh:
The proposed change doesn't fix the problem. I can reproduce it in both cases. I will see what's going on here.

vlasovskikh:
An alternative fix is available in the build 0.46.1 via the Early Access Program, see https://github.com/JetBrains/ideavim#get-an-early-access for details. Give it a try!

Please post any comments to [VIM-1098](https://youtrack.jetbrains.com/issue/VIM-1098).

hopperath:
This fixed it for me. Thanks for the fast turnaround!

PR #102, opened 2016-01-06
Incsearch (VIM-271)

lubba:
https://youtrack.jetbrains.com/issue/VIM-271

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -3,8 +3,7 @@
   <component name="EntryPointsManager">
     <entry_points version="2.0" />
   </component>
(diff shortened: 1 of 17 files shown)


lubba:
There are a few problems right now. It would be great if somebody helps me. 
The first that in the text like "something<tag>text<tag>" highlighting doesn't work inside the tags. 
And the second is that there are some causes when highlights don't remove.

ayzenquwe:
It seems I can see the second problem - highlight is not cleared if you type something that eventually is not found. This happens because you remove highlights only if search is successful.

And I can't reproduce the first problem. Actually I don't get it. Can you please clarify it?

vlasovskikh:
@lubba Thank you for your initial work on this feature! See 2cf7d895810884b5704f297849a5fef314834927.

PR #103, opened 2016-01-26
Framework for IdeaVim extensions + sketch of vim-surround port

vlasovskikh:
I've added the following:

1. An extension point `IdeaVIM.vimExtension`. for registering IdeaVim extensions. See plugin.xml and `VimSurroundExtension` as an example. You have to add a `<vimExtension/>` tag into plugin.xml and inherit from `VimExtension` in order to get your plugin registered and get the `set <plguin-name>` option working.
2. The `VimExtensionFacade` contains the parts of the API for extension authors that is similar to Vim.
3. Extension actions are registered via `putExtensionHandlerMapping()` similar to `:map ... :call <your-extension-handler>` in Vim.
4. If your extension handler requires an argument, use `operatorfunc` (available as `setOperatorFunction()`) and `g@` as in Vim.
5. `getchar()` is implemented as a modal dialog that looks like Vim input line at the bottom of the editor, see `getKeyStroke()`
  * I'll add a unit-testable version of the modal dialog-based implementation

Next steps:

1. Add vim-surround commands with unit tests
2. Extend `VimExtensionFacade` according to the needs of the vim-surround port

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -372,6 +372,7 @@
     <action id="VimUndo" class="com.maddyhome.idea.vim.action.change.UndoAction" text="Undo"/>
 
     <!-- Keys -->
-    <action id="VimShortcutKeyAction" class="com.maddyhome.idea.vim.action.VimShortcutKeyAction" text="Vim Shortcuts"/>
+    <action id="VimShortcutKeyAction" class="com.maddyhome.idea.vim.action.VimShortcutKeyAction" text="Shortcuts"/>
+    <action id="VimOperatorAction" class="com.maddyhome.idea.vim.action.change.OperatorAction" text="Operator"/>
   </actions>
 </idea-plugin>
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -78,9 +78,17 @@
     </component>
   </application-components>
 
+  <extensionPoints>
+    <extensionPoint qualifiedName="IdeaVIM.vimExtension" interface="com.maddyhome.idea.vim.extension.VimExtension"/>
+  </extensionPoints>
+
   <extensions defaultExtensionNs="com.intellij">
     <errorHandler implementation="com.intellij.diagnostic.ITNReporter"/>
     <applicationConfigurable instance="com.maddyhome.idea.vim.ui.VimEmulationConfigurable"/>
+  </extensions>
+
+  <extensions defaultExtensionNs="IdeaVIM">
+    <vimExtension implementation="com.maddyhome.idea.vim.extension.surround.VimSurroundExtension"/>
   </extensions>
 
   <actions>
--- a/resources/messages.properties
+++ b/resources/messages.properties
@@ -48,3 +48,4 @@ E385=E385: search hit BOTTOM without match for: {0}
 e_patnotf2=Pattern not found: {0}
 unkopt=Unknown option: {0}
 e_invarg=Invalid argument: {0}
+E774=E774: 'operatorfunc' is empty
--- a/src/com/maddyhome/idea/vim/KeyHandler.java
+++ b/src/com/maddyhome/idea/vim/KeyHandler.java
@@ -337,9 +337,7 @@ public class KeyHandler {
 
   private boolean isEditorReset(@NotNull KeyStroke key, @NotNull CommandState editorState) {
     return (editorState.getMode() == CommandState.Mode.COMMAND || state == State.COMMAND) &&
-           (key.getKeyCode() == KeyEvent.VK_ESCAPE ||
-            (key.getKeyCode() == KeyEvent.VK_C && (key.getModifiers() & KeyEvent.CTRL_MASK) != 0) ||
-            (key.getKeyCode() == '[' && (key.getModifiers() & KeyEvent.CTRL_MASK) != 0));
+           StringHelper.isCloseKeyStroke(key);
   }
 
   private void handleCharArgument(@NotNull KeyStroke key, char chKey) {
--- a/src/com/maddyhome/idea/vim/KeyHandler.java
+++ b/src/com/maddyhome/idea/vim/KeyHandler.java
@@ -33,6 +33,7 @@ import com.maddyhome.idea.vim.command.Argument;
 import com.maddyhome.idea.vim.command.Command;
 import com.maddyhome.idea.vim.command.CommandState;
 import com.maddyhome.idea.vim.command.MappingMode;
+import com.maddyhome.idea.vim.extension.VimExtensionHandler;
 import com.maddyhome.idea.vim.group.RegisterGroup;
 import com.maddyhome.idea.vim.helper.*;
 import com.maddyhome.idea.vim.key.*;
@@ -252,12 +253,19 @@ public class KeyHandler {
       final Runnable handleMappedKeys = new Runnable() {
         @Override
         public void run() {
-          final boolean fromIsPrefix = isPrefix(mappingInfo.getFromKeys(), mappingInfo.getToKeys());
-          boolean first = true;
-          for (KeyStroke keyStroke : mappingInfo.getToKeys()) {
-            final boolean recursive = mappingInfo.isRecursive() && !(first && fromIsPrefix);
-            handleKey(editor, keyStroke, new EditorDataContext(editor), recursive);
-            first = false;
+          final List<KeyStroke> toKeys = mappingInfo.getToKeys();
+          final VimExtensionHandler extensionHandler = mappingInfo.getExtensionHandler();
+          if (toKeys != null) {
+            final boolean fromIsPrefix = isPrefix(mappingInfo.getFromKeys(), toKeys);
+            boolean first = true;
+            for (KeyStroke keyStroke : toKeys) {
+              final boolean recursive = mappingInfo.isRecursive() && !(first && fromIsPrefix);
+              handleKey(editor, keyStroke, new EditorDataContext(editor), recursive);
+              first = false;
+            }
+          }
+          else if (extensionHandler != null) {
+            extensionHandler.execute(editor, context);
           }
         }
       };
--- a/src/com/maddyhome/idea/vim/KeyHandler.java
+++ b/src/com/maddyhome/idea/vim/KeyHandler.java
@@ -265,7 +265,12 @@ public class KeyHandler {
             }
           }
           else if (extensionHandler != null) {
-            extensionHandler.execute(editor, context);
+            RunnableHelper.runWriteCommand(editor.getProject(), new Runnable() {
+              @Override
+              public void run() {
+                extensionHandler.execute(editor, context);
+              }
+            }, "Vim " + extensionHandler.getClass().getSimpleName(), null);
           }
         }
       };
--- a/src/com/maddyhome/idea/vim/action/change/OperatorAction.java
+++ b/src/com/maddyhome/idea/vim/action/change/OperatorAction.java
@@ -1,0 +1,99 @@
+/*
+ * IdeaVim - Vim emulator for IDEs based on the IntelliJ platform
+ * Copyright (C) 2003-2016 The IdeaVim authors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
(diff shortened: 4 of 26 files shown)


vlasovskikh:
@dhleong I'm going to proceed with this PR and refactor parts of your code so it works with this infrastructure. What are your thoughts on this PR?

dhleong:
This actually looks pretty awesome so far. Would you like me to port over my work on the surround plugin? What's the best way for me to do that?

Some thoughts:
1. What about `<Plug>*` mappings? `vim-surround` actually declares these first, then maps `ys`, et al to those mappings. They allow users to do their own mappings, and are also the basis for [vim-repeat](https://github.com/tpope/vim-repeat). Perhaps a `VimExtensionFacade#putExtensionPlugMapping` method?  In this case, `putExtensionHandlerMapping` should also accept a string argument for triggering these; user maps from `.ideavimrc` would also need to check here.
2. Should `putExtensionHanderMapping` rather be a method on `VimExtension`? This could enable it to be associated with the plugin, and automatically `dispose()`d. Perhaps simply a convenience method that calls out to the Facade would be sufficient.
3. Plugin configuration: I don't remember; does `.ideavimrc` currently support setting arbitrary `g:` vars? would this happen before `VimExtension#init()`? I guess they would have to go before the `set <pluginname>`, right?
4. I know I proposed the `set` syntax for enabling plugins, but would it be worth it to mimic some of the native vim plugin managers like `Plug`, and have a syntax like `Plug 'vim-surround'`? Or even `Plug 'tpope/vim-surround'` to give credit to the original author, or `Plug 'dhleong/vim-surround'` if you'd rather credit the port's author.

dhleong:
Also, I think we should definitely consider repeat support. I can't imagine using `vim-surround` without `vim-repeat`. We could even consider doing it as an extension, if you like, but we would also somehow have to support extension-defined functions.

vlasovskikh:
@dhleong I'm going to try poring your `ys` command to the new infrastructure as an example. You might clone my PR and try adding other commands then. I would like to keep an eye on your changes to `VimExtensionFacade` and maybe help you with things like emulating `input()`.

1. Yes, `<Plug>*` mappings could be useful. There is absolutely no problem in adding them now or later on. The approach described by you would be OK.
2. Currently `:unmap` is not supported so there is no mechanism for removing key mappings set up by IdeaVim extensions. Once we have `:unmap` we can easily implement disabling plugins in one way or another.
3. `VimExtension.init()` is executed during the evaluation of `set <plugin-name>` in the middle of the script. `g:` vars are not supported yet.
4. I have no strong opinion here. What are the benefits of `Plug` over `set`? IdeaVim won't support most Vim plugins, so most `Plug` will be left uninterpreted. Does `Plug 'vim-surround'` work with the original Vim and vim-plug/vundle? (I haven't tried them). How popular are pathogen/vim-plug/vundle?
5. Sure it would be a good idea to support `.` repeat for IdeaVim extension handlers. I'm just saying that it looks like that there is no easy and universal way of doing it. Maybe I'm wrong here. It's definitely worth investigating.

vlasovskikh:
I'm not sure about using `ChangeGroup.insertText()` for modifying the document inside IdeaVim extensions. It seems too low-level to me there. I'm leaving it as is for now.

The last thing I'm about to do tonight is surrounding with HTML tags. It will require something like `input()`.

dhleong:
3. Understood. We can always add it when we get desperate for plugin config
4. `Plug 'tpope/vim-surround'` works for vim-plug, but not vundle, etc. There's no real advantage, just thought it might be more familiar syntax—plus, `set` is, I think, just for built-in settings, and these are "extensions" rather than "settings." I'm not passionate about it, though; just wanted to throw it out there while we're working on this
5. Fair enough. I glanced at the source of `vim-repeat` and it's not as obvious as I had first assumed. 
6. Up to you about `ChangeGroup.insertText()`. I think was unsure how else to go about it at the time. `vim-surround` seems to use normal mode maps with `d` and `p`/`P`, but it's kind of a mess. Another point of concern is if we rely on normal mode maps, do we have a way of triggering the "original" behavior without maps like vim does (eg `norm!`)? I can't remember if user mappings override the built-ins or not.

vlasovskikh:
OK, I've added the initial implementation of `ys`. Could you please take a look at it and maybe add other commands?

2. I see your point, let's get the surround extension working first and then have another look.
4. There is a way to run `handleKey()` with mappings disabled. I'm not sure if it's equivalent to `norm!`.

dhleong:
2. Sounds good
4. I think this is exactly what's needed. Not sure if it's really worth trying to use that to mimic how the vim plugin works when we have simpler things like `ChangeGroup.insertText()`, though. 

Awesome! Looks great so far. I love the blocking input stuff; that'll definitely be great for future plugins.

Not sure if I'll have time to work on this until Friday, but I'm excited to get moving.

dhleong:
Okay, I implemented `ds`, `cs`, and (visual) `S` here: `https://github.com/dhleong/ideavim/tree/vlasovskikh/surround` 

I started thinking about how `<Plug>` mappings should be implemented. The simplest would add a single layer of indirection, where the registration stores a mapping of `<Plug>Name -> VimExtensionHandler`, and when you execute `:*map` (either directly, as from .ideavimrc, or from the `VimExtensionFacade`) it does a lookup and associates the keystrokes with the `VimExtensionHandler`. 

However, if the plugin were disabled when the map is executed, later enabling the plugin would not resolve the issue. I played around in Vim a bit, and it's perfectly valid to map to a non-existing `<Plug>` mapping; attempting to execute it just does nothing, without even any error output. Perhaps, then, executing `:*map` should actually bind to an instance of `PlugMappingHandler` (which extends `VimExtensionHandler` which stores all the existing `<Plug>` mappings statically, and looks up the argument on-demand. In other words, `nmap ys <Plug>YSurround` would actually map `ys` to `new PlugMappingHandler("<Plug>YSurround")`, and when executed, would look up `"<Plug>YSurround"` in a static Map and execute it if found. That way, the above-mentioned scenarios won't cause any weird or unexpected behavior.

Thoughts?

dhleong:
I went ahead and implemented `<Plug>`-mapping support based on my above proposal in a separate branch that builds on my first one: https://github.com/dhleong/ideavim/tree/vlasovskikh/surround-Plug

Let me know what you think

dhleong:
@vlasovskikh I just pushed another branch that builds on the `-Plug` branch which sketches support for repeating extension actions, based on [vim-repeat](https://github.com/tpope/vim-repeat): https://github.com/dhleong/ideavim/tree/vlasovskikh/surround-Plug-repeat

It's still incomplete and very proof-of-concept-y, but I wanted to get your thoughts on it before spending too much more time.

vlasovskikh:
Regarding `<Plug>` key mappings. I think we could use exactly the same key mapping mechanism as Vim does here. Take a look at an example from the original vim-surround:

{code}
nmap ys  <Plug>Ysurround
nnoremap <silent> <Plug>Ysurround  :<C-U>set opfunc=<SID>opfunc<CR>g@
{code}

Here `ys` is mapped to a key sequence: a special untypable 1-character "key" `<Plug>` followed by keys "Surround". This mapping will work no matter what. Then you can map the key sequence `<Plug>Ysurround` to anything you like.

In our case there could be two mappings:

{code}
putKeyMapping(N, parseKeys("ys"), parseKeys("<Plug>Ysurround", true);
putExtensionHandlerMapping(N, parseKeys("<Plug>Ysurround"), new Handler(), false);
{code}

We need to handle `<Plug>` in our key mapping code. Maybe associate it with some fake keystroke since everything there is limited to the `KeyStroke` type.

Then the user could put their own mappings for the IdeaVim extension into their ~/.ideavimrc independent of whether the plugin is enabled or not:

{code}
nmap ,s <Plug>Ysurround
{code}

vlasovskikh:
@dhleong I've merged your `cs` `ds` and `S` commits.

vlasovskikh:
@dhleong After a quick looks at surround-Plug-repeat I'm not sure about this approach. What do you think of adding a `VimRepeatableExtensionHandler extends VimExtensionHandler` that has `repeat()` so the repeat extension just asks the action of another plugin to repeat itself?

vlasovskikh:
@dhleong I may be unavailable till the weekend.

vlasovskikh:
@dhleong One of the surround tests is blinking when you run all the tests. Some bits of the editor state may have not reset after performing surround commands.

dhleong:
> @dhleong After a quick looks at surround-Plug-repeat I'm not sure about this approach. What do you think of adding a VimRepeatableExtensionHandler extends VimExtensionHandler that has repeat() so the repeat extension just asks the action of another plugin to repeat itself?

We would need to be able to provide some extra data for the repeat. So the interface would look something like:

```java
interface VimRepeatableExtensionHandler<R> {
  void repeat(R repeatData);
}
```

```
class VimRepeat {
  public static <R> void set(VimRepeatableExtensionHandler<R> handler, R data) {}
}
```

This would let Surround record the arguments, for example

vlasovskikh:
@dhleong Do we really need this parameterization of `repeat()`? Couldn't, say, `YSurroundHandler` just store the data it needs for repeating in its private state and then use it in `repeat()`?

dhleong:
My concern with that was having stale repeat data lying around. When they perform some other action and we need to clear out the repeat in `VimRepeat`, it's easy to just clear it. But, if it's stored by the plugin, we would need another method to say "Hey, get rid of that stuff." In most cases, this probably isn't a problem, but it does seem like a great way to leak something.

dhleong:
RE Using a fake `KeyStroke` for `<Plug>`: I like this idea. Where do we handle that special keystroke, though?

dhleong:
@vlasovskikh Pushed commits for change with tags. The `testVisualSurroundWordParens` fails when run with the whole suite, though; for some reason it's not exiting visual mode reliably, I guess. I'll keep looking around, but would appreciate any insight you might have.

Once you're satisfied with these, I think we should go ahead and merge this without `<Plug>` or repeat support. We can continue to discuss `<Plug>` here until we have some plan, but then I'd like to break into a separate PR so it's a bit less confusing.

vlasovskikh:
@dhleong If we manage to find a fake `KeyStroke` that is suitable for `<Plug>`, the place to handle it will be `StringHelper.parseKeys()`. The `"plug"` stub is already there in its body. The only other place we'll have to update is `StringHelper.toKeyNotation(KeyStroke)` so it will return the correct string representation of this fake `KeyStroke`.

vlasovskikh:
@dhleong At this point I have no strong preference about the mechanism for repeating handlers. Go ahead and try your latest approach.

dhleong:
> @dhleong If we manage to find a fake KeyStroke that is suitable for <Plug>, the place to handle it will be StringHelper.parseKeys(). The "plug" stub is already there in its body. The only other place we'll have to update is StringHelper.toKeyNotation(KeyStroke) so it will return the correct string representation of this fake KeyStroke.

I was referring more to how to execute it. I would expect `KeyHandler#handleKey`, but there's no great way to know when we've reached the end of the input mapping name (EG: `<Plug>Something` vs `<Plug>SomethingMore`). Also, if we wanted to do something like `<Plug>YSurroundeb` (mostly when thinking about the repeat plugin) I think the execution would block before we could feed the `eb` characters... right? Your `repeat()` approach would obviate the need to support that, but my first point is still valid.

vlasovskikh:
@dhleong As far as I can tell you, all the needed machinery for mapping keys is already there in IdeaVim and the only thing left here is parsing of `<Plug>` into a fake `KeyStroke`. My current understanding is that Vim handles `<Plug>something-more-keys-etc` using the only key handling procedure and the list of custom mappings set by `:map` commands.

See `KeyHandler.handleKeyMapping()` and the related stuff. It handles ambiguous mapping prefixes, key typing timeouts and whatnot.

dhleong:
Alright, I'll look into it tomorrow, maybe. Thanks!

dhleong:
@vlasovskikh Okay I pushed a fix for the visual surround unit test. If you know of a cleaner way to do it, I'd love to use it.

vlasovskikh:
Merged your fixes, see ab3cbb1 for my take on restoring the caret position.

dhleong:
Oh, brilliant. I wasn't sure if the TextRange returned there would reflect changes to the editor or not. 

What else remains before we can merge this PR and start on repeat/<Plug>?

vlasovskikh:
Do you want to develop `<Plug>` mappings here or in a separate request?

dhleong:
I think it'll be less confusing to do a separate request, and it's not strictly needed for extensions to work. My plan is to try with the fake KeyStroke approach this weekend.

vlasovskikh:
I'm going to clean up a few things here and merge it in master.

dhleong:
Awesome!

vlasovskikh:
Merged. Thanks for your contribution!

dhleong:
Great! Thanks for your help with it.

vlasovskikh:
There are issues with `cs` in some cases, still figuring out when exactly. And occasionally with getting focus for the modal input panel.

PR #106, opened 2016-02-14
Refactor modal input for Vim Extensions

dhleong:
Doesn't use dialogs, so should fix VIM-1086


vlasovskikh:
@dhleong It's a Java 7+ API. Would it be possible to achieve the same with `EventQueue.push()`?

dhleong:
Ah, you're right. I'll look into `.push()`.

dhleong:
Okay I have some code locally that *should* replicate the behavior with regular `EventQueue`, but (apparently due to http://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8144759) it doesn't work on JDK7+. So I think I'll need to write a compatibility layer that uses `SecondaryLoop` if available, and falls back to `EventQueue`.

vlasovskikh:
Actually IntelliJ 16 will support only Java 8+, but having a fallback for Java 6+ (if possible) would be really nice. I believe we can use `SecondaryLoop` via the Java Reflection API.

dhleong:
Okay @vlasovskikh check out the latest push. `EventQueue.push()` was a bust, but using a `Proxy`, I was able to replicate what the actual `SecondaryLoop` on JDK7 does with JDK6 methods. This gives us a nice compat layer that is a drop-in replacement for `SecondaryLoop` and should have roughly identical behavior across platforms. It degrades to using a real `SecondaryLoop` if available; I think that will be more performant since there's no Proxy in between every event dispatch call. 

Let me know what you think

vlasovskikh:
Thank you for your ideas! Let's use the `SecondaryLoop` compatibility class as the default implementation and fall back to the `Proxy`-based implementation if it's not available. I think we'll be able to drop Java 6 and 7 support in a few months when IntelliJ 16 and others are released. The master branch of IdeaVim already supports only IntelliJ 15+.

I'll clean up your code a bit at the weekend and fix a few issues with handling `<Esc>` and other non-normal key strokes when the secondary loop is active. I've observed a few cases where the loop doesn't terminate upon the first escape or backspace.

dhleong:
Sounds great! I believe the code as-is prefers `SecondaryLoop` if it can find it.

vlasovskikh:
Thanks for your PR! I've added back the commented out `Jdk7SecondaryLoopCompat` and refactored the code a bit. There was a problem with handling `<Esc>` during modal input, I hope I've fixed it.

dhleong:
Ah woops! I must've forgotten to uncomment from when I was testing it
before committing.

Happy to help :)

PR #107, opened 2016-02-14
VIM-1086 Possible fix for focus loss in ModalEntryDialog based on DialogWrapper

vlasovskikh:
@dhleong Could you please help me test it?

At the moment I'm no longer getting focus losses, but my solution is based on guesswork and I'm not sure if I've actually fixed it.


vlasovskikh:
Still getting problems with focus...

PR #111, opened 2016-05-28
VIM 259 and VIM-714 bugs

unloadedroe:
259 : register action for ==
714: incorrect cursor position after delete visual mode.


AlexPl292:
Hi, @unloadedroe!

PR #113, opened 2016-07-15
VIM-1110 Fix caret jumping after escape from block-insert mode

qzed:
A fix for the VIM-1110 issue: After entering the block-insert mode via the block-selection mode (CTRL-v), doing some changes and then leaving the block-insert mode, the caret jumped upwards. Furthermore the jump-amount was related to the amount of lines folded away.

This issue comes from the calculation of the new caret-offset based on the visual line number. Using the logical line number instead fixes this.

### DIFF
--- a/src/com/maddyhome/idea/vim/group/ChangeGroup.java
+++ b/src/com/maddyhome/idea/vim/group/ChangeGroup.java
@@ -448,7 +448,7 @@ public class ChangeGroup {
     if (repeatLines > 0) {
       int vline = editor.getCaretModel().getVisualPosition().line;
       int lline = editor.getCaretModel().getLogicalPosition().line;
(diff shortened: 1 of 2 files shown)


vlasovskikh:
@qzed Thank you for your pull request. Could you please add a unit test for this issue?

qzed:
Sure, i'll have a look at it.

qzed:
I added the unit test.

madelfio:
This would be a helpful fix for me.  Is there anything that needs to be done before merging?

cypok:
This bug is annoying me too. :(
Maybe we can help somehow to merge it in master?

vlasovskikh:
@qzed Thanks for your pull request! Sorry it took me so long to merge it. I've rebased your fix on top of master and added a simpler version of your test.

PR #115, opened 2016-08-10
added :shell command

johngrib:
`:shell` to activate terminal window.


AlexPl292:
https://youtrack.jetbrains.com/issue/VIM-1586

PR #116, opened 2016-10-02
added :vim command

johngrib:
`:vim` to open file with gui vim (gvim, macvim, etc)
use after `set gvimpath` in .ideavimrc

example :
    `set gvimpath=/usr/local/opt/macvim/bin/mvim`


src/com/maddyhome/idea/vim/ex/handler/OpenVimHandler.java
```diff
@@ -0,0 +1,63 @@
...
+    public OpenVimHandler() {
+        super("vi", "m", RANGE_FORBIDDEN | ARGUMENT_FORBIDDEN);
+    }
+
+    public boolean execute(@NotNull Editor editor, @NotNull DataContext context, @NotNull ExCommand cmd) {
+
+        VirtualFile vf = EditorData.getVirtualFile(editor);
+        String filePath = vf.getCanonicalPath();
+
+        StringOption o = (StringOption) Options.getInstance().getOption(Options.GVIM_PATH);
```
Baldrs: Why can't you try to detect system vim in `PATH` if gvimpath is not defined? It should be something simple,  like this example from stackoverflow: 
http://stackoverflow.com/a/32827512/1583006
```java
public static String findExecutableOnPath(String name) {
    for (String dirname : System.getEnv("PATH").split(File.pathSeparator)) {
        File file = new File(dirname, name);
        if (file.isFile() && file.canExecute()) {
            return file.getAbsolutePath();
        }
    }
    throw new AssertionError("should have found the executable");
}
```

sethk: MacOS apps launched by LaunchServices (such as from Finder or the Dock) typically have a very basic `PATH` environment. Don't know about other platforms...

zhuyucheng52:
Place add ex model :g command, this is very useful.

src/com/maddyhome/idea/vim/ex/handler/OpenVimHandler.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+public class OpenVimHandler extends CommandHandler {
+    public OpenVimHandler() {
+        super("vi", "m", RANGE_FORBIDDEN | ARGUMENT_FORBIDDEN);
+    }
+
+    public boolean execute(@NotNull Editor editor, @NotNull DataContext context, @NotNull ExCommand cmd) {
+
+        VirtualFile vf = EditorData.getVirtualFile(editor);
+        String filePath = vf.getCanonicalPath();
```
AlexPl292: vf could be null

src/com/maddyhome/idea/vim/ex/handler/OpenVimHandler.java
```diff
@@ -0,0 +1,63 @@
...
+
+        StringOption o = (StringOption) Options.getInstance().getOption(Options.GVIM_PATH);
+        String gVimPath = o.getValue();
+
+        try {
+            String command = gVimPath + " " + filePath;
+            Runtime.getRuntime().exec(command);
+        } catch (IOException e) {
+            e.printStackTrace();
```
AlexPl292: Do not suppress exceptions. Please show error message to user, or rethrow it with RuntimeException. In the last case the user will be able to report this error to JetBrains.

AlexPl292:
Hi!
After some discussions, we decided not to include this command in IdeaVim because the original vim has a completely different meaning of this command. If the user wants to have this command, he should define it using `:command`.

vlasovskikh:
I would add that in the original Vim this case is handled by `:!vim %` and we would like to have something similar in IdeaVim so that an interactive command is executed in the Terminal tool window similar to what you've implemented for `:shell`.

PR #117, opened 2016-10-10
VIM-862 | Allow an optional Ex range for :action input

abrookins:
`ActionHandler` does not allow Ex ranges with its input, which leads to VIM-862 -- ranges given to an `:action` mapping are ignored. This branch changes `ActionHandler` to allow an optional range.


vlasovskikh:
Thanks for your pull request! I'll review it soon.

abrookins:
This seems to work for character ranges on the same line, but not multi-line ranges. I think I'll have to go deeper.

abrookins:
I'm going to work on this more and submit another PR soon.

PR #119, opened 2016-10-30
Fix NPE in pattern search.

erito:
This Pull request fixes a Null Pointer Exception inside Regex.java due to the conditional `br != lastnode` returning false on an optional sequence (`Magic.PERCENT` switch case)

Steps to reproduce: Perform a search or a search and replace with an optional atom like `\%[foo]bar`.


vlasovskikh:
@erito Thanks for your pull request!

Could you please check for a null pointer before invoking `equals()` on the nullable `br` and add a unit test to `SearchGroupTest` similar to `testSearchMotion()`?

vlasovskikh:
Closing this pull request since there are still unresolved issues after code review.

PR #120, opened 2016-11-06
Fixed . reseting the last find movement where repeating change that a…

gaganis:
…lso uses movement.

On the original vim when we repeat a change with . that also contain an find movement the last find that was typed before the repeat is the one used by the motion repeat ;. Before the commit the plugin would have reset the last movement to use the movement contained in the command that was repeated.

To solve this I took a simple approach where I save the last find motion before starting the repeat and then restoring it to its original value after the repeat is over. In order to access the find motion fields from RepeatChangeAction I created getters in MotionGroup for the movement type and character.

### DIFF
--- a/.idea/codeStyleSettings.xml
+++ b/.idea/codeStyleSettings.xml
@@ -86,6 +86,9 @@
         <option name="WHILE_BRACE_FORCE" value="1" />
         <option name="FOR_BRACE_FORCE" value="1" />
         <option name="FIELD_ANNOTATION_WRAP" value="0" />
+        <MarkdownNavigatorCodeStyleSettings>
+          <option name="RIGHT_MARGIN" value="72" />
+        </MarkdownNavigatorCodeStyleSettings>
         <XML>
           <option name="XML_LEGACY_SETTINGS_IMPORTED" value="true" />
         </XML>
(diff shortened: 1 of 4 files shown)


vlasovskikh:
The corresponding issue is [VIM-1216](https://youtrack.jetbrains.com/issue/VIM-1216).

vlasovskikh:
@gaganis Thanks for your pull request! The only minor issue I've found is could you please revert your changes to .idea/codeStyleSettings.xml and send another pull request? Alternatively, I could edit your commit if you don't mind.

gaganis:
I am embarrassed I missed that! ):
I will do another pull request np.
ThanksGiorgos

Sent from Yahoo Mail on Android 
 
  On Mon, Nov 28, 2016 at 4:19 PM, Andrey Vlasovskikh<notifications@github.com> wrote:   
@gaganis Thanks for your pull request! The only minor issue I've found is could you please revert your changes to .idea/codeStyleSettings.xml and send another pull request? Alternatively, I could edit your commit if you don't mind.

vlasovskikh:
@gaganis Thanks! Looking forward to your new pull request for this bug!

gaganis:
Submitted pull request #121 that fixes the extra file change. 

:)

PR #121, opened 2016-11-28
Fixed . reseting the last find movement 

gaganis:
On the original vim when we repeat a change with . that also contain an find movement the last find that was typed before the repeat is the one used by the motion repeat ;. Before the commit the plugin would have reset the last movement to use the movement contained in the command that was repeated.

To solve this I took a simple approach where I save the last find motion before starting the repeat and then restoring it to its original value after the repeat is over. In order to access the find motion fields from RepeatChangeAction I created getters in MotionGroup for the movement type and character.

	modified:   src/com/maddyhome/idea/vim/action/change/RepeatChangeAction.java
	modified:   src/com/maddyhome/idea/vim/group/MotionGroup.java
	modified:   test/org/jetbrains/plugins/ideavim/action/ChangeActionTest.java

### DIFF
--- a/src/com/maddyhome/idea/vim/action/change/RepeatChangeAction.java
+++ b/src/com/maddyhome/idea/vim/action/change/RepeatChangeAction.java
@@ -41,6 +41,7 @@ public class RepeatChangeAction extends EditorAction {
     public boolean execute(@NotNull Editor editor, @NotNull DataContext context, @NotNull Command command) {
       CommandState state = CommandState.getInstance(editor);
       Command cmd = state.getLastChangeCommand();
+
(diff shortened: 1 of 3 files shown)


vlasovskikh:
@gaganis Thanks! You can get an update with the fix by tomorrow as a part of EAP https://github.com/JetBrains/ideavim#get-an-early-access

gaganis:
I was running the build from my fork branch so I was ok for the bug. I will try the EAP though because I enjoyed working on this and I feel that I would like to contribute more. I will start looking at the patch_welcome issues.

Have a good night!
Giorgos

PR #122, opened 2016-12-04
Fix VIM-636 Broken behavior with shift-s to substitute line

gaganis:
|S| aka |cc| was not behaving properly on files that were empty or
contained 1 to 3 blank lines.

I have found the problem for the 2 and 3 line files to be not counting
newlines when getting of the file size in deleteLine. I have given it
some thought and we are comparing to an offset which includes newlines.
So I am pretty confident that using getFileSize(editor, true) is more
accurate regardless of the fix.

For the empty and one blank line problems I have added special case
checks. I did it this way because I was able to confine the change only
for change line thus minimizing the breakage risk.

### DIFF
--- a/src/com/maddyhome/idea/vim/group/ChangeGroup.java
+++ b/src/com/maddyhome/idea/vim/group/ChangeGroup.java
@@ -856,7 +856,7 @@ public class ChangeGroup {
     }
     if (offset != -1) {
       boolean res = deleteText(editor, new TextRange(start, offset), SelectionType.LINE_WISE);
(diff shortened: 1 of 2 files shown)


AlexPl292:
Hi @gaganis!

gaganis:
Yes! Will do.

gaganis:
OK this one is done :)

AlexPl292:
Hi @gaganis ! I'll close this PR because there is no activity for a long time.
Don't hesitate to reopen it in case of any new changes or create a new one.

PR #123, opened 2016-12-05
VIM-714 Fixed down movement column after visual delete.

gaganis:
It seems that the movements that were done while in visual selection
mode were also moving the editor column information.

Despite the above it seems that original vim will use the column of the
position where the deletion is started, so this is what  I have
implemented to fix this.

### DIFF
--- a/src/com/maddyhome/idea/vim/action/change/delete/DeleteVisualAction.java
+++ b/src/com/maddyhome/idea/vim/action/change/delete/DeleteVisualAction.java
@@ -21,12 +21,15 @@ package com.maddyhome.idea.vim.action.change.delete;
 import com.intellij.openapi.actionSystem.DataContext;
 import com.intellij.openapi.editor.Caret;
 import com.intellij.openapi.editor.Editor;
+import com.intellij.openapi.editor.LogicalPosition;
 import com.maddyhome.idea.vim.VimPlugin;
 import com.maddyhome.idea.vim.action.VimCommandAction;
 import com.maddyhome.idea.vim.command.*;
 import com.maddyhome.idea.vim.common.TextRange;
 import com.maddyhome.idea.vim.handler.CaretOrder;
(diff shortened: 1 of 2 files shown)


AlexPl292:
Hi @gaganis!

gaganis:
Yes! Will do.

gaganis:
Done too :)

AlexPl292:
Hi!
You've made changes only for visual deletion, but it seems that the problem is more deep, than just visual deletion.
This issue also occurs if you do it with commands like `d^j`, `dbj` or `c^<ESC>j` (caret should be positioned **not** at the line start).

gaganis:
Hello @AlexPl292! Thanks for the feedback!

I am not sure I understand the expected behavior from your description for the deletion commands or I was not able to reproduce it. Could you craft some tests like we have for other cases:? 

```java
  // VIM-714 |v|
  public void testDeleteVisualColumnPositionOneLine() {
    doTest(parseKeys("vwxj"),
           "<caret>lorem ipsum dolor sit amet\n" +
           "lorem ipsum dolor sit amet\n",
           "psum dolor sit amet\n" +
           "<caret>lorem ipsum dolor sit amet\n");
  }
```
Since this is unexpected behavior we will have to fix it so we will have to write them eventually.

AlexPl292:
Hi!
Of course. Check [there](https://github.com/JetBrains/ideavim/commit/c0819a83f0af78dc96ea320ec3acf8ad376d301b) tests. Replace `ignored` with `test` in the method name to activate them.
Notice that each test contains simple motion before execution (`l` command). **This motion is important to reproduce the issue**. While writing the tests, I researched some code and  I have assumption that there are some problems with `MotionGroup#moveCaretVertical:1263`. But be careful with it.

AlexPl292:
Hi! We've ended up to fix this issue with this commit: https://github.com/JetBrains/ideavim/commit/16016cb13d0b648a4961802609c3db372c7627e6

PR #124, opened 2017-01-26
Add rudimentary command support

hari-rangarajan:
Experimenting with a simple ":command" interface. Allows mapping intellij custom actions to shorter commands as desired. For example  ":command shell action openTerminal"  to open shell instead of requiring a custom action in java code.

No unit test (as of yet). Would appreciate initial comments.

### DIFF
--- a/src/com/maddyhome/idea/vim/ex/CommandParser.java
+++ b/src/com/maddyhome/idea/vim/ex/CommandParser.java
@@ -17,6 +17,12 @@
  */
 package com.maddyhome.idea.vim.ex;
 
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
 import com.intellij.openapi.actionSystem.DataContext;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.editor.Editor;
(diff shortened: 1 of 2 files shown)


hari-rangarajan:
@johngrib Might be of interest as you have been adding custom command wrappers for :shell and :vim.

vlasovskikh:
@hari-rangarajan What's wrong with existing `:action`? E.g. `:action ActivateTerminalToolWindow`?

hari-rangarajan:
Too verbose and I can't remember any of the action commands. Without auto completion, it is practically not usable IMHO.  It would be easier to map it to simpler commands that work personally (in the spirit of the original vim :command). Also, I prefer typing a command than using a shortcut.

I see similar fine-grained attempts at this in the pull request:
https://github.com/JetBrains/ideavim/pull/116
https://github.com/JetBrains/ideavim/pull/115

vlasovskikh:
See the docs in README about using `:actionlist`. You can map `:action ...` to any shortcuts you like using `:map` commands.

gaganis:
Typing long actions is not ideal and I hadn't considered using :map until now. Looks like a really cool feature that I will definitely look into. Thanks @vlasovskikh 

@hari-rangarajan Can you provide a small clarification, please? In your PR would :command be only able to map to actions?

In the original vim :command is used as a much more general mechanism where you can assign any replacement text as a user command. If we were to introduce :command I would have liked that we tried to make more generic and not cover only a special case.

hari-rangarajan:
@gaganis It is wrapping any Ex command into a command name. So you can execute any other Ex command with it too.

@vlasovskikh It is just me. I prefer to use Ex commands instead of mapping for commands I use rarely. It is hard for me to remember many cryptic mappings and I have to type it fast  if you are treating that as a shortcut.

gaganis:
I also am a fan of Ex commands to be honest so I look favorably at this.

I will try to pull this into my local build to check it out over the next days.

icholy:
I found this issue while trying to figure out how to map `:bd` to `:action CloseActiveTab`

hari-rangarajan:
For anyone interested, I'm experimenting with an auto-completeable Vim command line interface: https://github.com/hari-rangarajan/IdeaExActionBar

I used stock IntelliJ components to get the auto-complete working plus ideaVim's command line panel code. It would require some work to integrate it back to ideaVim because it uses Swing components directly but it can live as an individual plugin and complement ideaVim.

icholy:
@hari-rangarajan that looks great!!

mikeholler:
Any chance of this getting in anytime soon? I'm really interested in this feature.

AlexPl292:
Hi @hari-rangarajan !
What about the status of this PR? As I see, it's not yet complete and allows only mapping to `action` commands, right?
We'd like to have more powerful `command` support. More similar to real vim. Do you have plans to continue this PR?

AlexPl292:
Closed due to no response and new PR: https://github.com/JetBrains/ideavim/pull/181

PR #125, opened 2017-03-07
Focus correctly on action command execution (based on IntelliJ's openapi gotoAction implementation)

hari-rangarajan:
focussing correctly after execution of action commands. These commands
can also be popup menus.

With this fix, vim :action commands can be used where new dialogs are
popped up.

Example:  nmap <leader>ru :action RecentFiles<CR>

This will work with this fix; originally, it never focussed on the
dialog that popped up.


AlexPl292:
Hi @hari-rangarajan !
Thank you for such a long wait! We will be happy to review the pull request, but do you have the opportunity to update it in connection with the latest changes and resolve all conflicts?

PR #128, opened 2017-03-24
VIM-1170 Fix to enable ideaVim plugin in read-only editors (console output window)

hari-rangarajan:
Fixes the plugin from working correctly in console output
window and possibly other usages elsewhere.


caleb-allen:
Any update on this? This would be very helpful

vlasovskikh:
Thanks for your pull request. I went for another solution ed3f87e6a8b88036131441e3758b6f84b50e7704 that removed a legacy workaround.

PR #131, opened 2017-08-07
[WIP] VIM-1177 Cannot move to last blank line

sei40kr:
Fixed the issue that the caret won't go to the final line when it's blank.


vlasovskikh:
Sorry for my late reply. I closed [VIM-1177](https://youtrack.jetbrains.com/issue/VIM-1177) with the explanation of the current behavior.

PR #132, opened 2017-08-18
Added :tabn(ext) and :tabp(revious) functionality

andreigasparovici:


WillEccles:
I need this very badly, I have F7 and F8 set to these (and F9 set to `:tabe`, but that's irrelevant for an IDE) in vim, and moving to an IDE is hard when your keybinds don't work 😭

niteraleph:
Agree, this would be a very much needed feature

giant995:
I see this feature was requested one year ago and that someone just proposed something. 

So why is this PR stil on hold ? 

It's a really needed feature as it is often used.

src/com/maddyhome/idea/vim/ex/handler/PreviousTabHandler.java
```diff
@@ -0,0 +1,22 @@
...
+public class PreviousTabHandler extends CommandHandler {
+    public PreviousTabHandler(){
+        super(new CommandName[]{
+                new CommandName("tabp", "revious"),
+                new CommandName("tabN", "ext")
+        }, ARGUMENT_OPTIONAL);
+    }
+
+    public boolean execute(@NotNull Editor editor, @NotNull DataContext context, @NotNull ExCommand cmd) {
+        VimPlugin.getMotion().moveCaretGotoPreviousTab(editor, context,1 );
```
AlexPl292: What about :tabprevious {count} command?

src/com/maddyhome/idea/vim/ex/handler/NextTabHandler.java
```diff
@@ -0,0 +1,23 @@
...
+public class NextTabHandler extends CommandHandler {
+
+    public NextTabHandler() {
+        super(new CommandName[]{
+                new CommandName("tabn", "ext")
+        }, ARGUMENT_OPTIONAL);
+    }
+
+    public boolean execute(@NotNull Editor editor, @NotNull DataContext context, @NotNull ExCommand cmd) {
+        VimPlugin.getMotion().moveCaretGotoNextTab(editor, context, -1);
```
AlexPl292: What about :tabnext {count} command?

andreigasparovici:
Hi, @AlexPl292. I'm not able to work on the project anymore. Sorry about that.

AlexPl292:
OK, no problem. Then we merge it and solve all the problems ourselves. Thank you for PR!

PR #133, opened 2017-09-30
Add argtextobj.vim plugin emulation

igrekster:
This is a "port" of VIMs [argtextobj.vim](https://vim.sourceforge.io/scripts/script.php?script_id=2699) script. It adds two new text objects `ia` and `aa` that represent an function argument. This implementation also supports angle brackets and curly braces which is handy for C++ templates and Java generics and C++ brace initialization lists.

### DIFF
--- a/src/com/maddyhome/idea/vim/extension/argtextobj/VimArgTextObjExtension.java
+++ b/src/com/maddyhome/idea/vim/extension/argtextobj/VimArgTextObjExtension.java
@@ -7,8 +7,8 @@ import com.intellij.openapi.editor.Editor;
 import com.maddyhome.idea.vim.VimPlugin;
 import com.maddyhome.idea.vim.command.*;
 import com.maddyhome.idea.vim.common.TextRange;
+import com.maddyhome.idea.vim.extension.VimExtension;
 import com.maddyhome.idea.vim.extension.VimExtensionHandler;
-import com.maddyhome.idea.vim.extension.VimNonDisposableExtension;
 import com.maddyhome.idea.vim.handler.TextObjectActionHandler;
 import com.maddyhome.idea.vim.listener.SelectionVimListenerSuppressor;
 import com.maddyhome.idea.vim.listener.VimListenerSuppressor;
@@ -24,12 +24,27 @@ import static com.maddyhome.idea.vim.group.visual.VisualGroupKt.vimSetSelection;
 import static com.maddyhome.idea.vim.helper.StringHelper.parseKeys;
 import static java.util.Collections.emptyList;
 
-public class VimArgTextObjExtension extends VimNonDisposableExtension {
+/**
+ * @author igrekster
+ */
+
+public class VimArgTextObjExtension implements VimExtension {
 
   @NotNull
   @Override
   public String getName() {
     return "argtextobj";
+  }
+
+  @Override
+  public void init() {
+
+    putExtensionHandlerMapping(MappingMode.XO, parseKeys("<Plug>InnerArgument"), getOwner(), new VimArgTextObjExtension.ArgumentHandler(true), false);
+    putExtensionHandlerMapping(MappingMode.XO, parseKeys("<Plug>OuterArgument"), getOwner(), new VimArgTextObjExtension.ArgumentHandler(false), false);
+
+    putKeyMapping(MappingMode.XO, parseKeys("ia"), getOwner(), parseKeys("<Plug>InnerArgument"), true);
+    putKeyMapping(MappingMode.XO, parseKeys("aa"), getOwner(), parseKeys("<Plug>OuterArgument"), true);
+
   }
 
   /**
@@ -108,17 +123,6 @@ public class VimArgTextObjExtension extends VimNonDisposableExtension {
     }
   }
 
-  @Override
-  protected void initOnce() {
-
-    putExtensionHandlerMapping(MappingMode.XO, parseKeys("<Plug>InnerArgument"), getOwner(), new VimArgTextObjExtension.ArgumentHandler(true), false);
-    putExtensionHandlerMapping(MappingMode.XO, parseKeys("<Plug>OuterArgument"), getOwner(), new VimArgTextObjExtension.ArgumentHandler(false), false);
-
-    putKeyMapping(MappingMode.XO, parseKeys("ia"), getOwner(), parseKeys("<Plug>InnerArgument"), true);
-    putKeyMapping(MappingMode.XO, parseKeys("aa"), getOwner(), parseKeys("<Plug>OuterArgument"), true);
-
-  }
-
   /**
    * Helper class to find argument boundaries starting at the specified
    * position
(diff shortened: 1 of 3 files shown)


yetone:
LGTM!!!

igrekster:
@vlasovskikh Anything I can do to help to get this merged?

abonec:
May be Idea already has same feature outside vim plugin?

README.md
```diff
@@ -96,6 +96,7 @@ Emulated Vim plugins:
 
 * vim-surround
 * vim-multiple-cursors
+* argtextobj.vim
```
AlexPl292: Could you please remove `argtextobj` from readme? We will add this information after the public release with this functionality.

src/com/maddyhome/idea/vim/extension/argtextobj/VimArgTextObjExtension.java
```diff
@@ -0,0 +1,562 @@
...
+import java.util.Stack;
+
+public class VimArgTextObjExtension extends VimNonDisposableExtension {
+
+  @NotNull @Override public String getName() { return "argtextobj"; }
+
+  /**
+   * A text object for an argument to a function definition or a call.
+   */
+  @SuppressWarnings("ComponentNotRegistered") // Manually registered
```
AlexPl292: Redundant suppression

fan-tom:
> May be Idea already has same feature outside vim plugin?

At least it has `Move element left/right` actions

README.md
igrekster: @AlexPl292, thank you very much for the review!

> I can see A LOT of pretty complicated logic here. I'm not sure I can understand it completely and such code requires good tests.

The extension uses a heuristic algorithm and thus has a somewhat limited application given the huge syntax space of multiple languages. An alternative would be to use the a proper language support framework, but it would limit the usefulness of the extension. In particular it can work inside comments and incomplete source code. In addition, if the heuristic fails, you'd simply undo it's effect. While I definitely agree the existing level of testing was subpar, I'm not sure how valuable or feasible it would be to have an exhaustive set of tests -- I've been using it daily for a couple of years now without issues. Nevertheless, I've added a couple of corner-case tests (and fixed a few bugs ;-) and got to 100% coverage:
Screenshot_20190710_233923
Please let me know if you would like to test any specific cases -- happy to add them.

I completely understand if you decide that the new level of testing is still not acceptable. In this I would like to know if it would be technically feasible to release the extension as a separate plugin.

> After some playing with this extension, I got an exception while executing `daa` on empty line

Fixed.

>  and I've cleared a body of a function with `cia`.

As I mentioned above, the algorithm is heuristic and it doesn't distinguish between a function body and a brace-initialization in C++. I did put a mitigation for this in the form of limiting the search to 10 lines or 800 characters from the caret position.

> Tests with good coverage are necessary for this PR. And there are a lot of test cases that are out of your existing test: empty file, empty line, executing not on argument, etc. Negative testing should also be provided (when syntax is broken like `function (int x \n{ ... }`, `function (<int x [\n{ ... }`).

src/com/maddyhome/idea/vim/extension/argtextobj/VimArgTextObjExtension.java
igrekster: Removed

igrekster:
I've updated the extension to use the latest extension API. I'm glad to see
the introduction of the API. I think it would help plugin developers
tremendously. I did, however, encounter a few limitations:
* when implementing a text object extension (such as this one), there were
  only two options to define the boundaries of the objects: visual selection
  and moving the caret from its original position. The selection option had a
  number of side effects (undo would select the text, for ex.), while with the
  caret motion option it is impossible to change the starting position;
* the current command counter (`d3aa` -- delete 3 arguments) is not accessible
  from within the extension;
* I couldn't find a clean way of reporting an error and aborting the current
  command;

I don't have a deep understanding of the current architecture, so I've
implemented workarounds for these limitations. Ideally, I'd like to see proper
API support for text-object-like extensions.

igrekster:
I've rebased my change on top of the latest master. Thanks to the great work of @citizenmatt on refactoring the key handling, I was able to remove my crutches, and add the functionality somewhat cleanly.

AlexPl292:
Hey @igrekster !
Thank you for your contribution and patience.
The PR is merged with some changes (eb01b25f35caee0afa8ae850157cbdf820d9f866). Unfortunatelly, I've found some issues with the implementation and therefore disabled support for angle and curly (which aren't supported in original plugin as well). I think this can be handled in a separate PR in case anyone's interested in that.

igrekster:
Thank you very much for taking care of it. I completely understand that this is somewhat complicated code to take onboard. 
The support for `[],{},<>` is extremely useful in C++ I will look into adding the support while preserving the current behaviour.

igrekster:
I spent some time looking at the edge cases @AlexPl292 identified, and, I
couldn't find a reliable heuristic that works well in those cases as well as
in the desired C++ cases without using a proper language parser. Mainly due to
the C++ syntax. Consider an ordinary C++ lambda:

```c++
[capture = x * 10, someCapVar, &capByReef] (auto arg1, auto arg2) {
   return Clazz<int, std::vector<bool>>{capture, arg1, someCapVar + arg2};
}
```

As you can see -- argument lists can be anywhere, in square brackets, in
parentheses without function name, in angle brackets for templates and, in
curly braces for class instance constructors.

I'd like to avoid going into the language parser path, as I believe the
heuristic nature of the plugin is its strength -- it works on broken syntax
and on any language which has some resemblance of argument lists.

While, practically speaking, the identified edge cases do not manifest
themselves as these are not the places where you would typically use the
plugin, I think it is important to keep the default plugin behaviour stable
and predictable.

Thus, would it be acceptable to have the extended list of supported
brackets user configurable? The pull request is here: https://github.com/JetBrains/ideavim/pull/221

AlexPl292:
Hey, @igrekster. I hope you're doing well.
I've just received this issue saying that the extension doesn't remove an argument if there are too many lines. So I wonder where `MAX_SEARCH_LINES` constant was taken from? Is it from the original plugin or this is some sensible value?
https://youtrack.jetbrains.com/issue/VIM-2680

By the way, just want to say that this seems to be the most used extension for me 🔥

igrekster:
Hi, @AlexPl292,

> So I wonder where MAX_SEARCH_LINES constant was taken from? 

Right, so the whole thing is a heuristic, in which we try to figure out if something looks like an argument. And it was by design, as I wanted it to be able to work with broken code (e.g. as we type it), and places where no semantic information is available (such as commented out code), but I didn't want it to take half of the file if it couldn't figure out argument bounds ;-)

So the limit is purely artificial, there is no technical or performance limitation -- I believe the algorithm is linear.

One idea we could explore to address the issue, is to express the limit not in terms of absolute file offsets, but something like `max-runs-forward` and `max-runs-backward` where a _run_  is either a sequence of alpha-numeric characters, or a sequence of non-alpha-numeric characters, or a quoted string. So `foo + "test, string"` will have be 3 runs: `foo`, ` + `, `"test, string"`.

I'd be happy to try it, but I'm not sure when I'll have the time available.

A simpler solution would be to make the limits user-configurable, so they can tailor it to their specific need.

PR #134, opened 2017-11-01
VIM-635. Command gn. 

jonatantierno:
Work in progress to achieve a decent emulation of the gn command.


ashimoon:
Awesome - very excited! Does `.` repeat work with `cgn` and `dgn`?

jonatantierno:
@ashimoon, yes, the dot works fine, both forwards and backwards, as far as I have tested it.

ashimoon:
Just tested it out locally - works flawlessly. THANK YOU!

jonatantierno:
I just updated rebased the branch, adapted the commands to multi-caret (just adapted the interface, I don't thing multicaret is compatible with gn), and also squashed the commits to make the PR cleaner.

Let me know what you think. Is this a good candidate to be merged and released?

twilco:
Would love to see this.

AlexPl292:
Hi @jonatantierno!
Thank you for such a long wait! We will be happy to review the pull request in short time, but do you have the opportunity to update it in connection with the latest changes and resolve all conflicts?

jonatantierno:
Hi @AlexPl292, 
Of course, I will push the fix shortly. Looking forward to the review!

jonatantierno:
Done. Let me know if there is anything else I can do.

src/com/maddyhome/idea/vim/group/ChangeGroup.java
```diff
@@ -1910,6 +1910,56 @@ private void exitAllSingleCommandInsertModes(@NotNull Editor editor) {
...
+   *
+   * @param editor  The editor to insert the text into
+   * @param caret
+   * @param context The data context
+   * @param count   The number times to search before deleting
+   * @param forwards
+   * @return true if able to find and delete the search, false if not
+   */
+  public boolean deleteNextSearch(Editor editor, Caret caret, DataContext context, int count, boolean forwards) {
+    String lastSearch = VimPlugin.getSearch().getLastSearch();
```
AlexPl292: What if `lastSearch` is null?

src/com/maddyhome/idea/vim/group/ChangeGroup.java
```diff
@@ -1910,6 +1910,56 @@ private void exitAllSingleCommandInsertModes(@NotNull Editor editor) {
...
+    int currentPos = editor.getCaretModel().getOffset();
+    TextRange nextRange = new TextRange(currentPos,currentPos);
+    for(int i = 0; i< count; i++) {
+      nextRange = VimPlugin.getSearch().findCurrentOrNext(editor, lastSearch, nextRange.getStartOffset(), false, forwards);
```
AlexPl292: `findCurrentOrNext ` is a static method. It can be called like `ChangeGroup.findCurrentOrNext(...)` isn't it?

src/com/maddyhome/idea/vim/group/ChangeGroup.java
```diff
@@ -1910,6 +1910,56 @@ private void exitAllSingleCommandInsertModes(@NotNull Editor editor) {
...
+   * Change next search, or current if cursor is on a match.
+   *
+   * @param editor  The editor to insert the text into
+   * @param context The data context
+   * @param count   The number times to search before deleting
+   * @param forwards
+   * @return true if able to find and delete the search, false if not
+   */
+  public boolean changeNextSearch(Editor editor, DataContext context, int count, boolean forwards) {
+    String lastSearch = VimPlugin.getSearch().getLastSearch();
```
AlexPl292: See commend above

src/com/maddyhome/idea/vim/group/ChangeGroup.java
```diff
@@ -1910,6 +1910,56 @@ private void exitAllSingleCommandInsertModes(@NotNull Editor editor) {
...
+    int currentPos = editor.getCaretModel().getOffset();
+    TextRange nextRange = new TextRange(currentPos,currentPos);
+    for(int i = 0; i< count; i++) {
+      nextRange = VimPlugin.getSearch().findCurrentOrNext(editor, lastSearch, nextRange.getStartOffset(), false, forwards);
```
AlexPl292: See comment above

src/com/maddyhome/idea/vim/group/MotionGroup.java
```diff
@@ -1987,6 +1987,45 @@ public char getLastFTChar() {
     return lastFTChar;
   }
 
+  public int selectNextSearch(Editor editor, int count, boolean forwards) {
+    String lastSearch = VimPlugin.getSearch().getLastSearch();
```
AlexPl292: See comment above

src/com/maddyhome/idea/vim/group/MotionGroup.java
```diff
@@ -1987,6 +1987,45 @@ public char getLastFTChar() {
...
+    int currentPos = editor.getCaretModel().getOffset();
+    TextRange nextRange = new TextRange(currentPos, currentPos);
+    for (int i = 0; i < count; i++) {
+      nextRange = VimPlugin.getSearch().findNext(editor, lastSearch, nextRange.getStartOffset(), false,forwards);
```
AlexPl292: See comment above

src/com/maddyhome/idea/vim/group/MotionGroup.java
```diff
@@ -1987,6 +1987,45 @@ public char getLastFTChar() {
...
+    final int endOffset = endOffset(nextRange, forwards);
+    for (Caret caret : editor.getCaretModel().getAllCarets()) {
+      MotionGroup.moveCaret(editor, caret, startOffset);
+      setVisualMode(editor, CommandState.SubMode.VISUAL_CHARACTER);
+      MotionGroup.moveCaret(editor, caret, endOffset);
+    }
+    return endOffset;
+  }
+
+  private int startOffset(TextRange nextRange, boolean forwards) {
```
AlexPl292: Hmm, looks like `startOffset` and `endOffset` functions are used only in `selectNextSearch`. Maybe it would be better to use ternary operator instead without creating new functions?

src/com/maddyhome/idea/vim/group/MotionGroup.java
```diff
@@ -1987,6 +1987,45 @@ public char getLastFTChar() {
...
+    if (forwards) {
+      return nextRange.getStartOffset();
+    }
+    else {
+      return nextRange.getEndOffset() - 1;
```
AlexPl292: Are you sure that endOffset cannot be 0?

src/com/maddyhome/idea/vim/group/ChangeGroup.java
```diff
@@ -1910,6 +1910,56 @@ private void exitAllSingleCommandInsertModes(@NotNull Editor editor) {
...
+   * Deletes next search, or current if cursor is on a match.
+   *
+   * @param editor  The editor to insert the text into
+   * @param caret
+   * @param context The data context
+   * @param count   The number times to search before deleting
+   * @param forwards
+   * @return true if able to find and delete the search, false if not
+   */
+  public boolean deleteNextSearch(Editor editor, Caret caret, DataContext context, int count, boolean forwards) {
```
AlexPl292: Did you forget `@NotNull` before `DataContext context`?

src/com/maddyhome/idea/vim/group/ChangeGroup.java
```diff
@@ -1910,6 +1910,56 @@ private void exitAllSingleCommandInsertModes(@NotNull Editor editor) {
...
+  /**
+   * Change next search, or current if cursor is on a match.
+   *
+   * @param editor  The editor to insert the text into
+   * @param context The data context
+   * @param count   The number times to search before deleting
+   * @param forwards
+   * @return true if able to find and delete the search, false if not
+   */
+  public boolean changeNextSearch(Editor editor, DataContext context, int count, boolean forwards) {
```
AlexPl292: Did you forget `@NotNull` before `DataContext context`?

src/com/maddyhome/idea/vim/gn/VisualSelectPreviousSearch.java
```diff
@@ -0,0 +1,38 @@
...
+import com.maddyhome.idea.vim.handler.MotionEditorActionHandler;
+import org.jetbrains.annotations.NotNull;
+
+public class VisualSelectPreviousSearch extends MotionEditorAction {
+    public VisualSelectPreviousSearch() {
+        super(new Handler());
+    }
+
+    private static class Handler extends MotionEditorActionHandler {
+        public int getOffset(@NotNull Editor editor, DataContext context, int count, int rawCount, Argument argument) {
```
AlexPl292: Did you forget `@NotNull` before `DataContext context`?

src/com/maddyhome/idea/vim/gn/VisualSelectNextSearch.java
```diff
@@ -0,0 +1,38 @@
...
+import com.maddyhome.idea.vim.handler.MotionEditorActionHandler;
+import org.jetbrains.annotations.NotNull;
+
+public class VisualSelectNextSearch extends MotionEditorAction {
+    public VisualSelectNextSearch() {
+        super(new Handler());
+    }
+
+    private static class Handler extends MotionEditorActionHandler {
+        public int getOffset(@NotNull Editor editor, DataContext context, int count, int rawCount, Argument argument) {
```
AlexPl292: Did you forget `@NotNull` before `DataContext context`?

src/com/maddyhome/idea/vim/group/ChangeGroup.java
```diff
@@ -1910,6 +1910,56 @@ private void exitAllSingleCommandInsertModes(@NotNull Editor editor) {
...
+    int currentPos = editor.getCaretModel().getOffset();
+    TextRange nextRange = new TextRange(currentPos,currentPos);
+    for(int i = 0; i< count; i++) {
+      nextRange = VimPlugin.getSearch().findCurrentOrNext(editor, lastSearch, nextRange.getStartOffset(), false, forwards);
+      if (nextRange == null) {
+        return false;
+      }
+    }
+
+    for (Caret caret : editor.getCaretModel().getAllCarets()) {
```
AlexPl292: IJ reports that "this is not a loop" because it has always only one iteration. Is here everything fine?

AlexPl292:
I hope I didn't scare you much :)
Some of the problems look just like the fact that I looked at it with a fresh look.

So, what are your plans for all this? Do you have a desire to continue this task?

jonatantierno:
Hi Alex.
Thanks for your comments and advice. Yes, for sure I would like to finish this task. Give me some time to process all the comments. I remember I intended initially to make  motions instead of commands as you say, but I was not able to do so. I do no remember why, out of the top of my head. Let me refresh the problems I found, and maybe you can suggest a solution. 

Again thanks for the review, I will be working on this soon.

jonatantierno:
Ok, so I had some time to look at some of your comments. 

Regarding the first one, that I should not have made a new command for each movement. As I said, that was my first approach too, but then I realized that gn is not in fact a motion in this respect. In a normal command with an argument, what happens is that the affected text is that between the initial position of the cursor, and the position after applying the movement (e.g. : dw deletes the text between current position and the end o the word). This area is defined by the value returned by the method MotionEditorActionHandler.getOffset(). 
However, this is not what happens with gn: The area between the current position and the next search must remain unaffected. I did not found a way to express this with MotionEditorAction, and that is why I created different commands. If I missed something there please let me know, the truth is that my grasps of the concepts handled in the code may be somehow limited. 

As for the multicaret. In my opinion, it does not make much sense to use this command with multicaret, more like "instead of" multicaret. Actually, if you google "vim multicaret", the first entry is an article titled [https://medium.com/@schtoeffel/you-don-t-need-more-than-one-cursor-in-vim-2c44117d51db]("You don’t need more than one cursor in vim"), that talks precisely about the gn command! So, what I would like to do is to remove the multicaret support. Also, I wouldn´t know how it So, what I would propose was to remove whatever changes you say suggest that gn may support multicaret. What do you think about this?

I will solve the rest of the issues you mention, including the new handler registration, but probably it would be better if we agreed on the issue of "commands or motions" first, since it may result in a big change.

AlexPl292:
After reading your comment and spending some time with playing `*gn` commands I found the following. Commands like `dgn` don't work like "Search for next pattern, delete it". They work like "Search for next pattern, visually select it, delete visual selected area". This led to the fact that I discovered that such strange commands like `dvj`, `dviw` do not work in ideavim (although they are buggy in the original vim). The full concept of "edit + (motion with selection)" doesn't work.

So, no surprises that you had some troubles with `gn` commands, which are based on this concept.

We had a long discussion with @vlasovskikh about this problem and decided that we should try to add properly handling of this situation. So, for now I'll make some changes in ideavim's core and after that I hope that you could finish `gn` task.

I'll write you when refactoring is finished.

jonatantierno:
Well, I am happy at least that I did not miss some obvious solution to the task. While you work on the refactor I will fix the rest of the issues you spotted, that needed to be fixed anyway. Let me know if I can help in anything else, and thanks again.

AlexPl292:
Okay, looks like it takes some time. Moreover, I'd like to have `gn` and `gN` commands already for some reasons.
What about next deal: you finish `gn` and `gN` commands and I merge PR. `dgn`, `cgn`, etc. commands will work after some core changes and you should not care about them.

jonatantierno:
That sounds great. I will do those changes soon, then.

src/com/maddyhome/idea/vim/group/ChangeGroup.java
jonatantierno: Yes, call changed.

src/com/maddyhome/idea/vim/group/ChangeGroup.java
jonatantierno: Yes, changed

src/com/maddyhome/idea/vim/group/ChangeGroup.java
jonatantierno: You are right. Check added.

src/com/maddyhome/idea/vim/group/ChangeGroup.java
jonatantierno: Changed.

src/com/maddyhome/idea/vim/group/MotionGroup.java
jonatantierno: Changed.

src/com/maddyhome/idea/vim/group/MotionGroup.java
jonatantierno: Changed.

src/com/maddyhome/idea/vim/group/MotionGroup.java
jonatantierno: Ternary operator added, but methods added for readability. Specially after adding the check of the comment below.

src/com/maddyhome/idea/vim/group/MotionGroup.java
jonatantierno: Taken care of with Math.max()

src/com/maddyhome/idea/vim/group/ChangeGroup.java
jonatantierno: You are right, but I deleted the whole method.

src/com/maddyhome/idea/vim/gn/VisualSelectPreviousSearch.java
jonatantierno: Yes, added.

src/com/maddyhome/idea/vim/group/ChangeGroup.java
jonatantierno: Yes, I missed that one. Changed.

src/com/maddyhome/idea/vim/gn/VisualSelectNextSearch.java
jonatantierno: Yes, added.

jonatantierno:
Done. I deleted the change and delete commands, and resolved all the bugs you reported, I believe. I also changed the registration as you suggested. Let me know what you think.

jonatantierno:
I just realized gN does not work right when repeated. Expect another upload soon...

jonatantierno:
Now it is fixed. I think that is all.

src/com/maddyhome/idea/vim/group/SearchGroup.java
```diff
@@ -647,6 +654,43 @@ else if (start >= offset) {
...
+  }
+
+  @Nullable
+  private static TextRange findNext(Editor editor,
+                                    String lastSearch,
+                                    TextRange nextRange,
+                                    boolean forwards,
+                                    int count) {
+    for (int i = 0; i < count; i++) {
+      nextRange = findNext(editor, lastSearch, Math.max(nextRange.getStartOffset() - (forwards?0:1),0), false, forwards);
```
AlexPl292: I'm concerned about the performance in this place. It is not good to make n searches. Could you check `n` command implementation? It seems to me that you can use functions from there and there you can specify the count.

jonatantierno:
Hi, thanks again for your exhaustive tests! I see that you are right, my code behaves differently than vim in the cases you mention. Let me work on it, I will upload a fix soon.

jonatantierno:
Hi again, I solved the bugs you mentioned. As for the loop you mentioned, to be honest I never worry about performance beforehand. That said, I am alredy working in a better solution without the loop. Expect a new commit soon.

jonatantierno:
Done! what I did was add a count parameter to the findNext method, while keeping the original signature for back compatibility. I think that is all, let me know what you think.

AlexPl292:
I think this can be merged :tada:
Okay, so, for now, I'm in work on big changes related to VIM-510, after that (if nothing will interrupt me) I'll start with completing `gn` command. I'll merge there changes into repo a bit later, but I approve it now.

jonatantierno:
Great news, thanks! What ticket should I watch to be up to date with this?

AlexPl292:
https://youtrack.jetbrains.com/issue/VIM-635 - about `gn`
https://youtrack.jetbrains.com/issue/VIM-510 - that's what I'm working on

jonatantierno:
Noted,thanks again!

AlexPl292:
Merged in master, my congratulations again! 🎉

jonatantierno:
Great news! Thank you!

vlastoun:
When the release is planned?

PR #135, opened 2017-11-19
VIM-862 Make :action commands work in visual mode

ikenox:
Fixes the following issue VIM-862 ":action ExtractMethod doesn't work in visual mode".
https://youtrack.jetbrains.com/issue/VIM-862

Not only `:action ExtractMethod` works but also generic :action commands do. 
(e.g. `CommandByLineComment`, `CommandByBlockComment`, `ReformatCode`, ...)

### DIFF
--- a/src/com/maddyhome/idea/vim/ex/handler/ActionHandler.java
+++ b/src/com/maddyhome/idea/vim/ex/handler/ActionHandler.java
@@ -37,7 +37,7 @@ import org.jetbrains.annotations.NotNull;
  */
 public class ActionHandler extends CommandHandler {
   public ActionHandler() {
-    super("action", "", RANGE_FORBIDDEN | DONT_REOPEN);
+    super("action", "", RANGE_OPTIONAL | DONT_REOPEN);
   }
 
   public boolean execute(@NotNull Editor editor, @NotNull final DataContext context,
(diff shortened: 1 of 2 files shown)


vlasovskikh:
@ikenox Thanks for your pull request! There were a couple of more issues I had to fix before it started working as people expected it to work in [VIM-862](https://youtrack.jetbrains.com/issue/VIM-862). I fixed an off-by-one error in restoring the visual selection so it now works fine with visual character mode ranges 2e67c704f793b11a063de39c5d56828a1b6d1085 and I added enabling and resetting the visual mode while the `:action` is being run 1393e60ca8beb47e07157a39dd63828249d3376e.

ikenox:
@vlasovskikh Thanks!
I upgraded to latest EAP build (v0.49.2) to use this feature, and then I found a bug.
While `:'<,'>action CommentByLineComment` works well,`:'<,'>action ReformatCode` doesn't work.
It certainly worked until 2e67c704f793b11a063de39c5d56828a1b6d1085 and seems not to work since 1393e60ca8beb47e07157a39dd63828249d3376e.
I guess that `selectPreviousVisualMode` is incompatible with some :action commands like `ReformatCode` for some reason.
Do you have any ideas?

vlasovskikh:
@ikenox Yes, it turns out `ReformatCodeAction` runs its `AbstractLayoutCodeProcessor` on a pooled thread, which then applies changes in EDT via `invokeLater`. See `AbstractLayoutCodeProcessor#runLayoutCodeProcess`. So there is no way to reliably do anything after formatting is finished. Postponing `<Esc>` via `invokeLater` won't be enough since it will be put into event dispatch queue earlier than the reformatters' `invokeLater`.

There may be other actions like that with async completion.

Which of the following workarounds do you prefer?

1. Stay in Visual mode after running the action
2. Return to Normal mode, but keep IDE selection which the Vim emulation cannot access
3. Something else

ikenox:
@vlasovskikh
I actually tried `1`, stay in visual mode, but it feels like a little strange behavior for original vim user. So I prefer `2`, keep IDE selection. I think it is better solution at this time even if it is temporary solution.
An fundamental solution is that `AbstractLayoutCodeProcessor` implements a callback method which is called when async method is completed, and call `<Esc>` in the callback...?

vlasovskikh:
@ikenox Pushed an update with (2) to master.

vlasovskikh:
@ikenox Unfortunately `AnAction` is a platform API interface that's been there for ages and that is unlikely to change.

PR #136, opened 2017-11-19
VIM-1379 Changing visual block correctly handles short lines

cypok:
`TextRange.size()` should be used only for iterating over starts/ends.

I am not sure that this is a proper fix. However it seems consistent with current implementation and passes all tests.


noreun:
Thank you! It works great so far.

vlasovskikh:
@cypok Thank your for your pull request! Enjoyed your beautifully written tests!

PR #137, opened 2017-11-20
VIM-1380 Fixed 'cNw' at the last character of a word

cypok:
Bug was introduced while fixing VIM-200


vlasovskikh:
@cypok Thanks for your PR!

PR #138, opened 2017-12-10
VIM-1368: Keep delaying action until focus was actually acquired

dhleong:
See my comments here: https://youtrack.jetbrains.com/issue/VIM-1368

This seemed like a less fragile way of resolving the issue, but in theory it could potentially loop forever. Let me know if you have another, preferred way to handle this situation, if I should add a limit on how many loops it takes, what to do in such a case if so, etc.

Any IntelliJ version after 2017.1.5 is pretty much unusable for me without this fix ;)


vlasovskikh:
See an alternative fix #139. Did it work for you?

dhleong:
Closing in favor of #139

PR #130, opened 2017-05-31
Fix vim 1287

constpetrov:
Add support for commands like "yi(" or "di{" when caret and braces are inside the string

PR #140, opened 2017-12-27
VIM-1287 fix di,cs TextObject processing when carret is inside "string"

aardvark:
This align behavior with VIM of di and cs inside strings. 
Without this invoking di on: 
`{"{foo, "<caret>bar", baz}"}`
Wiil result in only: 
`{}`
When in VIM version result is:
`{"{}"}`
Same behavior is observed in surround plugin processing.  

This is not elegant fix but it is minimal. 
Please let me know if anything should be done additionally. 
Also, thank you for maintaining such amazing project.

### DIFF
--- a/src/com/maddyhome/idea/vim/helper/SearchHelper.java
+++ b/src/com/maddyhome/idea/vim/helper/SearchHelper.java
@@ -120,12 +120,12 @@ public class SearchHelper {
     int loc = blockChars.indexOf(type);
     char close = blockChars.charAt(loc + 1);
 
-    int bstart = findBlockLocation(chars, close, type, -1, pos, count, checkMatchForChar);
+    int bstart = findBlockLocation(chars, close, type, -1, pos, count, (a, b) -> true);
     if (bstart == -1) {
       return null;
     }
 
-    int bend = findBlockLocation(chars, type, close, 1, bstart + 1, 1, checkMatchForChar);
+    int bend = findBlockLocation(chars, type, close, 1, bstart + 1, 1, (a, b) -> true);
     if (bend == -1) {
       return null;
     }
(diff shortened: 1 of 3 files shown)


src/com/maddyhome/idea/vim/helper/SearchHelper.java
```diff
@@ -119,12 +120,12 @@ public static TextRange findBlockRange(@NotNull Editor editor, char type, int co
     int loc = blockChars.indexOf(type);
     char close = blockChars.charAt(loc + 1);
 
-    int bstart = findBlockLocation(chars, close, type, -1, pos, count);
+    int bstart = findBlockLocation(chars, close, type, -1, pos, count, checkMatchForChar);
```
vlasovskikh: This will skip `"` analysis and miss cases like: `{"{foo, <caret>bar", baz}}` that should remain intact after `di{`.

Also it will ignore `'` so `{'{foo, 'bar', baz}'}` after `di{` won't become `{'{}'}`.

src/com/maddyhome/idea/vim/helper/SearchHelper.java
```diff
@@ -119,12 +120,12 @@ public static TextRange findBlockRange(@NotNull Editor editor, char type, int co
...
     if (bstart == -1) {
       return null;
     }
 
-    int bend = findBlockLocation(chars, type, close, 1, bstart + 1, 1);
+    int bend = findBlockLocation(chars, type, close, 1, bstart + 1, 1, checkMatchForChar);
```
vlasovskikh: The comment above applies to this line as well.

aardvark:
Thank you for a feedback. 
Looks like i will need to add a lot more test scenarios to root this problem out and create correct fix.

vlasovskikh:
Closing this one in favor of #130.

PR #143, opened 2018-02-18
VIM-1217 Fix

HiloCoder:
https://youtrack.jetbrains.com/issue/VIM-1217


vlasovskikh:
Thanks for your pull request! I'm putting it aside until I can reproduce this problem. Follow the original issue for further discussion.

I'll be happy to re-open it when the problem is confirmed.

PR #147, opened 2018-05-05
VIM-780: Multicaret motions

vasalf:
This issue fixes [VIM-780](https://youtrack.jetbrains.com/issue/VIM-780).


src/com/maddyhome/idea/vim/group/MotionGroup.java
```diff
@@ -1119,11 +1183,25 @@ public int moveCaretToLineEndOffset(@NotNull Editor editor, int cntForward, bool
     }
   }
 
-  public int moveCaretToLineStart(@NotNull Editor editor) {
-    int logicalLine = editor.getCaretModel().getLogicalPosition().line;
+  /**
+   * @deprecated To move the caret, use {@link #moveCaretToLineEndOffset(Editor, Caret, int, boolean)}
+   */
+  public int moveCaretToLineEndOffset(@NotNull Editor editor, int cntForward, boolean allowPastEnd) {
```
vlasovskikh: This deprecated method is still used.

src/com/maddyhome/idea/vim/group/MotionGroup.java
```diff
@@ -1164,10 +1242,13 @@ public int moveCaretHorizontalWrap(@NotNull Editor editor, int count) {
     }
   }
 
+  /**
+   * @deprecated To move the caret, use {@link #moveCaret(Editor, Caret, int)}
+   */
```
vlasovskikh: This deprecated method is still used in one place.

src/com/maddyhome/idea/vim/group/MotionGroup.java
```diff
@@ -1229,25 +1323,37 @@ public int moveCaretGotoLineFirst(@NotNull Editor editor, int line) {
...
+  public static void moveCaret(@NotNull Editor editor, @NotNull Caret caret, int offset) {
+    moveCaret(editor, caret, offset, false);
+  }
+
+  /**
+   * @deprecated Use the {@link #moveCaret(Editor, Caret, int) multi-caret version} of this function.
+   */
   public static void moveCaret(@NotNull Editor editor, int offset) {
-    moveCaret(editor, offset, false);
+    moveCaret(editor, editor.getCaretModel().getPrimaryCaret(), offset);
```
vlasovskikh: This deprecated method is still use in lots of places.

src/com/maddyhome/idea/vim/ex/handler/JoinLinesHandler.java
```diff
@@ -46,6 +46,6 @@ public boolean execute(@NotNull Editor editor, @NotNull DataContext context, @No
     TextRange range = cmd.getTextRange(editor, context, true);
     range = new TextRange(range.getStartOffset(), range.getEndOffset() - 1);
 
-    return VimPlugin.getChange().deleteJoinRange(editor, range, spaces);
+    return VimPlugin.getChange().deleteJoinRange(editor, editor.getCaretModel().getPrimaryCaret(), range, spaces);
```
vlasovskikh: Why don't you join lines for every caret here?

src/com/maddyhome/idea/vim/action/motion/TextObjectAction.java
```diff
@@ -39,5 +41,11 @@ public TextRange getRange(Editor editor, DataContext context, int count, int raw
     return handler.getRange(editor, context, count, rawCount, argument);
   }
 
+  @Nullable
+  public TextRange getRange(@NotNull Editor editor, @NotNull Caret caret, @NotNull DataContext context, int count,
```
vlasovskikh: `getRange()` without a caret is no longer used.

src/com/maddyhome/idea/vim/action/motion/MotionEditorAction.java
```diff
@@ -36,5 +39,10 @@ public int getOffset(Editor editor, DataContext context, int count, int rawCount
     return handler.getOffset(editor, context, count, rawCount, argument);
   }
 
+  public int getOffset(@NotNull Editor editor, @NotNull Caret caret, @NotNull DataContext context, int count,
```
vlasovskikh: `getOffset()` without a caret is no longer used.

src/com/maddyhome/idea/vim/action/motion/updown/MotionDownAction.java
```diff
@@ -36,16 +41,40 @@ public MotionDownAction() {
...
-    public int getOffset(@NotNull Editor editor, DataContext context, int count, int rawCount, Argument argument) {
-      return VimPlugin.getMotion().moveCaretVertical(editor, count);
+    public Handler() {
+      super(true);
     }
 
-    protected void preMove(@NotNull Editor editor, DataContext context, Command cmd) {
-      col = EditorData.getLastColumn(editor);
+    @Override
+    public int getOffset(@NotNull Editor editor, @NotNull Caret caret, @NotNull DataContext context, int count,
```
vlasovskikh: The down motion doesn't work as expected in several cases when you move around in a file with multiple carets.

For example, given this file:

```
<caret>foo
<caret>bar
baz
```

when I type `lj`, I expect the result to be:

```
foo
b<caret>ar
b<caret>az
```

but the actual result is:

```
foo
<caret>bar
b<caret>az
```

src/com/maddyhome/idea/vim/action/motion/updown/MotionUpAction.java
```diff
@@ -36,16 +39,26 @@ public MotionUpAction() {
...
-    public int getOffset(@NotNull Editor editor, DataContext context, int count, int rawCount, Argument argument) {
-      return VimPlugin.getMotion().moveCaretVertical(editor, -count);
+    public Handler() {
+      super(true);
     }
 
-    protected void preMove(@NotNull Editor editor, DataContext context, Command cmd) {
-      col = EditorData.getLastColumn(editor);
+    @Override
+    public int getOffset(@NotNull Editor editor, @NotNull Caret caret, @NotNull DataContext context, int count,
```
vlasovskikh: There is a similar bug in the up action as it is with the down action (see the example).

src/com/maddyhome/idea/vim/action/change/insert/InsertAfterLineEndAction.java
```diff
@@ -36,7 +36,7 @@ public InsertAfterLineEndAction() {
 
   private static class Handler extends ChangeEditorActionHandler {
     public boolean execute(@NotNull Editor editor, @NotNull DataContext context, int count, int rawCount, @Nullable Argument argument) {
-      VimPlugin.getChange().insertAfterLineEnd(editor, context);
+      VimPlugin.getChange().insertAfterLineEnd(editor, editor.getCaretModel().getPrimaryCaret(), context);
```
vlasovskikh: Why do you run the handler only for the primary caret? It results in a bug in this example:

```
<caret>foo
<caret>bar
baz
```
when I type `A2`, I expect the result to be:

```
foo2
bar2
baz
```

but it is:

```
foo2
2bar
baz
```

src/com/maddyhome/idea/vim/action/change/delete/DeleteJoinVisualLinesAction.java
```diff
@@ -65,6 +69,6 @@ protected boolean execute(@NotNull Editor editor, @NotNull DataContext context,
 
   @Override
   public int getFlags() {
-    return Command.FLAG_EXIT_VISUAL;
+    return 0;
```
vlasovskikh: Don't we need to return to the normal mode and reset visual selection after we have joined the selected lines? When I run `gJ`, I expect my selection to be empty, and it isn't now.

src/com/maddyhome/idea/vim/action/change/delete/DeleteJoinVisualLinesSpacesAction.java
```diff
@@ -65,6 +69,6 @@ protected boolean execute(@NotNull Editor editor, @NotNull DataContext context,
 
   @Override
   public int getFlags() {
-    return Command.FLAG_EXIT_VISUAL;
+    return 0;
```
vlasovskikh: The same issue as for `DeleteJoinVisualLinesAction`.

src/com/maddyhome/idea/vim/action/change/delete/DeleteVisualLinesEndAction.java
```diff
@@ -39,10 +42,11 @@
...
   public DeleteVisualLinesEndAction() {
-    super(new VisualOperatorActionHandler() {
-      protected boolean execute(@NotNull Editor editor, @NotNull DataContext context, @NotNull Command cmd,
-                                @NotNull TextRange range) {
-        if (range.isMultiple()) {
+    super(new VisualOperatorActionHandler(true, CaretOrder.DECREASING_OFFSET) {
+      @Override
+      protected boolean execute(@NotNull Editor editor, @NotNull Caret caret, @NotNull DataContext context,
+                                @NotNull Command cmd, @NotNull TextRange range) {
+        if (CommandState.inVisualBlockMode(editor)/*range.isMultiple()*/) {
```
vlasovskikh: I guess the commented piece of code is no longer needed.

src/com/maddyhome/idea/vim/handler/EditorActionHandlerBase.java
```diff
@@ -51,7 +71,17 @@ public void process(Command cmd) {
     // No-op
   }
 
-  protected abstract boolean execute(@NotNull Editor editor, @NotNull DataContext context, @NotNull Command cmd);
+  /**
+   * @deprecated To implement the action logic, override {@link #execute(Editor, Caret, DataContext, Command)}.
+   */
+  protected boolean execute(@NotNull Editor editor, @NotNull DataContext context, @NotNull Command cmd) {
```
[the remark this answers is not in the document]
vasalf: Actually, any call to the default implementations of `execute()` means there is a bug: either the wrong parameter is given to the base class constructor or the wrong method is implemented. Is it better to return always `false` in these methods?

src/com/maddyhome/idea/vim/group/ChangeGroup.java
```diff
@@ -211,6 +215,34 @@ private void runEnterAction(Editor editor, @NotNull DataContext context) {
     }
   }
 
+  /**
+   * Inserts a new line above the caret position
+   */
+  public void insertNewLineAbove(@NotNull Editor editor, @NotNull Caret caret, @NotNull DataContext context) {
```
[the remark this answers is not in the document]
vasalf: One with a caret given in arguments is for inner use of `ChangeGroup` (it should be private, I guess). It is used by `ChangeLineAction` (help topic `cc`) and `ChangeVisual*Action` (for example, `v_c`). In these actions, the new line is inserted after deletion of some text range and it is decided for each caret whether it should be inserted below (at the end of file) or above (all other cases).

Another one is used by `InsertNewLineAboveAction` (help topic `O`) and `InsertNewLineBelowAction` (help topic `o`). It inserts a new line running an `ENTER` action in the editor. The positive effect of this is that the marks are moved down automatically (if necessary). The negative part is that it is done at once for all of the carets so it is impossible to do this for a concrete caret (so it is impossible to separate cases for each caret as it is done in `changeRange`).

src/com/maddyhome/idea/vim/handler/EditorActionHandlerBase.java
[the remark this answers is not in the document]
vlasovskikh: If it’s always a bug, it’s better to either make the method abstract or raise a not implemented exception.

src/com/maddyhome/idea/vim/group/MotionGroup.java
```diff
@@ -1046,23 +1084,44 @@ public int moveCaretToMiddleColumn(@NotNull Editor editor) {
...
 
     return editor.logicalPositionToOffset(new LogicalPosition(line, pos));
   }
 
-  public int moveCaretToLineStartSkipLeading(@NotNull Editor editor) {
-    int logicalLine = editor.getCaretModel().getLogicalPosition().line;
+  /**
+   * @deprecated To move the caret, use {@link #moveCaretToColumn(Editor, Caret, int, boolean)}
+   */
+  public int moveCaretToColumn(@NotNull Editor editor, int count, boolean allowEnd) {
```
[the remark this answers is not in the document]
vasalf: This method is used in movements to the first/middle/last screen column methods, which doesn't have a version with multiple carets in this PR because they are a bit buggy. Is it better to give the primary caret in their calls?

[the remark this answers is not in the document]
vlasovskikh: Yes, I believe it is, just to make it clear that we have some places in our codebase that work only with the primary caret. It is worth adding a TODO item for all the places where it's a temporary workaround.

src/com/maddyhome/idea/vim/group/ChangeGroup.java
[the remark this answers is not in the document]
vlasovskikh: The insert new line actions `O` and `o` should insert a new line via `<Enter>` action since they have to insert the correct indent and care about other possible extra actions that `<Enter>` does for various languages in IntelliJ. `O` and `o` should be run on all the carets one by one, not on the main caret. I believe the current implementation in this PR doesn't do it.

src/com/maddyhome/idea/vim/handler/MotionEditorActionHandler.java
```diff
@@ -19,35 +19,135 @@
...
+  public MotionEditorActionHandler() {
+    this(false);
+  }
+
+  public MotionEditorActionHandler(boolean runForEachCaret) {
+    super(runForEachCaret);
+  }
+
+  @Override
+  protected final boolean execute(@NotNull Editor editor, @Nullable Caret caret, @NotNull DataContext context,
```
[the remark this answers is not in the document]
vasalf: It is an override of `EditorActionHandlerBase#execute(Editor, Caret, DataContext, Command)`, which is not deprecated.

src/com/maddyhome/idea/vim/action/change/insert/InsertAfterLineEndAction.java
vasalf: This is a temporary kludge (only the primary caret is moved until multicaret `change.insert` actions are not implemented).
I agree that there should have been a TODO note.

src/com/maddyhome/idea/vim/group/ChangeGroup.java
```diff
@@ -211,6 +215,34 @@ private void runEnterAction(Editor editor, @NotNull DataContext context) {
...
+    MotionGroup.moveCaret(editor, caret, VimPlugin.getMotion().moveCaretToLineStart(editor, caret));
+    EditorData.setChangeSwitchMode(editor, CommandState.Mode.INSERT);
+    if (CommandState.getInstance(editor).getMode() != CommandState.Mode.REPEAT) {
+      editor.getDocument().insertString(caret.getOffset(), "\n");
```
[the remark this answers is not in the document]
vasalf: Does it?

I am not sure, what is supposed to be a bug here. Maybe it seems incorrect that only the given caret is being moved after insertion. In fact, `insertString` moves all carets that were to the right side of the insertion position, so this is actually not a bug.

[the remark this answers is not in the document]
vasalf: Actually, there was a bug in `insertNewLineAbove` function: actually we don't need to move the caret here as it is supposed to be at the beginning of the inserted line. But I believe `insertNewLineBelow` works correctly.

src/com/maddyhome/idea/vim/handler/ChangeEditorActionHandler.java
```diff
@@ -19,24 +19,78 @@
...
+    myIsMulticaretChangeAction = runForEachCaret;
+    myCaretOrder = caretOrder;
+  }
+
+  public ChangeEditorActionHandler() {
+    this(false, CaretOrder.NATIVE);
+  }
+
+  @Override
   protected final boolean execute(@NotNull Editor editor, @NotNull DataContext context, @NotNull Command cmd) {
```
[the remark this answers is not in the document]
vasalf: Is it OK not to inherit from `EditorActionHandlerBase`?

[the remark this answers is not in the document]
vasalf: Now I'm thinking whether `EditorActionHandlerBase#execute(Editor, DataContext, Command)` should be deprecated. It is normal for some actions to be run as a "one-off" action (yank, paste, etc).

src/com/maddyhome/idea/vim/handler/MotionEditorActionHandler.java
```diff
@@ -56,11 +156,27 @@ else if (offset >= 0) {
...
+  }
+
+  protected void preMove(@NotNull Editor editor, @NotNull DataContext context, @NotNull Command cmd) {
+  }
+
+  protected void postMove(@NotNull Editor editor, @NotNull DataContext context, @NotNull Command cmd) {
+  }
 
-  protected void preMove(Editor editor, DataContext context, Command cmd) {
+  protected void preMove(@NotNull Editor editor, @NotNull Caret caret, @NotNull DataContext context,
```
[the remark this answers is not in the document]
vasalf: `MotionUpAction` and `MotionDownAction` (at least them) want the caret to be `@NotNull`. Is it okay for them to override a method with `@Nullable` caret?

src/com/maddyhome/idea/vim/ex/handler/JoinLinesHandler.java
vasalf: Per-caret handlers are not written in this PR yet. I will add a TODO there.

src/com/maddyhome/idea/vim/handler/EditorActionHandlerBase.java
```diff
@@ -20,20 +20,40 @@
...
+    this(false);
+  }
+
+  public EditorActionHandlerBase(boolean runForEachCaret) {
+    super(runForEachCaret);
+    myRunForEachCaret = runForEachCaret;
+  }
+
+  @Override
   public final void execute(@NotNull Editor editor, @NotNull DataContext context) {
```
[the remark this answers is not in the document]
vasalf: Removed.

vasalf:
Thank you for the review! I've fixed all of the notices in which we have come to the agreement. I also have added change.insert and change.shift actions support for multiple carets case. @vlasovskikh could you please check them?

vlasovskikh:
@vasalf Thanks for your fixes! I've created [multiple-carets](https://github.com/JetBrains/ideavim/tree/multiple-carets) branch with your changes and identified the commands that don't work correctly with multiple carets. Could you please fix [the list](https://github.com/JetBrains/ideavim/blob/multiple-carets/src/com/maddyhome/idea/vim/package-info.java) for the commands marked with `E`? I would really like to get most of them fixed before I merge this PR into master.

tbrisbane:
Please don't let this PR die! <3

vlasovskikh:
@tbrisbane We keep working on it in https://github.com/JetBrains/ideavim/tree/multiple-carets.

vlasovskikh:
I'm closing this review for now since I've merged all the commits into https://github.com/JetBrains/ideavim/tree/multiple-carets. Please follow that branch as it contains extra fixes for multiple carets support. We'll handle new changes to this branch as separate pull requests.

The current status of `multiple-carets` is that almost all commands work, but we still need an emulated plugin for vim-multiple-cursors that provides commands for selecting the next / previous occurrence and putting a caret there. @karavaevitalii will work on that. A code review for any of new pull requests is welcome.

PR #150, opened 2018-05-29
VIM-1431 handle empty documents when checking if line is empty

jpalus:


src/com/maddyhome/idea/vim/helper/EditorHelper.java
```diff
@@ -521,7 +521,7 @@ public static CharBuffer getLineBuffer(@NotNull final Editor editor, final int l
   public static boolean isLineEmpty(@NotNull final Editor editor, final int line, final boolean allowBlanks) {
     CharSequence chars = editor.getDocument().getCharsSequence();
     int offset = getLineStartOffset(editor, line);
-    if (chars.charAt(offset) == '\n') {
+    if (chars.length() == 0 || chars.charAt(offset) == '\n') {
```
vlasovskikh: Could you please substitute this check with the check for `offset` being within the limits of `chars`'s size? It would be a more reliable fix.

jpalus:
Test added and offset check made more general, though I suppose offset outside of chars' size and different than 0 would probably constitute a bug.

vlasovskikh:
@jpalus Yes, I agree about the offset, but it's better to make it safer. Thanks! I've merged your PR.

PR #152, opened 2018-07-08
VIM-780 Multiple carets

karavaevitalii:


src/com/maddyhome/idea/vim/group/SearchGroup.java
```diff
@@ -526,7 +531,11 @@ else if (command.length() == 1) {
   }
 
   public int searchWord(@NotNull Editor editor, int count, boolean whole, int dir) {
-    TextRange range = SearchHelper.findWordUnderCursor(editor);
+    return searchWord(editor, editor.getCaretModel().getPrimaryCaret(), count, whole, dir);
+  }
+
+  public int searchWord(@NotNull Editor editor, @NotNull Caret caret, int count, boolean whole, int dir) {
```
vlasovskikh: Let's delete `searchWord()` without a caret and let's delete all other unused functions without a caret.

src/com/maddyhome/idea/vim/group/MotionGroup.java
```diff
@@ -412,28 +412,32 @@ public int moveCaretToNthCharacter(@NotNull Editor editor, int count) {
   }
 
   public int moveCaretToMarkLine(final @NotNull Editor editor, char ch) {
-    final Mark mark = VimPlugin.getMark().getMark(editor, ch);
-    if (mark != null) {
+    return moveCaretToMarkLine(editor, editor.getCaretModel().getPrimaryCaret(), ch);
+  }
+
+  public int moveCaretToMarkLine(@NotNull Editor editor, @NotNull Caret caret, char ch) {
+      final Mark mark = VimPlugin.getMark().getMark(editor, ch);
```
vlasovskikh: Wrong indent.

src/com/maddyhome/idea/vim/group/MotionGroup.java
```diff
@@ -521,6 +491,50 @@ public int moveCaretToFileMark(@NotNull Editor editor, char ch) {
...
+    final String fileName = jump.getFilename();
+    if (!vf.getPath().equals(fileName) && fileName != null) {
+      final VirtualFile newFile = LocalFileSystem.getInstance().findFileByPath(fileName.replace(File.separatorChar, '/'));
+      if (newFile == null) {
+        return -2;
+      }
+
+      final Editor newEditor = selectEditor(editor, newFile);
+      if (newEditor != null) {
+        if (spot != -1) {
```
vlasovskikh: It's a possible bug. Previously it was `if (spot == -1)`.

src/com/maddyhome/idea/vim/group/MotionGroup.java
```diff
@@ -1100,6 +1115,10 @@ public int moveCaretToLineStartSkipLeading(@NotNull Editor editor, @NotNull Care
     return moveCaretToLineStartSkipLeading(editor, logicalLine);
   }
 
+  public int moveCaretToLineStartSkipLeading(@NotNull Editor editor, int line) {
```
vlasovskikh: I'd suggest not to change the order of methods without a major reason to do it.

src/com/maddyhome/idea/vim/group/MotionGroup.java
```diff
@@ -1215,21 +1230,38 @@ public int moveCaretToLineStartOffset(@NotNull Editor editor) {
...
+  }
+
+  public int moveCaretToLineScreenStart(@NotNull Editor editor, @NotNull Caret caret) {
+    final int col = getLeftmostScreenColumn(editor);
+    return moveCaretToColumn(editor, caret, col, false);
+  }
+
+  private int getLeftmostScreenColumn(@NotNull Editor editor) {
+    final int leftmost = EditorHelper.getVisualColumnAtLeftOfScreen(editor);
+    return leftmost == 0 ?
```
vlasovskikh: It looks like this fix is actually applicable to all usages of `getVisualColumnAtLeftOfScreen()`. Also it's unrelated to the problem of multiple carets. Let's move it into a separate PR and check the other usages of `getVisualColumnAtLeftOfScreen()`.

src/com/maddyhome/idea/vim/group/MotionGroup.java
```diff
@@ -1315,6 +1347,14 @@ public int moveCaretGotoLineLast(@NotNull Editor editor, int rawCount, int line)
       .normalizeLine(editor, EditorHelper.getLineCount(editor) - 1) : line);
   }
 
+  public int moveCaretGotoLineLast(@NotNull Editor editor, int rawCount) {
```
vlasovskikh: Is it a bug fix? If so, let's add a test and move it into a separate PR. Otherwise let's just discard this change.

src/com/maddyhome/idea/vim/group/CopyGroup.java
```diff
@@ -104,141 +105,249 @@ public boolean yankRange(@NotNull Editor editor, @Nullable TextRange range, @Not
...
+  }
 
-      int start = range.getStartOffset();
-      int end = range.getEndOffset();
-      int endLine = editor.offsetToLogicalPosition(end).line;
-      if (logger.isDebugEnabled()) {
-        logger.debug("start=" + start);
-        logger.debug("end=" + end);
+
+  private int putText(@NotNull Editor editor, @NotNull Caret caret, @NotNull String text, int count, int startOffset) {
```
vlasovskikh: The refactoring of the original `putText()` looks good if we look at its decomposition into separate methods. Unfortunately, some of the original logic inside `putText()` was removed. Let's add tests for all the logic that was removed, or, even better, do this refactoring once again: initially without any changes to the logic, then you could proceed with removing some pieced of `putText()` logic, but only with tests for these pieces.

src/com/maddyhome/idea/vim/group/ChangeGroup.java
```diff
@@ -151,10 +162,7 @@ public void insertLineStart(@NotNull Editor editor, @NotNull Caret caret, @NotNu
    * @param context The data context
    */
   public void insertAfterCursor(@NotNull Editor editor, @NotNull DataContext context) {
-    // TODO: Add multiple carets support
-    MotionGroup.moveCaret(editor, editor.getCaretModel().getPrimaryCaret(), VimPlugin.getMotion()
-      .moveCaretHorizontal(editor, editor.getCaretModel().getPrimaryCaret(), 1, true));
-    initInsert(editor, context, CommandState.Mode.INSERT);
+    insertAfterCursor(editor, editor.getCaretModel().getPrimaryCaret());
```
vlasovskikh: This method is no longer used. Let's remove it and all the other unused methods.

src/com/maddyhome/idea/vim/action/motion/mark/MotionGotoFileMarkAction.java
```diff
@@ -34,8 +36,19 @@ public MotionGotoFileMarkAction() {
...
+
+    @Override
+    public int getOffset(@NotNull Editor editor, @NotNull Caret caret, @NotNull DataContext context, int count,
+                         int rawCount, @Nullable Argument argument) {
+      if (argument == null) {
+        return -1;
+      }
+
+      final char mark = argument.getCharacter();
+      return VimPlugin.getMotion().moveCaretToMark(editor, caret, mark, false);
```
vlasovskikh: It seems that jumping to a mark from another file has stopped working. Please check both file mark jumps and file line mark jumps.

vlasovskikh:
Please update actions for inserting text with no indent.

PR #153, opened 2018-08-02
Multiple carets

karavaevitalii:


resources/META-INF/plugin.xml
```diff
@@ -334,6 +334,8 @@
     <action id="VimSearchWordBackward" class="com.maddyhome.idea.vim.action.motion.search.SearchWordBackwardAction" text="Search Again"/>
     <action id="VimGotoDeclaration" class="com.maddyhome.idea.vim.action.motion.search.GotoDeclarationAction" text="Goto Declaration"/>
 
+    <action id="VimSelectNextOccureance" class="com.maddyhome.idea.vim.action.motion.visual.VisualSelectNextOccurrence" text="Select Next Occurrence"/>
```
nhat: Should be `VimSelectNextOccurrence`

karavaevitalii: Thank you for contributing and valuable remark!

src/com/maddyhome/idea/vim/group/SearchGroup.java
```diff
@@ -625,20 +617,17 @@ public static TextRange findNext(@NotNull Editor editor, @NotNull String pattern
...
+      final int d2 = distance(r2, offset, forwards, size);
+      if (d1 < 0 && d2 >= 0) {
+        return Integer.MAX_VALUE;
       }
+      return d2 - d1;
     });
     if (!Options.getInstance().isSet("wrapscan")) {
       final int start = max.getStartOffset();
-      if (forwards && start < offset || start >= offset) {
+      if (forwards || start >= offset) {
```
vlasovskikh: This refactoring looks unclear. Let's keep the old version or split it into two ifs.

boedy:
I've been waiting desperately for this feature. I just tried this out and on first glance it works just as I expected it to 🙌 ! Thank you @karavaevitalii and @vlasovskikh for your work! 💯

PR #154, opened 2018-08-15
Fix for VIM-1289 interaction bar hides last line...: scroll one more …

catskul:
scroll one more line so that the match is not hidden.

In theory this could be put in scrollPositionIntoView, but wasn't sure if we'd want this behavior all the time.

Caveat that I'm not a Java developer so I'm not sure if there's a terser way of adding one line to the position as when I tried to modify the position itself was told by the compiler that it's 'final'.


catskul:
Fixes: https://youtrack.jetbrains.com/issue/VIM-1289

vlasovskikh:
@catskul Thanks for your PR! I'm sorry I still haven't time to review and merge it. I'll get back to your PR during my next iteration on IdeaVim. There is no need to resolve this merge conflict, I can do it myself during rebasing.

AlexPl292:
Hi @catskul!
Thank you for your pull request, but I'm afraid this is not the solution we are looking for. Your solution is based on row height, but there are several options to change row height in intellij idea (for example, change font size or line spacing). As long as interaction bar height is not bound to there properties, it cannot be guaranteed that your solution will work correctly.
It looks like there must be a more difficult solution for this.

PR #161, opened 2018-09-24
VIM-1473 Fixed incorrect selection type for unnamed registers

thecodewarrior:
While compiling [test cases](https://gist.github.com/thecodewarrior/7d766a97c377090470497c8af8e29527) for [VIM-1473](https://youtrack.jetbrains.com/issue/VIM-1473) I discovered that the unnamed registers (aka system clipboard) weren't retaining the selection type when yanked into and were making untrue assumptions as to what text copied in visual block mode would look like. 

In this PR I add a condition so the unnamed registers won't be overwritten unless the clipboard changes (thus they will retain the selection mode IDEAVim initially assigned them until someone copies different text) and removed visual block selections from guess function, as I can't think of any situation in which one would copy text outside of IDEA and want to paste it as if it were copied in visual block selection mode.

I have opened [IJSDK-475](https://youtrack.jetbrains.com/issue/IJSDK-475) regarding the misleading behavior of `StringUtil.splitByLines` as well.

### DIFF
--- a/src/com/maddyhome/idea/vim/group/RegisterGroup.java
+++ b/src/com/maddyhome/idea/vim/group/RegisterGroup.java
@@ -22,8 +22,6 @@ import com.google.common.collect.ImmutableList;
 import com.intellij.openapi.actionSystem.AnAction;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.editor.Editor;
-import com.intellij.openapi.util.text.StringUtil;
-import com.intellij.util.containers.ContainerUtil;
 import com.maddyhome.idea.vim.VimPlugin;
 import com.maddyhome.idea.vim.action.motion.mark.MotionGotoFileMarkAction;
 import com.maddyhome.idea.vim.action.motion.search.SearchAgainNextAction;
@@ -433,7 +431,11 @@ public class RegisterGroup {
   @Nullable
   private Register refreshClipboardRegister(char r) {
     final String text = ClipboardHandler.getClipboardText();
+    final Register currentRegister = registers.get(r);
     if (text != null) {
+      if (currentRegister != null && text.equals(currentRegister.getText())) {
+        return currentRegister;
+      }
       return new Register(r, guessSelectionType(text), text);
     }
     return null;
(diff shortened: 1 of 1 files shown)


AlexPl292:
HI @thecodewarrior !
Thank you for PR! This is a great investigation!

Well, I like your approach, but it doesn't really work.
The copied text is not put into `registers` because of [this](https://github.com/JetBrains/ideavim/blob/master/src/com/maddyhome/idea/vim/group/RegisterGroup.java#L182) line. So, `registers.get("+");` is always `null`. Could you please add missing put to `registers`?

Also, that's would be nice to see a couple of tests for this case. This is not necessary to set `clipboard=unnamedplus`, `"+y` and `"+p` are enough to reproduce this issue. Could you please check that everything works fine for character / line / block yanking to unnamed registers?

thecodewarrior:
"Waiting for reply"
Seriously? Don't talk to me about "waiting for reply." Try waiting ten months for a reply, only to get "cool! Please fix it and make tests." Do I have to jump on it and fix those issues _right now_ or risk waiting another year for the next reply? 

I submitted an issue to YouTrack about this issue and over the course of three months assembled a list of test cases for when it happens, used those to figure out why it happens, and eventually gave up on ever getting a reply and decided that maybe it would get noticed if I fixed it myself. So I did, I created a fork, fixed the issue, and created a pull request. Over half a year after I create the PR and almost a full year after I first submitted this frankly trivial bug I finally get someone to reply. What do they say! Will it finally be merged in so it can be fixed!?

"Sorry it doesn't work because of an issue on this line, and can you please add tests for it? Thanks."
image

I would say that you can expect to wait about a year before I reply with a fix, but that wouldn't matter to you and would only make the eventual fix take an extra year on top of the year it'll take to get my next reply, which would put the fix somewhere around the tail end of 2020 or the start of 2021. That is unless you find another bug in two year's time, in which case we might be able to get this merged before climate change wipes us out.

AlexPl292:
For a long time now, IdeaVim has been worked on and developed as a 20% project. During this time, it has accumulated quite a lot of problems. Since I've been using IdeaVim for several years now, and all these problems are annoying to me too, I have asked a couple of months ago to work on this project full time, and now I am actively trying to solve all these problems.

Unfortunately, we can't accept this PR just because it doesn't change anything. I pointed you to the line that causes the `+` and `*` registers not to be added to the `registers` list. And your new `if` will just always be skipped. This is the reason why I asked you to write some tests. It would be useful not only for the project as a whole but also for finding the problems of this change right away.

thecodewarrior:
Sorry, I got a bit salty. I'll see about adding the tests.

AlexPl292:
Hi!
We ended up merge your changes because we're working on VIM-25 and your changes would be useful there. Thank you for PR!

PR #162, opened 2018-10-01
VIM-798 Arrow keys for window navigation commands

AlexPl292:
Add new mappings to allow window navigation with arrow keys:
- ^W-left as ^W-h
- ^W-down as ^W-j
- ^W-up as ^W-k
- ^W-right as ^W-l

`package-info.java` already contains links to actions and doesn't require editing.


vlasovskikh:
@AlexPl292 Thanks for your pull request!

PR #168, opened 2019-01-02
Add whichwrap option

brtsos:
I'm not Java developer so please refactoring my code ;)


AlexPl292:
Hi @brtsos !
Thanks for the PR, but I'm afraid we can't merge it. `whichwrap` option [accepts much more complicated values](http://vimdoc.sourceforge.net/htmldoc/options.html#'whichwrap') than just `true/false`. Most likely, this option should be implemented somewhere else in the code. Besides, this code looks rather complicated and has no tests at all.

PR #170, opened 2019-01-15
VIM-1187: Fix performance issue with relative line numbers + large files

citizenmatt:
Fixes [VIM-1187](https://youtrack.jetbrains.com/issue/VIM-1187), which is a performance issue with large files and the `relativenumber` option. The fix uses the alternative API as specified in [this comment](https://youtrack.jetbrains.com/issue/VIM-1187#focus=streamItem-27-3057288-0-0). However, this also means bumping required IJ version to 2018.3.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-ideaVersion IC-2018.2.5
+ideaVersion IC-2018.3
 downloadIdeaSources true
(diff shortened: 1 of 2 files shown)


vlasovskikh:
@citizenmatt Thanks for your PR!

PR #172, opened 2019-01-21
VIM-1569: Strip tag attributes from closing tag for Vim-Surround

TonyArra:
Fixes [VIM-1569](https://youtrack.jetbrains.com/issue/VIM-1569) - Surrounding visually selected text adds class attribute to end tag

There appears to be an additional issue - that the selected text is not surrounded by line breaks in Visual-Line mode.


AlexPl292:
Thank you for the fix!

PR #175, opened 2019-02-07
VIM-1303 Update UI when look and feel changes

citizenmatt:
Causes the command bar and the output panel to redraw with the correct theme colours when the look and feel is changed.

Fixes [VIM-1303](https://youtrack.jetbrains.com/issue/VIM-1303)


AlexPl292:
Please check this PR for only "colour scheme" switch (not entire look and feel).

PR #177, opened 2019-02-16
Fixed Bug: VIM-1569 - Surrounding visually selected text with tag + attributes

LhKipp:
Fixed the bug :O
Added a test to the test file.

Note: 
Typing VS<div class = "container"> while on e.g. "Hello" will result in 
<div class = "container">Hello
</div>
and not in:
<div class="container">
<p>Hello txt</p>
</div>
as it is in normal vim.
But inserting a LF (\n) after "container"> results in failed tests.
I think inserting a LF at the end depends on the tag, which is to be inserted?

I hope i didn't do to much wrong. First pull request :)

### DIFF
--- a/src/com/maddyhome/idea/vim/extension/surround/VimSurroundExtension.java
+++ b/src/com/maddyhome/idea/vim/extension/surround/VimSurroundExtension.java
@@ -106,17 +106,19 @@ public class VimSurroundExtension extends VimNonDisposableExtension {
     }
   }
 
-  @Nullable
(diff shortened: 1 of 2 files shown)


LhKipp:
I am sorry. I havent seen, that there is already a pull request, which fixes this issue. And furthermore not noticing that this editor understands html... :|

AlexPl292:
Yes, sorry, but this PR was earlier: https://github.com/JetBrains/ideavim/pull/172

PR #178, opened 2019-02-21
Fix findBlockRange

sumoooru2:
In insert mode, "a(" to
`((int) nu<caret>m)`
should cover the whole block, but it covers only `(int)`.
I fixed the bug.


AlexPl292:
Does it relate to any https://youtrack.jetbrains.com/issues/VIM issue?

sumoooru2:
I searched for it and found https://youtrack.jetbrains.com/issue/VIM-907.

sumoooru2:
I added a test for the functionality.

AlexPl292:
Than you for PR!

PR #181, opened 2019-03-21
VIM-1105 Adding support for the command `command` (aliases).

ECourant:
This adds support for the command `command`.
Allowing users to add their own aliases for certain commands.

For example:
```vim
command WQ wq
```
Will allow `WQ` to be used as an alias for `wq`.

Documentation: http://vimdoc.sourceforge.net/htmldoc/map.html#:command

- [x] `:command {cmd}` support.
- [x] `!` flag.
- [x] `:com[mand]` support (shorter version).
- [x] Listing added aliases.
- [x] `:delc[ommand]` support.
- [x] Do not allow defining ":X", ":Next" and ":Print"
- [x] Tests

Please give me some feedback. 

Fixes [VIM-1105](https://youtrack.jetbrains.com/issue/VIM-1105)

Example:
ideavimcommand

### DIFF
--- a/src/com/maddyhome/idea/vim/ex/CommandParser.java
+++ b/src/com/maddyhome/idea/vim/ex/CommandParser.java
@@ -40,6 +40,7 @@ import java.util.regex.Pattern;
  * executes Ex commands entered by the user.
  */
 public class CommandParser {
+  public static final int MAX_RECURSION = 100;
   public static final int RES_EMPTY = 1;
   public static final int RES_ERROR = 1;
   public static final int RES_READONLY = 1;
@@ -73,7 +74,9 @@ public class CommandParser {
     new ActionListHandler();
     new AsciiHandler();
     new CmdFilterHandler();
+    new CmdHandler();
     new CopyTextHandler();
+    new DelCmdHandler();
     new DeleteLinesHandler();
     new DigraphHandler();
     new DumpLineHandler();
@@ -164,14 +167,51 @@ public class CommandParser {
    */
   public int processCommand(@NotNull Editor editor, @NotNull DataContext context, @NotNull String cmd,
                             int count) throws ExException {
+    return processCommand(editor, context, cmd, count, MAX_RECURSION);
+  }
+
+  /**
+   * Parse and execute an Ex command entered by the user
+   *
+   * @param editor  The editor to run the command in
+   * @param context The data context
+   * @param cmd     The text entered by the user
+   * @param count   The count entered before the colon
+   * @param aliasCountdown A countdown for the depth of alias recursion that is allowed
+   * @return A bitwise collection of flags, if any, from the result of running the command.
+   * @throws ExException if any part of the command is invalid or unknown
+   */
+  private int processCommand(@NotNull Editor editor, @NotNull DataContext context, @NotNull String cmd,
+                             int count, int aliasCountdown) throws ExException {
     // Nothing entered
     int result = 0;
     if (cmd.length() == 0) {
       return result | RES_EMPTY;
     }
 
-    // Save the command history
-    VimPlugin.getHistory().addEntry(HistoryGroup.COMMAND, cmd);
+    // Only save the command to the history if it is at the top of the stack.
+    // We don't want to save the aliases that will be executed, only the actual
+    // user input.
+    if (aliasCountdown == MAX_RECURSION) {
+      // Save the command history
+      VimPlugin.getHistory().addEntry(HistoryGroup.COMMAND, cmd);
+    }
+
+    // If there is a command alias for the entered text, then process the alias and return that
+    // instead of the original command.
+    if (VimPlugin.getCommand().isAlias(cmd)) {
+      if (aliasCountdown > 0) {
+        String commandAlias = VimPlugin.getCommand().getAliasCommand(cmd, count);
+        if (commandAlias.isEmpty()) {
+          return result |= RES_ERROR;
+        }
+        return processCommand(editor, context, commandAlias, count, aliasCountdown - 1);
+      } else {
+        VimPlugin.showMessage("Recursion detected, maximum alias depth reached.");
+        VimPlugin.indicateError();
+        return result |= RES_ERROR;
+      }
+    }
 
     // Parse the command
     final ExCommand command = parse(cmd);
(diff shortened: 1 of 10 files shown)


src/com/maddyhome/idea/vim/ex/CommandParser.java
```diff
@@ -170,6 +172,14 @@ public int processCommand(@NotNull Editor editor, @NotNull DataContext context,
       return result | RES_EMPTY;
     }
 
+    // If there is a command alias for the entered text, then process the alias and return that
+    // instead of the original command. We want to do this before the history and then log the
+    // resulting command to the history instead of the alias. This is a bit weird, but it
+    // prevents it from doubling up.
+    if (VimPlugin.getCommand().hasAlias(cmd)) {
```
AlexPl292: Please add a mechanism to avoid infinite recursion!

src/com/maddyhome/idea/vim/ex/handler/CmdHandler.kt
```diff
@@ -0,0 +1,78 @@
...
+        // We want to get the first character sequence in the arguments.
+        // eg. command! Wq wq
+        // We want to extract the Wq only, and then just use the rest of
+        // the argument as the alias result.
+        val alias = argument.split(" ")[0]
+        argument = argument.removePrefix(alias).trim()
+
+        // User-aliases need to begin with an uppercase character.
+        if (!alias[0].isUpperCase()) {
+            return false
```
AlexPl292: Do not leave the user without understanding what is wrong. Use `com.maddyhome.idea.vim.VimPlugin#showMessage`

ECourant:
I will do some research on `-nargs` and potentially add it, but initially I had not planned on it.

I will add tests and black list those variables so that this won't overwrite them. 

Thank you very much for the feedback!

src/com/maddyhome/idea/vim/ex/handler/CmdHandler.kt
ECourant: I also added messages in other spots too, matching the messages returned from Vim for the same scenarios.

ECourant:
I've added some tests, the tests cover 93% of lines in `CmdHandler.kt` and 100% of lines in `DelCmdHandler.kt`. I need to fix the infinite recursion thing still, and do you want the state of commands to save? It makes sense to me that any aliases added via `ideavimrc.vim` would be enough, but what is your preference.

src/com/maddyhome/idea/vim/ex/CommandParser.java
ECourant: I just made it so that a command alias cannot reference another alias. If you try to execute an alias that references another alias then an error is returned. This might not be the best way, but the other way that I would approach this would be to set a max recurussion threshold and only return an error if that threshold gets reached. If you have a preference for the threshold then I can rewrite it.

src/com/maddyhome/idea/vim/package-info.java
```diff
@@ -602,6 +602,8 @@
  * |:quitall|             {@link com.maddyhome.idea.vim.ex.handler.ExitHandler}
  * |:wqall|               {@link com.maddyhome.idea.vim.ex.handler.ExitHandler}
  * |:xall|                {@link com.maddyhome.idea.vim.ex.handler.ExitHandler}
+ * |:command|             {@link com.maddyhome.idea.vim.ex.handler.CmdHandler}
+ * |:command|             {@link com.maddyhome.idea.vim.ex.handler.DelCmdHandler}
```
AlexPl292: :delcommand you mean?

src/com/maddyhome/idea/vim/ex/CommandParser.java
```diff
@@ -163,13 +165,24 @@ public boolean processLastCommand(@NotNull Editor editor, @NotNull DataContext c
...
     // Nothing entered
     int result = 0;
     if (cmd.length() == 0) {
       return result | RES_EMPTY;
     }
 
+    // If there is a command alias for the entered text, then process the alias and return that
+    // instead of the original command. We also only want to do this if the previous command
+    // wasn't already an alias. Basically, an alias cannot reference an alias. (Just to be safe)
+    if (VimPlugin.getCommand().hasAlias(cmd)) {
```
AlexPl292: Weeell... Adding a new boolean to arguments is always a sign of "think a bit more" for me :)
And as I can see, vim allows you to create aliases with existing aliases.

Probably you can use [this](https://stackoverflow.com/a/11300253/3124227) approach?
The first method would be untouched
```
public int processCommand(@NotNull Editor editor, @NotNull DataContext context, @NotNull String cmd,
                            int count) {
    return processCommand(editor, context, cmd, count, 100);
}
```

and the second one with actual logic
```
private int processCommand(@NotNull Editor editor, @NotNull DataContext context, @NotNull String cmd,
                            int count, int aliasCountdown) {
    ...
    if (hasAlias) {
        return processCommand(editor, context, cmd, count, aliasCountdown - 1);
    }
    ...
}
```

src/com/maddyhome/idea/vim/ex/CommandParser.java
```diff
@@ -163,13 +165,24 @@ public boolean processLastCommand(@NotNull Editor editor, @NotNull DataContext c
...
+      if (!isAlias) {
+        return processCommand(editor, context, VimPlugin.getCommand().getAlias(cmd), count, true);
+      } else {
+        throw new InvalidCommandException("Command alias cannot reference another command alias.", cmd);
```
AlexPl292: It's better to show a message as original vim does
``` Vim.showMessage("Recursion detected...")```

src/com/maddyhome/idea/vim/ex/CommandParser.java
ECourant: I like this much more, basically limits the nesting to 100 aliases.

ECourant: I added a limit based on your suggestion. I still need to add tests though to make sure it works.

ECourant:
Looking at `-nargs` I do like the idea of having that. But I think I will add that in a separate pull request, it might take me a bit to make that work, and having basic command aliases would be really nice to have sooner.

AlexPl292:
Actually, we had already very bad feedback about uncompleted features. So, at the moment we are trying to support commands as much as possible before the release.
I think that for now, I can approve changes, but without merging it into master. And you will have some time to decide what to do with `-nargs`. In any case, we are not planning to make any releases yet.

ECourant:
Sounds good, I will work on `-nargs` a bit this week/weekend and add that. Thank you very much Alex!

ECourant:
I've implemented nargs handling based on documentation and experiments with vim itself. This code should be reviewed again @AlexPl292 due to the changes I've made. 

Supported symbols at the moment are `<lt>` `<count>` `<args>` `<q-args>` and `<f-args>`. Granted I don't know if `<f-args>` has an application yet in IdeaVim. 

I've also changed the format of the list alias output to look more like Vim's actual output and include the number of arguments.

AlexPl292:
I'm sorry, I've updated `flags` method. That's why the checks failed.

ECourant:
I should be able to just rebase onto master then fix it.

ECourant:
Mmmmmm, I seem to have messed up rebasing my branch/fork. I'm going to recreate this pull request on a clean branch. That seems easier.

PR #182, opened 2019-03-27
VIM-1615 Fix handling of so=999

citizenmatt:
This PR will fix handling of `scrolloff` being larger than half the height of the screen, and will keep the current line centred in the screen.

- [x] Fix issue which caused the position of the current line to shift up and down when pressing `h` and `l` ([VIM-1615](https://youtrack.jetbrains.com/issue/VIM-1615)) 
- [x] Fix centring of current line when `scrolloff` is larger than half the height of the screen but less than the full height
- [x] Fix centring of current line when block inlays are enabled (e.g. Rider)
- [x] Keeps the bottom line aligned to the bottom of the screen when scrolling text up (e.g. `j` at the bottom of the screen) and the top line aligned to the top when scrolling text down (`k` at top of screen)

### DIFF
--- a/src/com/maddyhome/idea/vim/group/MotionGroup.java
+++ b/src/com/maddyhome/idea/vim/group/MotionGroup.java
@@ -910,10 +910,18 @@ public class MotionGroup {
   }
 
   public boolean scrollLine(@NotNull Editor editor, int lines) {
-    int visualLine = EditorHelper.getVisualLineAtTopOfScreen(editor);
+    assert lines != 0 : "lines cannot be 0";
 
-    visualLine = EditorHelper.normalizeVisualLine(editor, visualLine + lines);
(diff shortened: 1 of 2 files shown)


AlexPl292:
Hi @citizenmatt !
Thank you for PR! I've checked it and it seems to be good. The only thing is that the caret does not always behave correctly with `so = 0` and enabled inlays. This is of course not so critical as jumping `so=999`, but could you take a look at it? (`KeyHandler.java` from IdeaVim on this gif)
Screen Recording 2019-03-29 at 12 09 34

citizenmatt:
The original behaviour was to try and always keep the top line aligned with the top of the screen, even when the caret was at the bottom of the screen and text was scrolling up. I've changed the behaviour so that now, when scrolling text up, the bottom line is aligned with the bottom of the screen, and when scrolling text down, the top line is aligned with the top of the screen.

PR #183, opened 2019-03-30
Implement TabOnlyHandler

danielemegna:
Implement `:tabonly` command as described in [vim tab tips page](https://vim.fandom.com/wiki/Using_tab_pages).

_____

My congrats to maintainers, I found this repo very tidy and code very easy to browse and understand. :clap: 

This pull request has been developed during [Open Source Saturday #23](https://www.meetup.com/it-IT/Open-Source-Saturday-Milano/events/259527969/) meetup in Milan.

### DIFF
--- a/src/com/maddyhome/idea/vim/ex/handler/TabOnlyHandler.kt
+++ b/src/com/maddyhome/idea/vim/ex/handler/TabOnlyHandler.kt
@@ -1,0 +1,39 @@
+/*
+ * IdeaVim - Vim emulator for IDEs based on the IntelliJ platform
+ * Copyright (C) 2003-2019 The IdeaVim authors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.maddyhome.idea.vim.ex.handler
+
+import com.intellij.openapi.actionSystem.DataContext
+import com.intellij.openapi.editor.Editor
+import com.maddyhome.idea.vim.VimPlugin
+import com.maddyhome.idea.vim.ex.CommandHandler
+import com.maddyhome.idea.vim.ex.CommandHandler.Flag.RANGE_OPTIONAL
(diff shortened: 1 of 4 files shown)


src/com/maddyhome/idea/vim/ex/handler/TabOnlyHandler.kt
```diff
@@ -0,0 +1,39 @@
...
+import com.maddyhome.idea.vim.ex.CommandHandler
+import com.maddyhome.idea.vim.ex.CommandHandler.Flag.ARGUMENT_OPTIONAL
+import com.maddyhome.idea.vim.ex.CommandHandler.Flag.DONT_REOPEN
+import com.maddyhome.idea.vim.ex.ExCommand
+import com.maddyhome.idea.vim.ex.commands
+import com.maddyhome.idea.vim.ex.flags
+
+class TabOnlyHandler : CommandHandler(
+        commands("tabo[nly]"),
+        flags(ARGUMENT_OPTIONAL, DONT_REOPEN)
```
AlexPl292: `ARGUMENT_OPTIONAL` - Does `tabonly` command take any argument? From the docs I can see that syntax of this command looks like `:{count}tabo[nly][!]` and there is no optional argument here. Or do I miss something?

And could you please add `RANGE_OPTIONAL` and `RANGE_IS_COUNT` flags? These flags are exactly for cases when command starts with `:{count}...` (even if you don't use this {count}).

danielemegna:
Hello Alex! Thank you, I'm sorry for the delay.

You're absolutely right, I didn't take care of the flags.
Does it look right now?

src/com/maddyhome/idea/vim/ex/handler/TabOnlyHandler.kt
danielemegna: > :{count}tabo[nly][!]

Where did you see the `{count}` optional parameter? I cannot find it in the docs:

- http://vimdoc.sourceforge.net/htmldoc/tabpage.html
- https://vim.fandom.com/wiki/Using_tab_pages

I always use this command in vim without params anyway :-)

danielemegna:
> :{count}tabo[nly][!]

Where did you see the `{count}` optional parameter? I cannot find it in the docs:

- http://vimdoc.sourceforge.net/htmldoc/tabpage.html
- https://vim.fandom.com/wiki/Using_tab_pages

I always use this command in vim without params anyway :-)

AlexPl292:
> Where did you see the {count} optional parameter? I cannot find it in the docs:
Because you use the documentation for 7.3 version :)
https://github.com/vim/vim/blob/master/runtime/doc/tabpage.txt#L159

> Does it look right now?
In general, I meant that `ARGUMENT_OPTIONAL` should be replaced by `ARGUMENT_FORBIDDEN`. But, it seems, it is our problem that developers do not understand what flags should be put. I think I will merge your pull request and make the API more understandable.

AlexPl292:
Related issue: https://youtrack.jetbrains.com/issue/VIM-1630

danielemegna:
Awesome! Thank you Alex!

PR #186, opened 2019-04-08
[Open for test] VIM-510 - IdeaVim selection model

AlexPl292:
This huge pull request was created as a result of [VIM-510](https://youtrack.jetbrains.com/issue/VIM-510). It provides many changes aiming to solve the problems associated with selection.

Main changes:
- Introducing [select mode](https://vimhelp.org/visual.txt.html#Select-mode).
- [[VIM-510](https://youtrack.jetbrains.com/issue/VIM-510)] Structural selection enables visual mode.
- [[VIM-274](https://youtrack.jetbrains.com/issue/VIM-274)] During refactoring, select mode is enabled (which works almost like normal IntelliJ IDEA selection).

The list of solved bugs: https://youtrack.jetbrains.com/issues/VIM?q=Project:%20VIM%20tag:%20vim510%20sort%20by:%20votes

These changes are being tested. If you are interested in trying them, install the plugin from this branch, or use the attached file. Any feedback is highly appreciated!

**Current version**
[IdeaVIM-SNAPSHOT-v5-[14-05-2019]-RC.zip](https://github.com/JetBrains/ideavim/files/3179197/IdeaVIM-SNAPSHOT.zip)

**Older versions**
[IdeaVIM-SNAPSHOT-v4-[08-05-2019]-RC.zip](https://github.com/JetBrains/ideavim/files/3158377/IdeaVIM-SNAPSHOT.zip)
[IdeaVIM-SNAPSHOT-v3-[07-05-2019].zip](https://github.com/JetBrains/ideavim/files/3152702/IdeaVIM-SNAPSHOT.zip)
[IdeaVIM-SNAPSHOT-v2-[30-04-2019].zip](https://github.com/JetBrains/ideavim/files/3132139/IdeaVIM-SNAPSHOT.zip)
[IdeaVIM-SNAPSHOT-v1.zip](https://github.com/JetBrains/ideavim/files/3130585/IdeaVIM-SNAPSHOT.zip)


AlexPl292:
@ayzenquwe which problem do you mean? Any examples?

ayzenquwe:
I have this in my .ideavimrc:
`map <Leader>ev :action IntroduceVariable<CR>`

It works, but doesn't support selections. So if I selected something it will just do nothing.

AlexPl292:
@ayzenquwe Yeah, that's one of the problems we want to fix with this PR. `Add SAVE_VISUAL flag for commands` - we started with this commit, but this problem hasn't been completely fixed yet. That's why this PR is still a draft.

ayzenquwe:
Thanks for the info. Glad to hear that this problem will be fixed. :)

jrbrodie77:
Thanks for doing this.  I had submitted one of those bugs and it drives me nuts.  I did find a weird behavior.

If I'm working on an indented block (like the body of an if statement) and then I want to drop back to the unindented level by hitting backspace on a newline, the cursor switches from the insert mode | to the regular mode block cursor.  I'm still in insert mode, but to the user it looks like ESC has been hit.

(start in insert mode)

    if foo:
        print("baz")<cursor is now here>

(hit enter)

    if foo:
        print("baz")
        <cursor is now here, everything is still normal>

(hit backspace)

    if foo:
        print("baz")
    <cursor is now here, cursor is now block style>

I'm on Pycharm 2018.3, using the zip file included at the top of this thread for IdeaVim.

AlexPl292:
@jrbrodie77 Oh, yeah! It was reproduced right away. Thank you for the report!

What about your bug? Does it work well now?

marek2901:
Im using multiple-cursors and it seems like the screen doesn't follow the most recent selection. It's stuck on a first one instead. Apart from that it works better than before.

the flow I use is to visually select piece of text and hit alt/option + n

marek2901:
Hitting u (undo) multiple times quickly gets me back to insert mode

src/com/maddyhome/idea/vim/command/CommandFlags.kt
```diff
@@ -28,7 +28,7 @@ enum class CommandFlags {
     FLAG_MOT_INCLUSIVE,
     FLAG_MOT_EXCLUSIVE,
     /**
-     * Indicates that the cursor position should be saved prior to this motion command
+     * Indicates that the cursor position should be saved prior to this visual command
```
vlasovskikh: I believe it was an accidental change. It should remain "motion" here.

src/com/maddyhome/idea/vim/command/CommandState.java
```diff
@@ -86,9 +88,22 @@ public static boolean inVisualCharacterMode(@Nullable Editor editor) {
...
+
+  public static boolean inVisualLineMode(@Nullable Editor editor) {
+    final CommandState state = getInstance(editor);
+    return state.getMode() == Mode.VISUAL && state.getSubMode() == SubMode.VISUAL_LINE;
+  }
+
   public static boolean inVisualBlockMode(@Nullable Editor editor) {
     final CommandState state = getInstance(editor);
-    return state.getMode() == Mode.VISUAL && state.getSubMode() == SubMode.VISUAL_BLOCK;
+    return state.getSubMode() == SubMode.VISUAL_BLOCK;
```
vlasovskikh: It makes sense to check if the main mode is still visual. I see no reason to drop this check.

src/com/maddyhome/idea/vim/common/TextRange.java
```diff
@@ -20,6 +20,10 @@
 
 import org.jetbrains.annotations.NotNull;
 
+/**
+ * @deprecated Please use com.maddyhome.idea.vim.group.visual.VimSelection
+ */
+@Deprecated
```
vlasovskikh: Let's drop it completely if it's our class, it's deprecated, and it's used only in our code base.

src/com/maddyhome/idea/vim/ex/CommandHandler.kt
```diff
@@ -91,7 +91,15 @@ abstract class CommandHandler {
      * Indicates that this command does not modify the editor
      */
     READ_ONLY,
-    DONT_SAVE_LAST
+      DONT_SAVE_LAST,
```
vlasovskikh: There's something wrong with indents here.

src/com/maddyhome/idea/vim/ex/CommandHandler.kt
```diff
@@ -165,27 +173,31 @@ abstract class CommandHandler {
       throw NoArgumentAllowedException()
     }
     CommandState.getInstance(editor).flags = optFlags
+      if (CommandState.getInstance(editor).mode == CommandState.Mode.VISUAL && Flag.SAVE_VISUAL !in argFlags.flags) {
```
vlasovskikh: The indent level is wrong.

src/com/maddyhome/idea/vim/group/visual/VisualGroup.kt
```diff
@@ -0,0 +1,252 @@
...
+/**
+ * @author Alex Plate
+ */
+
+/**
+ * Set selection for caret
+ * This method doesn't change CommandState and operates only with caret and it's properties
+ * if [moveCaretToSelectionEnd] is true, caret movement to [end] will be performed
+ */
+fun Caret.vimSetSelection(start: Int, end: Int = start, moveCaretToSelectionEnd: Boolean = false) {
```
vlasovskikh: I think for methods it's better to follow the word order for English sentences: "set vim selection" rather than to use "vim" as a prefix. See also the methods below.

src/com/maddyhome/idea/vim/group/visual/VisualModeHelper.kt
```diff
@@ -0,0 +1,46 @@
...
+import com.maddyhome.idea.vim.listener.SelectionVimListenerSuppressor
+
+/**
+ * @author Alex Plate
+ */
+
+/**
+ * Set selection without calling SelectionListener
+ */
+fun SelectionModel.vimSetSystemSelectionSilently(start: Int, end: Int) =
```
vlasovskikh: Other helpers live in helper/, I'd suggest to move it there as well.

src/com/maddyhome/idea/vim/group/visual/VisualOperationChange.kt
```diff
@@ -0,0 +1,105 @@
...
+    companion object {
+        fun default(subMode: CommandState.SubMode) =
+                when (val type = SelectionType.fromSubMode(subMode)) {
+                    SelectionType.LINE_WISE, SelectionType.CHARACTER_WISE -> VisualChange(1, 1, type)
+                    SelectionType.BLOCK_WISE -> VisualChange(0, 1, type)
+                }
+    }
+}
+
+class VisualOperation {
```
vlasovskikh: It could be an object, it has no methods besides its companion object methods.

src/com/maddyhome/idea/vim/group/MotionGroup.java
```diff
@@ -118,257 +118,58 @@ public void turnOff() {
...
-
-  @Nullable
-  public TextRange getParagraphRange(@NotNull Editor editor, @NotNull Caret caret, int count, boolean isOuter) {
-    return SearchHelper.findParagraphRange(editor, caret, count, isOuter);
-  }
-
   /**
-   * This helper method calculates the complete range a motion will move over taking into account whether
-   * the motion is FLAG_MOT_LINEWISE or FLAG_MOT_CHARACTERWISE (FLAG_MOT_INCLUSIVE or FLAG_MOT_EXCLUSIVE).
+   * This helper method calculates the complete range a visual will move over taking into account whether
```
vlasovskikh: Here and everywhere else in this file I believe "motion" has been changed to "visual" accidentally. Could you please revert it?

src/com/maddyhome/idea/vim/helper/EditorHelper.java
```diff
@@ -609,7 +609,7 @@ else if (order == CaretOrder.DECREASING_OFFSET) {
    * Due to block inlays, the caret location is maintained as a scroll offset, rather than the number of lines from the
    * top of the screen. This means the line offset can change if the number of inlays above the caret changes during
    * scrolling. It also means that after scrolling, the top screen line isn't guaranteed to be aligned to the top of
-   * the screen, unlike most other motions ('M' is the only other motion that doesn't align the top line).
+   * the screen, unlike most other motions ('M' is the only other visual that doesn't align the top line).
```
vlasovskikh: It should be "motion".

src/com/maddyhome/idea/vim/key/ArgumentNode.java
```diff
@@ -29,7 +29,7 @@
 
 /**
  * This represents a command argument node in the key/action tree. Currently arguments of argType character
- * and motion command are used.
+ * and visual command are used.
```
vlasovskikh: "motion"

src/com/maddyhome/idea/vim/listener/IdeaSpecifics.kt
```diff
@@ -0,0 +1,101 @@
...
+import com.maddyhome.idea.vim.group.visual.moveCaretOneCharLeftFromSelectionEnd
+import com.maddyhome.idea.vim.helper.EditorDataContext
+
+/**
+ * @author Alex Plate
+ */
+object IdeaSpecifics {
+    fun addIdeaSpecificsListener(project: Project) {
+        val connection = project.messageBus.connect()
+        connection.subscribe(AnActionListener.TOPIC, VimActionListener)
```
vlasovskikh: Please move all IDE connections / registrations to `EventFacade` so we can easily trace all the events that IdeaVim receives from the IDE.

src/com/maddyhome/idea/vim/group/MotionGroup.java
AlexPl292: I've found the bad commit and reverted changes according to them

src/com/maddyhome/idea/vim/common/TextRange.java
AlexPl292: I've ended up with leaving both classes because `TextRange` has too many usages and it's removing can cause unwanted behavior.

reniar1:
Sorry can someone help me? I don't know exactly how to test this. Do I just have to replace the old files with the new files or what is the process of implementing these changes?

AlexPl292:
@reniar1 Hi! In order to test these changes you should install EAP version of plugin.
https://github.com/JetBrains/ideavim#get-an-early-access

reniar1:
Sorry again,

I don't understand this

Add https://plugins.jetbrains.com/plugins/eap/ideavim to "Settings | Plugins | Cog icon | Manage Plugin Repositories"

What is a COG icon?

PR #188, opened 2019-04-08
(VIM-1105) Adding support for command aliasing.

ECourant:
Commands `command`, `comclear` and `delcommand` have been added. `command` also
supports -nargs, allowing arguments to be passed to the aliased command.
Symbols supported for -nargs are: `<args> <q-args> <count> <lt>`.

For example:
```vim
command WQ wq
```
Will allow `WQ` to be used as an alias for `wq`.

Documentation: http://vimdoc.sourceforge.net/htmldoc/map.html#:command

- [x] `:command {cmd}` support.
- [x] `!` flag.
- [x] `:com[mand]` support (shorter version).
- [x] Listing added aliases.
- [x] `:delc[ommand]` support.
- [x] Do not allow defining ":X", ":Next" and ":Print"
- [x] Tests

Please give me some feedback. 

Fixes [VIM-1105](https://youtrack.jetbrains.com/issue/VIM-1105)

Example:
ideavimcommand

@AlexPl292 This was just easier than trying to fix my fork that I broke when I tried to rebase.

### DIFF
--- a/src/com/maddyhome/idea/vim/common/Alias.kt
+++ b/src/com/maddyhome/idea/vim/common/Alias.kt
@@ -1,0 +1,62 @@
+
+package com.maddyhome.idea.vim.common
+
+import com.maddyhome.idea.vim.VimPlugin
+
+/**
+ * @author Elliot Courant
+ */
+class Alias(
+        var minimumNumberOfArguments: Int,
+        var maximumNumberOfArguments: Int,
+        var name: String,
+        var command: String
+) {
+    private companion object {
+        const val LessThan = "<lt>"
+        const val Count = "<count>"
+        const val Arguments = "<args>"
+        const val QuotedArguments = "<q-args>"
+    }
+
+    fun getNumberOfArguments(): String {
+        if (this.minimumNumberOfArguments == 0 && this.maximumNumberOfArguments == 0) {
+            return "0" // No arguments
+        } else if (this.minimumNumberOfArguments == 0 && this.maximumNumberOfArguments == -1) {
+            return "*" // Any number of arguments
+        } else if (this.minimumNumberOfArguments == 0 && this.maximumNumberOfArguments == 1) {
+            return "?" // Zero or one argument
+        } else if (this.minimumNumberOfArguments == 1 && this.maximumNumberOfArguments == -1) {
+            return "+" // One or more arguments
+        }
+        return this.minimumNumberOfArguments.toString() // Specified number of arguments
+    }
+
+    fun getCommand(input: String, count: Int): String {
+        if (this.maximumNumberOfArguments == 0 && this.maximumNumberOfArguments == 0) {
+            return this.command
+        }
+        var compiledCommand = this.command
+        val cleanedInput = input.trim().removePrefix(name).trim()
+        if (minimumNumberOfArguments > 0 && cleanedInput.isEmpty()) {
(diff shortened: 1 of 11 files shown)


ECourant:
This also has the flag fix that you had mentioned.

src/com/maddyhome/idea/vim/ex/handler/CmdHandler.kt
```diff
@@ -0,0 +1,160 @@
...
+import com.maddyhome.idea.vim.VimPlugin
+import com.maddyhome.idea.vim.common.Alias
+import com.maddyhome.idea.vim.ex.*
+import com.maddyhome.idea.vim.ex.vimscript.VimScriptCommandHandler
+
+/**
+ * @author Elliot Courant
+ */
+class CmdHandler : CommandHandler(
+        commands("com[mand]", "command[-nargs]", "com[-nargs]"),
```
AlexPl292: > "command[-nargs]", "com[-nargs]"

Why do you need these mappings?
Moreover, it looks invalid. `command[-nargs]` - that means that `command`, `command-`, `command-n`, `command-na`, etc. are also allowed.

src/com/maddyhome/idea/vim/common/Alias.kt
```diff
@@ -0,0 +1,104 @@
...
+
+        // We want to escape <lt> after we've dropped in all of our args, if they are
+        // using <lt> its because they are escaping something that we don't want to handle
+        // yet.
+        compiledCommand = compiledCommand.replace(LessThan, "<")
+
+        return compiledCommand
+    }
+
+    private fun getFunctionArguments(input: String): Array<String> {
```
AlexPl292: Could you please add some tests for this logic?

src/com/maddyhome/idea/vim/ex/CommandParser.java
```diff
@@ -40,6 +40,7 @@
  * executes Ex commands entered by the user.
  */
 public class CommandParser {
+  public static final int MAX_RECURSION = 100;
```
AlexPl292: `public`?

src/com/maddyhome/idea/vim/common/Alias.kt
ECourant: I'm still working on this, it might be a few more days though, the escaping of certain characters is proving more difficult than my original approach.

ECourant:
Given that IdeaVim doesn't currently support function calls, would it make sense to remove `<f-args>` support for now given that it's causing some problems. Then I can just resolve the build errors that are happening in the merge?

And in the future when IdeaVim supports functions I can add `<f-args>` then once I have better knowledge and understanding.

AlexPl292:
Oh, yeah. Sure, no problem.

ECourant:
@AlexPl292 Thank you for the feedback with this PR. I've remove `<f-args>` for now and squashed my commits. All my tests passed locally so it should be good to go! 👍

AlexPl292:
@ECourant great news!  I've updated your code a bit to make it more "kotlinable".

A big thank for contributing such a big command! This is really marvellous!

AlexPl292:
I see the people start to use `command` command.
http://karllorey.com/blog/how-to-set-up-black-with-debian-pycharm-ideavim/

PR #179, opened 2019-02-24
Fix VIM-1090 and VIM-1100 tag motion with duplicate tags.

Vvalter:
The current implementation of the tag motion does not handle multiple tags with the same name correctly. I was not able to adapt the current implementation. Instead I reimplemented the whole algorithm, trying to stay as close to the behavior of vim as possible. The runtime complexity is also improved as it should be at most linear in the size of the file instead of quadratic as before. (Try executing `dat` at the end of a file with 10000 `<a>` followed by 10000 `</a>` and IdeaVim will freeze with the current implementation.)

This pull request fixes [VIM-1090](https://youtrack.jetbrains.com/issue/VIM-1090) and [VIM-1100](https://youtrack.jetbrains.com/issue/VIM-1100). Possibly also [VIM-1155](https://youtrack.jetbrains.com/issue/VIM-1155), but the description of this issue is not clear enough to reproduce it.

PR #189, opened 2019-04-13
VIM-1633 - fix block selection

Nimamoh:
PR about [VIM-1633](https://youtrack.jetbrains.com/issue/VIM-1633). It should fix operations about block selection. Feel free to test and ask


src/com/maddyhome/idea/vim/helper/SearchHelper.java
```diff
@@ -113,15 +113,25 @@ public static TextRange findBlockRange(@NotNull Editor editor, @NotNull Caret ca
                                          boolean isOuter) {
     CharSequence chars = editor.getDocument().getCharsSequence();
     int pos = caret.getOffset();
-    int start = caret.getSelectionStart();
-    int end = caret.getSelectionEnd();
-    if (start != end) {
-      pos = Math.min(start, end);
-    }
+    int start = Math.min(caret.getSelectionStart(), caret.getSelectionEnd());
+    int end = Math.max(caret.getSelectionStart(), caret.getSelectionEnd());
```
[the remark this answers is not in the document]
Nimamoh: You're right, a little bit too cautious there.

src/com/maddyhome/idea/vim/helper/SearchHelper.java
```diff
@@ -113,15 +113,25 @@ public static TextRange findBlockRange(@NotNull Editor editor, @NotNull Caret ca
...
 
     int loc = blockChars.indexOf(type);
     char close = blockChars.charAt(loc + 1);
 
+    if (!isOuter
```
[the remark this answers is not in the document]
Nimamoh: see next commit

Nimamoh:
I added some comments, let me know if something isn't clear enough.

I also noted that tag selection suffers the same issue.
After a quick look it seems the problem is very similar, the solution would also be ( probably).
I let you choose if you want to keep this PR about block selection (and merge sooner) or if you wanna wait for tag selection fix eventually.

AlexPl292:
There's no reason at this time for us to merge the PR as soon as possible. So, if you want, you can also add fix for tags (and, of course, we'll be very happy if you do that). As for the current change, we are ready to merge them.

src/com/maddyhome/idea/vim/helper/TagBlockHelper.java
```diff
@@ -0,0 +1,215 @@
+package com.maddyhome.idea.vim.helper;
+
+import com.intellij.lexer.Lexer;
+import com.intellij.lexer.LexerPosition;
+import com.intellij.lexer.XmlLexer;
```
Nimamoh: I use the jetbrains XML lexer to parse tags. I assumed that it would be no problem but idk what is the politics about jetbrains platform dependencies.

test/org/jetbrains/plugins/ideavim/action/MotionActionTest.java
```diff
@@ -510,7 +510,7 @@ public void testDeleteInnerTagBlockBetween() {
   //|d| |v_it|
   public void testDeleteInnerTagBlockBetweenTagWithRegex() {
     typeTextInFile(parseKeys("dit"), "abcde<[abc]*>af<caret>gbc</[abc]*>hi");
-    myFixture.checkResult("abcde<[abc]*></[abc]*>hi");
+    myFixture.checkResult("abcde<[abc]*>afgbc</[abc]*>hi");
```
Nimamoh: I changed the test in accordance to the behavior I get in vanilla vim.
IMO the closest ideavim behave like vanilla vim the better.

test/org/jetbrains/plugins/ideavim/action/MotionActionTest.java
```diff
@@ -630,7 +630,7 @@ public void testDeleteInnerTagBlockInCloseWrongOrder() {
   //|d| |v_it|
   public void testDeleteInnerTagBlockTwoTagsWrongOrder() {
     typeTextInFile(parseKeys("dit"), "<foo><html>t<caret>ext</foo></html>");
-    myFixture.checkResult("<foo></foo></html>");
+    myFixture.checkResult("<foo><html></html>");
```
Nimamoh: see review above

AlexPl292:
Oh oh! I completely forgot to warn you about https://github.com/JetBrains/ideavim/pull/179. Okay, I'll review your PR right now and give you feedback asap. Anyway, everything's fine, that pull request needs improvement.

Nimamoh:
Oh that's too bad I didn't see #179 either.
After playing a little bit with #179 I'd compare the two PR like this:
- this PR is weaker performance-wise (although I can think about a few things to improve the performances).
- this PR goes further in emulating correct vi behavior (successive tag block selection especially, same behavior on edge cases).

Vvalter:
I am the author of #179. I actually chose to use the regex approach instead of a parser to keep the behavior closer to vanilla vim, because it uses regex as well.
I added the following two test cases in my pull request where your approach seems to work incorrectly:

```
  // VIM-1090 |d| |v_at|
  public void testDeleteOuterTagDuplicateTags() {
    typeTextInFile(parseKeys("dat"), "<a><a></a></a<caret>>");
    myFixture.checkResult("");
  }
  // VIM-1090 |d| |v_it|
  // Adapted from vim source file "test_textobjects.vim"
  public void testDeleteInnerTagDuplicateTags() {
    typeTextInFile(parseKeys("dit"), "<b>as<caret>d<i>as<b />df</i>asdf</b>");
    myFixture.checkResult("<b></b>");
  }
```

Nimamoh:
thanks for your feedback @Vvalter , 

I cannot reproduce any issue with the first test, although the second one has to do with self closing tag that were not handled in the parser. added a fix.

Vvalter:
It is strange that you are unable to reproduce the first test. If I run your code (after a clean checkout) and type dat at `<a><a></a></a<caret>>` the result is `<a>`, but it should delete everything. Are you sure you are unable to reproduce this? Does `testDeleteOuterTagDuplicateTags` pass if you run it? Do you maybe have uncommited changes?

test/org/jetbrains/plugins/ideavim/action/MotionActionTest.java
```diff
@@ -786,6 +893,172 @@ public void testFileStartsWithSlash() {
     assertPluginError(true);
   }
 
+  // VIM-1633, from VIM-1090 |d| |v_at|
+  public void testDeleteOuterTagDuplicateTags() {
```
Nimamoh: @Vvalter

Nimamoh:
I don't get it, I just did a 
```
git clone git@github.com:Nimamoh/ideavim.git --branch VIM-1633
cd ideavim
./gradlew clean test
```
Tests are passing, I echoed you in the corresponding test in source code. then with
```
./gradlew runIde
```
created a scratch xml file, paste
```
<a><a></a></a>
```
running dat on the last '>' character erased everything. Did the same with a scratch plain text file.


Are you sure you pulled the VIM-1633 branch? I see you use master on your repo, my changes are in the VIM-1633 branch.

test/org/jetbrains/plugins/ideavim/action/MotionActionTest.java
```diff
@@ -786,6 +893,172 @@ public void testFileStartsWithSlash() {
...
+    typeTextInFile(parseKeys("dat"), "<a><a></a></a<caret>>");
+    myFixture.checkResult("");
+  }
```
Nimamoh: @Vvalter

Vvalter:
Oh I am sorry I forgot to checkout the proper branch - my bad.

I found another issue:
`<a "> </a>`  is not recognized as a tag pair. (Interestingly it works if you add another `</a>`) " can be replaced by other punctuation marks like ;,.'. This probably stems from the fact that you are parsing valid xml, but vim simply uses a regex to match stuff between angle brackets? Of course it is questionable whether this is a problem, because most of the times people will be working with proper xml or html...

I think it is pretty amazing how many special cases there are for something like this...

AlexPl292:
@Nimamoh 
@Vvalter 
Okay, this is going to be a tough decision. The main problem with https://github.com/JetBrains/ideavim/pull/189 is the use of a lexer that is not part of OpenApi and I am still not sure if it is legal to use it. I am now clarifying this point.

I am also actively trying to find out @vlasovskikh  opinion. He has been developing IdeaVim for a very long time and his opinion can have a great impact.

Unfortunately, I have to ask you to wait one more day before I can clarify all the necessary points. I now have the highest priority to deal with this difficult situation.

Nimamoh:
@Vvalter indeed that would be a "limitation" of using a parser, that might be different from vi behavior but I considered it acceptable, that's an arbitrary decision.

@AlexPl292 that's okay, thanks for paying attention to our work. @Vvalter sumed up the pro/cons pretty well in #179 . Would it be a tie I want to remind @Vvalter made the PR before ;)

vlasovskikh:
I would prefer the fix for tag blocks from #179 for a couple of reasons:

* With regex-based code it's easier to emulate all the intricacies of the original Vim behavior for `it` and `at` motions
* `XmlLexer` is not in a public API package (it's from `intellij.xml.psi.impl`)

AlexPl292:
Yeah, we ended up deciding on a solution from @Vvalter.

@Nimamoh thank you for your work. Bracket fixes are going to master, of course. Could you please roll back the changes that are related to the tags?

@Vvalter, do I understand correctly that there are still moments left that you would like to finish?
It would be great if you could use the tests that @Nimamoh wrote. They look pretty detailed.

Nimamoh:
Okay, should be good to go for regular block selections

AlexPl292:
Great! Thank you for your contribution!

PR #190, opened 2019-04-18
change gD to goto java Implementation (VimGotoImplementation)

wanc361:


AlexPl292:
Hi!
Than you for PR!
According to docs, `gD` command is also mapped to "Goto global Declaration."
If you want `gD` command for "GoToImplementation", you can put the next lines into your .ideavimrc:
`map gD :action GotoImplementation<CR>`

wanc361:
thanks for your reply.
yes, map gD :action GotoImplementation<CR>` will works 
but it seems do not save current location, thus <C-O> will not jump back where i left.
After test, sometimes it jump back, some times it doesn't

AlexPl292:
Well, I see your point of view, but we still can't match `gD` with another action. We are trying to emulate vim as close as possible, and this is a direct change in functionality. We need another approach to solve this problem. At this point, I can ask you to leave a bug [here](https://youtrack.jetbrains.com/issues/VIM) that it's impossible to go back after calling `:action`.

PR #191, opened 2019-04-26
Only beep once every half second

citizenmatt:
Simple change to only play the system beep once every half second, the same as Vim (see [:help 'visualbell'](https://vimhelp.org/options.txt.html#%27visualbell%27)).


vlasovskikh:
@citizenmatt Thanks for your PR! Out of curiosity, what motivated you to fix this one? How have you encountered many beeps?

citizenmatt:
Yes, but it's correct Vim behaviour. You can get beeps for hitting Esc in normal mode, or reaching the end of search or command history in ex mode, and probably loads of others. Looking up `visualbell` to add to `.ideavimrc`, I saw that Vim's behaviour played the beeps less frequently, which is much less annoying, and it was just too easy not to add it 😄

AlexPl292:
> Vim has two modes. One that beeps at you, and another that ruins everything. 

Thank you for making IdeaVim less beep-aggressive.

PR #195, opened 2019-05-19
Fix and improve search highlighting and behaviour

citizenmatt:
Fixes some bugs with search highlights and behaviour, bringing behaviour more in line with Vim.

- [x] [VIM-1535](https://youtrack.jetbrains.com/issue/VIM-1535) - Use same text attributes ("Text search results") and highlight layer (selection - 1) as IntelliJ's own Find feature. This makes search results more visible. Previously, it would add the highlight below e.g. "identifier under caret" and would use the same colours as Highlight Usages, meaning the current search highlight could easily be lost.
- [x] Hide IntelliJ's Find results when starting incsearch, and hide IdeaVim search results when starting IntelliJ Find
- [x] Add tests for search highlights and incremental search behaviour
- [x] Move the current line during incremental search. Original caret location and scroll position is restored if the search is cancelled
- [x] Search highlights are reset and redrawn when the `ignorecase`, `smartcase` and `hlsearch` options are modified. Does not redraw if the highlights have been hidden with `:nohlsearch`
- [x] Highlight all find results instead of just first with `incsearch` and `hlsearch` enabled
- [x] Remove old search highlights when starting to type a new search with `incsearch` and `hlsearch` enabled
- [x] Highlight first search result if `incsearch` is disabled, and `hlsearch` is enabled (as per Vim)
- [x] Incremental search correctly strips trailing options, e.g. `/Foo/+1`
- [x] Add incremental search while typing `:substitute` command (as per Vim). Does not handle `:global`, `:vglobal`, `:smagic` or `:snomagic`, as these aren't supported by IdeaVim
- [x] Fix disabling Vim emulation does not remove search highlights
- [x] [VIM-1413](https://youtrack.jetbrains.com/issue/VIM-1413) - Fix `smartcase` option being ignored in incremental search.
- [x] [VIM-128](https://youtrack.jetbrains.com/issue/VIM-128) - Fix `:substitute` not respecting `ignorecase` and `smartcase` options
- [x] Fix next/previous search not handling `smartcase` override for search next word. E.g. `*` on `Editor` overrides `smartcase`. Now `n`/`N` will remember if `smartcase` was overridden.
- [x] Only close ex entry on `<BS>` if there is no text
- [x] Treat `<Del>` at end of ex entry text as `<BS>`
- [x] [VIM-1682](https://youtrack.jetbrains.com/issue/VIM-1682) - Fix exception when searching backwards
- [x] Fix exception when trying to show search highlights for the last `\n` in the file


citizenmatt:
This PR has turned into a bit of a beast. I'm not going to add any more features. I've taken these items out, and will do them as a new PR, based on this, after it lands in master:

- Support `/_CTRL-G`/`c_CTRL-G` and `/_CTRL-T`/`c_CTRL-T` to move between next/previous match during incremental search
- Support `/_CTRL-L` to add one character to search from word at current match
- Support `c_CTRL-R_CTRL-W` to add word to search from current match

src/com/maddyhome/idea/vim/ui/ExEntryPanel.java
```diff
@@ -302,13 +307,50 @@ public void componentResized(ComponentEvent e) {
...
-      VimPlugin.getSearch().updateIncsearchHighlights(editor, pattern, forwards, caretOffset);
+      VimPlugin.getSearch().updateIncsearchHighlights(editor, pattern, forwards, caretOffset, searchRange);
+    }
+
+    @Nullable
+    private ExCommand getIncsearchCommand(String commandText) {
+      try {
+        final ExCommand exCommand = CommandParser.getInstance().parse(commandText);
+        final String command = exCommand.getCommand();
+        if (command.equals("s") || command.equals("substitute")
```
AlexPl292: Substitute command (and others) can be called using each amount of characters from short-size to full-size. So, `su`, `sub`, `subs`, etc. are also valid as well.

citizenmatt: Aha, I didn't know that. I'll get it fixed.

citizenmatt:
I think it's possible to add a new text attribute for "IdeaVim text search results", which has a fallback to the default "text search results" attribute. So it would have the same colour as the IDE by default, but you can override it in settings if you want to. How does that sound?

AlexPl292:
Thank you very much @citizenmatt ! IdeaVim search is much better now!

PR #198, opened 2019-06-11
Fix exception with dead keys in ex entry

citizenmatt:
The custom caret handling in the ex text field made an incorrect assumption that it would always be the only caret set in the text field, and this caused an invalid cast exception when dead keys were being used to compose text.

For example, on Mac, in the standard UK/US keyboards, typing `<A-N>` followed by `n` will generate the accented char `ñ`. After typing `<A-N>`, the text field adds the _composed text_ `˜` and changes the caret to an instance of `ComposedTextCaret`. The ex text field assumed the caret would always be the custom `CommandLineCaret`, and an exception was thrown, which breaks the ex entry for further use.

This PR also changes the style of composed text in the ex entry field, for Mac only. This means the first `˜` character shown when typing `<A-N>` is drawn with a highlight, similar to how other Mac apps behave. Most native Mac apps, and the standard text fields, use a different colour background as highlight, but some Mac apps, such as Safari and IntelliJ's editor use an underline, which is used here. The command line version of Vim for Mac depends on the terminal it's running in - Terminal.app uses a background colour highlight, while iTerm2 uses underline. The GUI MacVim app does not show any highlights in the ex entry field (although it does in the main text area). It appears that Java does not natively support showing highlights for dead key apps, so most other fields in IntelliJ don't show highlights. I've raise [IDEA-216033](https://youtrack.jetbrains.com/issue/IDEA-216033) to implement this in standard IntelliJ text fields. Note that Windows doesn't give any feedback for dead keys at all.

This PR does not address further dead key issues:

* Should composed text be ignored during search? E.g. typing `/ni<Alt+N>` is the same as typing `/ni`, and not `/ni˜`. If yes, I can include it as part of #195.
    Command line Vim for Mac does ignore it, but that might be a side effect of who is handling the dead keys - if it's the terminal, then Vim won't see the dead key. This is likely, as it would also explain why the composed text is shown in normal mode too.
* The IntelliJ editor inserts composed text directly into the text, without passing it through IdeaVim's key handling. See [VIM-1503](https://youtrack.jetbrains.com/issue/VIM-1503), [VIM-1531](https://youtrack.jetbrains.com/issue/VIM-1531) and [VIM-1628](https://youtrack.jetbrains.com/issue/VIM-1628)


src/com/maddyhome/idea/vim/ui/ExTextField.java
```diff
@@ -359,17 +361,14 @@ private void resetCaret() {
   // see :help 'guicursor'
   // Note that we can't easily support guicursor because we don't have arbitrary control over the IntelliJ editor caret
   private void setNormalModeCaret() {
-    CommandLineCaret caret = (CommandLineCaret) getCaret();
     caret.setBlockMode();
```
AlexPl292: is it any possibilities that caret would be invalidated and `caret.isValid()` would be `false`?

citizenmatt:
Right now, search will include the composed text char, so it would actually search for `n˜`, as two characters. The second isn't the standalone tilde character `~` (u+007E) but the diacritic tilde accent char `˜` (u+0303, I think).

Not sure what you mean by the "safe way" here - do you want the character to be removed from the search text?

AlexPl292:
> Not sure what you mean by the "safe way" here 

Oh, I'm sorry, I misunderstood you.

Well, yeah, I guess `/ni<Alt+N>` should be searched like `/ni˜`. So, in this case, is this PR ready for review?

citizenmatt:
Yes, I think this is complete

citizenmatt:
@AlexPl292 Are you happy to get this merged in? It would be good to fix these exceptions.

AlexPl292:
HI @citizenmatt. Yes, I think that can be merged. We've started to prepare to a stable release, so this PR will be merged afterwards (Release is planned to 23.07).

citizenmatt:
This fixes [VIM-1736](https://youtrack.jetbrains.com/issue/VIM-1736), and I think it'll fix [VIM-1739](https://youtrack.jetbrains.com/issue/VIM-1739) too, which has the same symptoms.

citizenmatt:
Looks like VIM-1736 + VIM-1739 was just fixed in ed2b8dd12913a6c52db9e54f26776030a31f0a63.

This PR goes a little further, including proper highlighting for the dead key character in the text field. It also makes the ex entry panel a bit more robust to similar issues. The UI is always hidden, even if there's a problem during deactivation. It will also free the existing cached panel instances when disabling the plugin, so we get clean panel instances by toggling VIM emulation.

Note that this still isn't a full reset, as the editor user data manager isn't reset, so we can still get stuck with stale state by toggling the editor.

PR #200, opened 2019-07-26
Fix ideajoin examples URL

NOLFXceptMe:
Recent commit d06f2a2d181e46c7737c2919fff9d7171d444f3b moved ideajoin examples page from doc/ to wiki.


AlexPl292:
Thank you for the fix!

PR #201, opened 2019-07-26
 Support f and F surround commands (function name)

jorgengranseth:
https://github.com/tpope/vim-surround/blob/master/doc/surround.txt#L138

I noticed that this was missing from IdeaVim, and it seems very useful and not all that complicated, so I gave it a go.

... and it turned out to be somewhat complicated after all since `<C-F>` keystrokes are not caught by the current extension (which is based on pure characters, so `<C-F>` is ignored). I made an attempt to implement that as well here:
https://github.com/JetBrains/ideavim/compare/master...jorgengranseth:attempt-at-ctrl-f

While the test passes, it doesn't work in the IDE since `<C-F>` gets hijacked by IdeaVim as the page-down motion. Setting the key mapping to undefined didn't work - I'm not sure if this is a bug - but I am also not sure if this is the right approach since I haven't looked into other possible side effects by allowing `<C-F>` at random times in the surround plugin.

Do you have any thoughts on how we can implement the `<C-F>` part of `vim-surround`? It is the equivalent of the `f` command for ML-like languages, so I would personally find it useful to have available. On the other hand I'd rather have `f` and `F` than nothing at all.

### DIFF
--- a/src/com/maddyhome/idea/vim/extension/surround/VimSurroundExtension.java
+++ b/src/com/maddyhome/idea/vim/extension/surround/VimSurroundExtension.java
@@ -128,8 +128,34 @@ public class VimSurroundExtension extends VimNonDisposableExtension {
   }
 
   @Nullable
+  private static Pair<String, String> inputFunctionName(
+    @NotNull Editor editor,
+    boolean withInternalSpaces
+  ) {
+    final String functionNameInput = inputString(editor, "function: ");
+
+    if (functionNameInput.isEmpty()) {
+      return null;
+    }
+
+    return withInternalSpaces
+      ? Pair.create(functionNameInput + "( ", " )")
+      : Pair.create(functionNameInput + "(", ")");
+  }
+
+  @Nullable
   private static Pair<String, String> getOrInputPair(char c, @NotNull Editor editor) {
-    return c == '<' || c == 't' ? inputTagPair(editor) : getSurroundPair(c);
+    switch (c) {
+      case '<':
+      case 't':
+        return inputTagPair(editor);
+      case 'f':
+        return inputFunctionName(editor, false);
+      case 'F':
+        return inputFunctionName(editor, true);
+      default:
+        return getSurroundPair(c);
+    }
   }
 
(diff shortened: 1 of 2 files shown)


jorgengranseth:
I added `s` as well:
https://github.com/tpope/vim-surround/blob/master/doc/surround.txt#L148

In the process I noticed that the caret actually moves if the function name is empty. We can achieve this by returning two empty strings instead of `null`. It would be more authentic `vim-surround` but I can't see a good reason to want that behaviour (it could be a bug in the original plugin). Should I add it anyway?

AlexPl292:
Hi @jorgengranseth!
Thank you for contribution! The PR looks nice, I don't see any obstacles to accepting it.
> `<C-F>` command. 

Currently the infrastructure for creating emulated plugins for IdeaVim is not fully complete. E.g. with the current system there are some problems with emulating easymotion plugin. So, we are going to have a chat with @vlasovskikh to make IdeaVim extensions more flexible. We will also discuss the problem with non-typed commands, so I think we can implement `<C-F>` surround motion after this upgrade.

> In the process I noticed that the caret actually moves if the function name is empty.

Could you please provide a small example?

jorgengranseth:
Cool, I'll look forward to hearing your thoughts.

>> In the process I noticed that the caret actually moves if the function name is empty.

> Could you please provide a small example?

Just to be clear, this happens in Vim, not in the PR as it is written now.

1. Screenshot 2019-07-29 at 17 10 10
1. type `ysiwf<CR>`

1. Screenshot 2019-07-29 at 17 10 31

AlexPl292:
Okay, I believe it's fine to keep the current behavior. It looks more convenient.

So, I put the "merge approved" label to this PR. It means that currently we are not going to merge a new functionality, but we are working on stabilizing the current release. I'll merge your PR right after the bugfix release.

Thank you for this PR and for bug reports in youtrack!

jorgengranseth:
Great! I'll emphasise this decision in the test.

AlexPl292:
Okay, this PR also goes to master 👍

PR #202, opened 2019-08-01
VIM-1720 Implement Ex command :delm[arks]

jorgengranseth:
Deletes existing marks. See
http://vimdoc.sourceforge.net/htmldoc/motion.html#:delmarks
for details.

[VIM-1720](https://youtrack.jetbrains.com/issue/VIM-1720)

Note that the `'` jump mark cannot be deleted with this command. This is why I decided to define the deletable marks separately from the constants already defined in `MarkGroup`, but I'd like some feedback on whether this is a good idea or not.

Similarly I've preprocessed the argument to account for all edge cases I could find. I imagine this is more readable than writing a parser, but if you'd prefer changes here I'm OK with changing it. A side effect of this choice is that an invalid command like `:delmarks %f-h` will throw the error message `E475: Invalid argument: %fgh` rather than `E475: Invalid argument: %f-h` (because it has already parsed the valid portion of the command). It has no functional side effect as far as I am aware: If the argument list is invalid no marks are deleted in Vim either.

Furthermore; for errors - what is preferable between the
```
          VimPlugin.showMessage(...)
          return false
```
pattern and throwing a suitable exception, and why?

On the topic of exceptions, I would like to know why messages from flag errors don't show. It seems the way these
https://github.com/JetBrains/ideavim/blob/master/src/com/maddyhome/idea/vim/ex/CommandHandler.kt#L129-L153
lines are written immediately replace the error messages with empty ones - is this intentional? It seems that moving the messages into the exceptions would give the desired behaviour.

### DIFF
--- a/resources/messages.properties
+++ b/resources/messages.properties
@@ -67,4 +67,5 @@ E385=E385: search hit BOTTOM without match for: {0}
 e_patnotf2=Pattern not found: {0}
 unkopt=Unknown option: {0}
 e_invarg=Invalid argument: {0}
+E475=E475: Invalid argument: {0}
 E774=E774: 'operatorfunc' is empty
--- a/src/com/maddyhome/idea/vim/ex/CommandParser.java
+++ b/src/com/maddyhome/idea/vim/ex/CommandParser.java
@@ -53,6 +53,7 @@ public class CommandParser {
     new CopyTextHandler(),
     new DelCmdHandler(),
     new DeleteLinesHandler(),
+    new DeleteMarksHandler(),
     new DigraphHandler(),
     new DumpLineHandler(),
     new EditFileHandler(),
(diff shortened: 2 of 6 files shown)


AlexPl292:
Hey,
I've updated your code a bit. Nothing special.
So, as for me, I would prefer to have a single "source of truth" where all the possible marks are presented. And it's not so hard to split `WR_FILE_MARKS` into two sets. The more complicated problem is... to find names for them :)

> lines are written immediately replace the error messages with empty ones - is this intentional? 

Well... The problem is that this code was written in 2003, and we just had no time to refactor this point (I mean, how to return from a function if something wrong). We've recently done a big work and reformatted all the code of the handlers. There exceptions and error messages are on the list for the further work.
Currently I suggest to use `return false`. For no particular reason, just because I used it before.

And could you please rebase your branch on the last master? `CommandFlags.Access` is now the required field for command handlers. That would be "READ_ONLY" for your case, I guess.

jorgengranseth:
@AlexPl292 Nice edit. I've made an attempt at naming variables in `MarkGroup`, and ended up with a compromise between keeping the number of variables down and the outcome be somewhat sensible. Thoughts?

AlexPl292:
Okay, I think this is a nice solution. Thank you for PR!

PR #203, opened 2019-08-05
Add tests for CharacterClasses

jancijen:
I've analysed your codebase and noticed that `com.maddyhome.idea.vim.regexp.CharacterClasses` is not fully tested.
I've written some tests for the methods in this class with the help of [Diffblue Cover](https://www.diffblue.com/opensource).
Hopefully, these tests will help you detect any regressions caused by future code changes. If you would find it useful to have additional tests written for this repository, I would be more than happy to look at other classes that you consider important in a subsequent PR.


AlexPl292:
Hi! Thank you for paying attention to our project. Unfortunately, we prefer the functional style of testing, and we don't really use unit tests.

PR #204, opened 2019-08-10
Improve relative line numbers and VCS annotations

citizenmatt:
This PR makes relative line numbers play (mostly) nice with VCS annotations. The current implementation is a `TextAnnotationGutterProvider`, which is a core platform API, but only really implemented by the VCS subsystem. 

The main problem today (notably [VIM-1032](https://youtrack.jetbrains.com/issue/VIM-1032)) is that the VCS subsystem doesn't expect other annotation providers, and incorrectly shows annotations as enabled when they're not. Toggling (twice) will show the VCS annotations, but moving the caret closes them and shows the relative line numbers again.

There isn't a solution that perfectly fixes everything, but this PR gives the best implementation, by firstly improving the presentation and behaviour of the gutter provider and secondly replacing it with a line number converter function, falling back in the unlikely event that converters aren't available.

For gutter providers:
* They no longer close VCS annotations when the caret moves
* In 183 and 191, the "VCS | Git Annotate…" toggle action is still incorrectly enabled. Toggling hides the relative numbers provider. Toggling again shows the VCS annotations. Moving the caret will now re-show the relative line numbers and leave the VCS annotations visible.
* In 192, the "VCS | Git Annotate…" toggle works correctly, and is only shown when the VCS annotations are visible. Toggling this does not affect the relative numbers provider.
* `:set nornu` closes all annotation provides, including VCS. 192 has an API that fixes this, but we're still targeting 183
* The presentation of the gutter provider is improved, but doesn't exactly match Vim - soft wrapped lines have a repeated line number on each wrapped line, while Vim only shows the number on the first line

For line number converters:
* Uses the built in line number presentation, just provides a different number to display. This is the same mechanism used by the diff views
* Does not affect VCS annotations at all
* Soft wrapped line numbers are shown the same as in Vim
* Technically, it is possible that another plugin could overwrite the line number converter. API watcher shows that no other plugins are using this API
* In the (very) unlikely event that the `EditorGutter` instance can't be downcast to `EditorGutterCopmonentEx`, the code will fall back to the gutter annotation providers.
* Minor issue: If `'number'` is not set, the current line number is shown as blank instead of "0". If `:set number` is added, the current line number is correctly displayed. This is an API decision, and unlikely to change in the platform (see [IDEA-211165](https://youtrack.jetbrains.com/issue/IDEA-211165))

So, the upshot is that gutter annotation providers work great if we're targeting 192 (and can use the proper 'close' API), although soft wrapped lines are not presented in the same way as Vim. They are ok, but not great in 183 and 191.

Line number converters work well in all versions, although `:set nonumber` can leave a blank current line.

IMHO, line number converters are the best solution.

This PR also:

* Sets the `since-build` in `plugin.xml` to the build of 2018.3 RTM, which is needed to allow use of `EditorColors.LINE_NUMBER_ON_CARET_ROW`
* Shows the current line number in the correct colour - [VIM-1762](https://youtrack.jetbrains.com/issue/VIM-1762)
* Avoids an exception when closing a project with a preview panel open in Find Usages that is showing relative line numbers - [VIM-1671](https://youtrack.jetbrains.com/issue/VIM-1671) (and see also [IDEA-184351](https://youtrack.jetbrains.com/issue/IDEA-184351))
* Does not replace line number behaviour in diff windows, which normally provide their own line numbers. This affects both relative and non-relative line numbers. This also only affects line numbers in editable diff panes, as historical diffs were already excluded because they're not backed by a virtual file. The diff views now always provide their own line numbers. The main reason behind this change was that diff views also use line number converters, and avoiding them avoids any chance of clashing.

### DIFF
--- a/src/com/maddyhome/idea/vim/group/EditorGroup.java
+++ b/src/com/maddyhome/idea/vim/group/EditorGroup.java
@@ -29,6 +29,7 @@ import com.intellij.openapi.editor.event.CaretEvent;
 import com.intellij.openapi.editor.event.CaretListener;
 import com.intellij.openapi.editor.event.EditorFactoryEvent;
 import com.intellij.openapi.editor.ex.EditorEx;
+import com.intellij.openapi.editor.ex.EditorGutterComponentEx;
 import com.intellij.openapi.project.Project;
 import com.maddyhome.idea.vim.KeyHandler;
 import com.maddyhome.idea.vim.VimPlugin;
@@ -38,6 +39,7 @@ import com.maddyhome.idea.vim.option.OptionChangeListener;
 import com.maddyhome.idea.vim.option.OptionsManager;
 import kotlin.text.StringsKt;
 import org.jdom.Element;
+import org.jetbrains.annotations.Contract;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
@@ -59,8 +61,9 @@ public class EditorGroup {
 
   private final CaretListener myLineNumbersCaretListener = new CaretListener() {
     @Override
-    public void caretPositionChanged(CaretEvent e) {
-      updateLineNumbers(e.getEditor());
+    public void caretPositionChanged(@NotNull CaretEvent e) {
+      final boolean requiresRepaint = e.getNewPosition().line != e.getOldPosition().line;
+      updateLineNumbers(e.getEditor(), requiresRepaint);
     }
   };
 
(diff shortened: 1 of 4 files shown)


src/com/maddyhome/idea/vim/group/EditorGroup.java
```diff
@@ -29,15 +29,18 @@
...
 import com.intellij.openapi.editor.ex.EditorEx;
+import com.intellij.openapi.editor.ex.EditorGutterComponentEx;
 import com.intellij.openapi.project.Project;
 import com.maddyhome.idea.vim.KeyHandler;
 import com.maddyhome.idea.vim.VimPlugin;
 import com.maddyhome.idea.vim.helper.*;
 import com.maddyhome.idea.vim.option.OptionChangeEvent;
 import com.maddyhome.idea.vim.option.OptionChangeListener;
 import com.maddyhome.idea.vim.option.OptionsManager;
+import gnu.trove.TIntFunction;
```
AlexPl292: Wow...

PR #206, opened 2019-08-22
Fix jump between files with moveCaretToJump

sumoooru2:
I found a bug about Ctrl-O and Ctrl-I in normal mode and fixed it in this PR.

Given you are in a file A and jump to a file B (e.g. by `gd` command).
Push Ctrl-O, which works and you go back to file A.
Push Ctrl-I, which doesn't work correctly. You go to file B, but the position of the cursor is different.
Push Ctrl-O again, which doesn't work correctly and the position of the cursor is different.

I searched for this bug in youtrack and wasn't able to find any issue.


AlexPl292:
Thank you for the fix!

PR #207, opened 2019-08-30
Fix block insert with parameter hint

sumoooru2:
I found a bug about block insert with parameter hint.

Given parameter hint is active like below.
function(hint: "str1")
function(hint: "str1")
Then you try to change two "str1" into "str_1" with block insert (e.g. ctrl-v j I _ <esc>), they become like below.
function(hint: "str_1")
function(hint: "st_r1")

I fixed the bug by using logicalPosition instead of VisualPosition.

I searched for this bug in youtrack and wasn't able to find any issue.

### DIFF
--- a/src/com/maddyhome/idea/vim/group/ChangeGroup.java
+++ b/src/com/maddyhome/idea/vim/group/ChangeGroup.java
@@ -641,7 +641,7 @@ public class ChangeGroup {
             repeatInsertText(editor, context, started ? (i == 0 ? count : count + 1) : count);
           }
           else if (EditorHelper.getVisualLineLength(editor, visualLine + i) >= repeatColumn) {
-            caret.moveToVisualPosition(new VisualPosition(visualLine + i, repeatColumn));
+            caret.moveToLogicalPosition(new LogicalPosition(logicalLine + i, repeatColumn));
             repeatInsertText(editor, context, started ? (i == 0 ? count : count + 1) : count);
           }
         }


sumoooru2:
Thank you for checking my change and finding that!
In that case, expected behavior is that 1 is also inserted between 'e' and 'm' in "System", isn't it?
I will try to fix the problem when I have time.

AlexPl292:
> 1 is also inserted between 'e' and 'm' in "System", isn't it?

This is a nice question because vim doesn't have such kind of folding :)
From my understanding, no. If you use `<C-V>j`, 1 should be inserted only as it's inserted on the gif (and selection of `m` is a wrong behavious as well).

sumoooru2:
I see.
Thanks!

sumoooru2:
I worked on this and noticed that I was not sure what is the expected behavior concerning parameter hint.
e.g.
```
function2(str1, str2);
function(hint: "str1");
```
If a cursor is at 's' of str2 and push j, the cursor
- seems to go to 's' of "str1" visually
- goes to ')' in practice (with or without my change) maybe because the hint is considered one space
- goes to ';' without parameter hint

Is there any rule about this behavior?

AlexPl292:
Hey. Yeah, the situation is like that:
The rule about the vertical caret motion is quite simple: caret should save its column offset during up/down motion. And it does mean that 1) It doesn't save the column offset __visually__. If you are landing on a line with a hint, caret will be a bit to the right. 2) Currently it doesn't work properly with the hints.

About point 2:

If you'll check the IJ code about caret moving ([link](https://github.com/JetBrains/intellij-community/blob/master/platform/platform-impl/src/com/intellij/openapi/editor/impl/CaretImpl.java#L187)), you'll notice that there seems to be some additional logic about inline elements handling. Unfortunately, this logic is not yet implemented for IdeaVim. So, if you'll put a caret at 's' of str2 and type `jk`, you should be back on `s`. But you're probably going to end up on the letter `t` what is a bug.

sumoooru2:
I checked that by typing `jk` the cursor went back to `s` as expected.
The issue you found seems to exist regardless of my change, so I think this is another bug.

Should I include the fix for the bug in this MR as well, or is It okay to leave it for now?

AlexPl292:
I think it's fine to leave this problem as it is.

AlexPl292:
Sorry, for broken check, just ignore it.

AlexPl292:
Hi @sumoooru2 ! I'll close this PR because there is no activity for a long time.
Don't hesitate to reopen it in case of any new changes or create a new one.

BTW. The `jk` issue should be fixed with the last release.

PR #209, opened 2019-10-30
Fix multiple cursors selection getting stuck with ignorecase

igrekster:
With `ignorecase` enabled, select next (`<A-n>`) would get stuck on the
second occurrence even though there are more occurrences to select. This
was happening because part of logic was still doing case sensitive
comparison against the current pattern. The fix is to use case
insensitive comparator everywhere the pattern is used, if `ignorecase`
option is set.


AlexPl292:
Looks good to me. Thank you for your contribution!

Issue: https://youtrack.jetbrains.com/issue/VIM-1823

PR #212, opened 2019-11-15
VIM-664 - read ideavimrc from XDG config home directory

hajdamak:
This pull request solves [VIM-664](https://youtrack.jetbrains.com/issue/VIM-664). If `ideavimrc` file exist in XDG config directory use it instead the one in home dir. Works according to [XDG spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), that is it uses directory specified in `XDG_CONFIG_HOME` environment variable if set and `~/.config` otherwise.
Any feedback appreciated.

### DIFF
--- a/src/com/maddyhome/idea/vim/ex/vimscript/VimScriptParser.java
+++ b/src/com/maddyhome/idea/vim/ex/vimscript/VimScriptParser.java
@@ -30,6 +30,7 @@ import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
 import java.util.Map;
+import java.util.Objects;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -37,7 +38,9 @@ import java.util.regex.Pattern;
  * @author vlan
  */
 public class VimScriptParser {
-  public static final String[] VIMRC_FILES = {".ideavimrc", "_ideavimrc"};
+  public static final String VIMRC_FILE_NAME = "ideavimrc";
+  public static final String[] HOME_VIMRC_PATHS = {"." + VIMRC_FILE_NAME, "_" + VIMRC_FILE_NAME};
+  public static final String XDG_VIMRC_PATH = "ideavim/" + VIMRC_FILE_NAME;
   public static final int BUFSIZE = 4096;
   private static final Pattern EOL_SPLIT_PATTERN = Pattern.compile(" *(\r\n|\n)+ *");
   private static final Pattern DOUBLE_QUOTED_STRING = Pattern.compile("\"([^\"]*)\"");
@@ -51,14 +54,29 @@ public class VimScriptParser {
   @Nullable
   public static File findIdeaVimRc() {
     final String homeDirName = System.getProperty("user.home");
+
+    // Check whether file exists in home dir
     if (homeDirName != null) {
-      for (String fileName : VIMRC_FILES) {
+      for (String fileName : HOME_VIMRC_PATHS) {
         final File file = new File(homeDirName, fileName);
         if (file.exists()) {
           return file;
         }
       }
     }
+
+    // Check in XDG config directory
+    final String xdgConfigHomeProperty = System.getenv("XDG_CONFIG_HOME");
+    final File xdgConfig;
+    if (xdgConfigHomeProperty == null || Objects.equals(xdgConfigHomeProperty, "")) {
+      xdgConfig = new File(homeDirName, ".config/" + XDG_VIMRC_PATH);
+    } else {
+      xdgConfig = new File(xdgConfigHomeProperty, XDG_VIMRC_PATH);
+    }
+    if (xdgConfig.exists()) {
+      return xdgConfig;
+    }
+
     return null;
   }
 
@@ -69,7 +87,7 @@ public class VimScriptParser {
 
     final String homeDirName = System.getProperty("user.home");
     if (homeDirName != null) {
-      for (String fileName : VIMRC_FILES) {
+      for (String fileName : HOME_VIMRC_PATHS) {
         try {
           final File file = new File(homeDirName, fileName);
           //noinspection ResultOfMethodCallIgnored


AlexPl292:
Hi! Thank you for the contribution.
The PR looks good. Don't you want to support `XDG_CONFIG_DIRS` as well?
Also, does the original vim use the same order of lookup? I mean, if `.vimrc` is located in both home dir and XDG dir, which one will be chosen?

hajdamak:
VIM does not yet follow XDG (vim/vim#2034) . neovim does (neovim/neovim#728) but it does not support config in home. Would you prefer home take prio over XDG?

Yeah, I believe `XDG_CONFIG_DIRS` should also be supported. According to the spec when it is empty `/etc/xdg` should be used. And currently IdeaVim only reads ideavimrc from home dir. Are you ok to go full with this?

Most applications use dedicated directory for the app. Like for example `~/.config/ideavim`. And only place config file/s there. This PR just reads ideavimrc file form XDG dir. Are you ok to go with directory approach? And even maybe just name configuration file `config` like many apps do? Or is to much of a change?

Currently when file does not exist IdeaVim creates new one in `~` . Should it create it in XDG_CONFIG_HOME ?

AlexPl292:
> Would you prefer home take prio over XDG?

Well, technically, there should be no situations where there are two `ideavimrc` files in both directories. 
But yes, I'd suggest to check the home directory firstly because it's a little bit safer (and `ideavimrc` is a very important file). Let's imagine that someone has copied the XDG configurations of his friend and this friend has an `ideavimrc` file. It would be pretty difficult to understand why the home `ideavimrc` stopped working after the IdeaVim update.

> According to the spec when it is empty /etc/xdg should be used. Are you ok to go full with this?

Yes, but what about windows? Is there some kind of standard for it? If no, we'd better skip this :)

> And even maybe just name configuration file config like many apps do?

Oh, I understand, but I think IdeaVim is not yet ready for that. At least not in this PR.

> Currently when file does not exist IdeaVim creates new one in ~ . Should it create it in XDG_CONFIG_HOME ?

You mean `com.maddyhome.idea.vim.ex.vimscript.VimScriptParser#findOrCreateIdeaVimRc`, right?
Another interesting question. I may suggest:
1) Leave it as it is
2) Show a small dialog with "create in the home dir" / "create in XDG"
3) Show a clear notification of where the new `ideavimrc` is located.
What would be a better way in your opinion?

hajdamak:
> Yes, but what about windows? Is there some kind of standard for it? If no, we'd better skip this :)

The whole story on Windows, Linux and Mac is documented for example [here](https://github.com/ActiveState/appdirs).  As you can see it is rather complicated state of affairs. :-)

> You mean com.maddyhome.idea.vim.ex.vimscript.VimScriptParser#findOrCreateIdeaVimRc, right?

Yeap.

Now that I think of this the really proper way to go would be to fully support platform specific directories as described in link above and support home dir only as compatibility option for older installations. This is however a lot bigger and intrusive change.

So for this PR I would just keep IdeaVim working as it is currently in master with the only exception that if it does not find ideavimrc in home it will check `$XDG_CONFIG_HOME/ideavim/ideavimrc` or` ~/.config/ideavim/ideavimrc` .

What do you think?

AlexPl292:
Okay, yeah, that would be probably too complicated to support all the requirements. So, we can keep only the necessary things. I think that `$XDG_CONFIG_HOME/ideavimrc` can be supported as well as a simplified version of `$XDG_CONFIG_HOME/ideavim/ideavimrc`.

Regarding the new file creation, I think that if the user has `XDG_CONFIG_HOME` configured, that means they are familiar with it. So, let's create a new `ideavimrc` file in a`XDG_CONFIG_HOME`. But show the notification about the file location, if it doesn't exist yet.

So, to collect everything:
- Search path:
  1. `~/.ideavimrc`
  1. `~/_ideavimrc`
  1. `$XDG_CONFIG_HOME/ideavimrc`
  1. `$XDG_CONFIG_HOME/ideavim/ideavimrc`
- `XDG_CONFIG_HOME` defaults to `~/.config`
- If `.ideavimrc` doesn't exists, it's created at `$XDG_CONFIG_HOME/ideavimrc` or at `~/.ideavimrc` if  `XDG_CONFIG_HOME` is not set.

hajdamak:
> I think that `$XDG_CONFIG_HOME/ideavimrc` can be supported as well as a simplified version of `$XDG_CONFIG_HOME/ideavim/ideavimrc`

I believe that for the sake of KISS we should have one or the other, not both. `$XDG_CONFIG_HOME/ideavim/ideavimrc` is more popular and is more future proof (in case other configuration files will be needed) so maybe lets stick only with this. Ok?

> Regarding the new file creation, I think that if the user has`XDG_CONFIG_HOME` configured, that means they are familiar with it. So, let's create a new ideavimrc file in a `XDG_CONFIG_HOME`. But show the notification about the file location, if it doesn't exist yet.

The thing with XDG is that most people that use it don't have `XDG_CONFIG_HOME` set and just use default `~/.config`. So creating `ideavimrc` in XDG directory only if `XDG_CONFIG_HOME` is set and in `~` otherwise is like moving to XDG halfway for file creation.  Therefore I believe there are two sensible options here:
1. On Linux move fully to XDG (including file creation) and support `~` only as backward compatibility if there is a file from previous installation.
2. Keep IdeaVim working as it currently does but in case `ideavimrc` exists in `XDG_CONFIG_HOME/ideavim` or `~/.config/ideavim` and not in `~` just use it. This solves VIM-664 without breaking anything on other platforms and is compatible if option 1 will be implemented in the future.

Option 1 should be IdeaVim's target but in this PR I would stick to 2. What do you think?

AlexPl292:
> I believe that for the sake of KISS we should have one or the other, not both. $XDG_CONFIG_HOME/ideavim/ideavimrc approach is more popular and is more future proof (in case other configuration files will be needed) so maybe lets stick only with this. Ok?

I agree.

> Option 1 should be IdeaVim's target but in this PR I would stick to 2. What do you think?

Yes, let's choose the second option.

The thing about the file creation should be probably rethought as a separate issue. Maybe another option:
- If `XDG_CONFIG_HOME` exists, create the file in the $XDG_CONFIG_HOME/ideavim
- If not, create under the ~, but show the notification with an action `move to ~/.config`

But anyway I'll create a separate issue for that.

hajdamak:
Option 2 implemented. Please take a look.

hajdamak:
Thank you.

AlexPl292:
Yes, I think it's good now. Thank you for the contribution!

AlexPl292:
@austintraver Hi, thank you for the report. I've filed an issue in the bug tracker: https://youtrack.jetbrains.com/issue/VIM-2001

mikeduminy:
I just ran into the same issue as @austintraver. 

The problem is that the IDE doesn't have access to your `XDG_CONFIG_HOME` environment variable when it starts, so it defaults to looking in `~/.ideavimrc` and then `~/.config/ideavim/ideavimrc`. If your XDG config is located somewhere else then you have two options:
1. Start your IDE from the terminal (this passes starts the application within a context that it can read your terminal environment variables)
2. Share some terminal environment variables with launchd (sort of like Apple's process launcher) - https://stackoverflow.com/a/26586170.

Here's what I ran to fix this: ```launchctl setenv XDG_CONFIG_HOME ~/.xdg/config```

citizenmatt:
Following up from a recent issue ([VIM-3844](https://youtrack.jetbrains.com/issue/VIM-3844)), IdeaVim does work with `XDG_CONFIG_HOME`, but it requires two things. 

Firstly, as mentioned above, it must be defined as above, so that the GUI launcher (such as macOS' Finder, Spotlight or JetBrains Toolbox) can see the environment variable. Otherwise, it's not defined for the launched IDE and IdeaVim does not know about it.

Secondly, the environment variable should not contain `~`. This is a shell shortcut and not supported by the JVM, which IdeaVIm is built with. Instead, it should be defined in terms of `$HOME`, e.g. `$HOME/.xdg./config`. If the above `launchctl` command is entered in a terminal, the `~` is likely expanded before calling the app, which would then mean that the environment variable seen by the IDE is already a fully qualified path. Other ways of setting the variable might not be as lucky.

The PR #1139 will expand the `~`, although this is a workaround for what is likely an incorrectly configured environment variable.

PR #213, opened 2019-12-18
VIM-664 Fix usage of invalid path separator

hajdamak:
XDG config read implementation was changed to use invalid path separator.
This PR fixes this.


AlexPl292:
Oh, such a stupid mistake. Thanks for the fix.

hajdamak:
Thanks for merge.

PR #215, opened 2019-12-28
Refactor key handler state

citizenmatt:
This PR extracts all of the key handler state from `KeyHandler` and moves it to per-editor. Some state was already per-editor, and some was global, which lead to weird bugs like typing `3d` in one editor, switching to another and typing `2w` and deleting a load of text.

This PR also fixes:
* [VIM-1284](https://youtrack.jetbrains.com/issue/VIM-1284) by allowing numbers to be mapped, while still handling `0`, which can't be mapped, as per Vim (see [:help :map-modes](https://vimhelp.org/map.txt.html#%3Amap-modes))
* A bug with counts, e.g. `3d2w` would delete 32 words instead of 6.
* The way digraphs and literals key strokes are handled, meaning `<C-K>` and `<C-V>`/`<C-Q>` can be remapped


src/com/maddyhome/idea/vim/KeyHandler.java
```diff
@@ -200,67 +208,54 @@ public void handleKey(@NotNull Editor editor,
...
-      // Update the count
-      count = count * 10 + (chKey - '0');
-    }
-    else if (!waitCommandFinish(editor) && allowKeyMappings && handleKeyMapping(editor, key, context)) {
-      if (editorState.getMappingMode() != MappingMode.OP_PENDING ||
-          currentCmd.isEmpty() ||
-          currentCmd.peek().getArgument() == null ||
-          Objects.requireNonNull(currentCmd.peek().getArgument()).getType() != Argument.Type.OFFSETS) {
+    if (allowKeyMappings && handleKeyMapping(editor, key, context)) {
+      if (!editorState.isOperatorPending() || commandBuilder.peekCurrentCommandPartArgumentType() != Argument.Type.OFFSETS) {
```
citizenmatt: I'd like to get rid off `peekCurrentCommandPartArgumentType` if possible, and replace it with something a little more semantic, but I don't understand what's going on here - why are we returning at this point?

src/com/maddyhome/idea/vim/key/KeyMapping.java
```diff
@@ -51,7 +51,10 @@
   }
 
   @Nullable
-  public MappingInfo get(@NotNull List<KeyStroke> keys) {
+  public MappingInfo get(@NotNull Iterable<KeyStroke> keys) {
+    // Having a parameter of Iterable allows for a nicer API, because we know when a given list is immutable.
```
AlexPl292: Note for myself: TODO

src/com/maddyhome/idea/vim/command/CommandState.java
```diff
@@ -73,6 +86,79 @@ public static CommandState getInstance(@Nullable Editor editor) {
...
+  public void setCommandArgument(Argument argument) {
+    commands.peek().setArgument(argument);
+  }
+
+  public boolean hasCommandArgument() {
+    final Command command = commands.peek();
+    return command != null && command.getArgument() != null;
+  }
+
+  // TODO: What's the difference between this and getCurrentActionType?
```
AlexPl292: Note for myself: TODO

src/com/maddyhome/idea/vim/KeyHandler.java
```diff
@@ -352,188 +336,261 @@ private void handleEditorReset(@NotNull Editor editor, @NotNull KeyStroke key, @
...
-            @Nullable Integer startOffset = startOffsets.get(caret);
-            if (caret.hasSelection()) {
-              final VimSelection vimSelection = VimSelection.Companion
-                .create(UserDataManager.getVimSelectionStart(caret), caret.getOffset(),
-                        SelectionType.fromSubMode(CommandStateHelper.getSubMode(editor)), editor);
-              offsets.put(caret, vimSelection);
-              commandState.popState();
+      // Cache isOperatorPending in case the extension changes the mode while moving the caret
+      // See CommonExtensionTest
+      // TODO: Is this legal? Should we assert in this case?
```
AlexPl292: Note for myself: TODO

src/com/maddyhome/idea/vim/command/CommandState.java
```diff
@@ -296,19 +435,17 @@ public void setDotRepeatInProgress(boolean dotRepeatInProgress) {
...
-  private static class State {
+  private static class ModeState {
     @NotNull private final Mode myMode;
-    @NotNull private SubMode mySubMode;
-    @NotNull private final MappingMode myMappingMode;
+    @NotNull private final SubMode mySubMode;
 
     @Contract(pure = true)
-    public State(@NotNull Mode mode, @NotNull SubMode subMode, @NotNull MappingMode mappingMode) {
+    public ModeState(@NotNull Mode mode, @NotNull SubMode subMode) {
```
AlexPl292: This is great. I've been wanting to remove this `MappingMode` for a long time.

src/com/maddyhome/idea/vim/KeyHandler.java
AlexPl292: This is a part of the logic that allows plugins to provide custom text objects. It's a bit messy right now, but since it should be updated to implement `:normal` command, I'll take care of this method.

src/com/maddyhome/idea/vim/KeyHandler.java
AlexPl292: @citizenmatt sorry, what do you mean to assert?

citizenmatt: My lack of understanding of what's happening here. Later in this method, we're acting based on what the state _was_, not what it currently is, because the extension has changed the mode. Is it correct that the extension can change mode, and is it correct that we ignore the current mode?

`CommonExtensionTest` is changing the mode, but is that because it's a valid thing to do, or is it because it's a convenient way to test this bit of code?

And if it's not valid to change the mode, then I think we should add an assert that the extension did not change it, so that any extension that is incorrectly written will throw an exception during development.

AlexPl292: Aha, I see. Yes, this is a correct behavior that the extension can change the mode and IdeaVim should handle this case. E.g. suppose that an extension that provides a text object "between commas". So, for this text: `hello, this <caret>is my, text` the "between commas" object would be `this is my`. Since this is a text object, the user should have an ability to use them along with `d`, `y`, `c`, etc. commands. But here is a problem: vim doesn't really provide an API to create a text object. Thus this extension should just add a visual selection for everything between commas. Vim understands it fine and if a selection appears during the operator pending mode, it just applies the command (for example, `d`) to this selected area. So, from the user point of view, it looks like a text object, even there is no command in vim such as `create a new text object`.
The same thing works with `:normal` command (not supported yet). You can map some keys to `:normal T,vt,` (also, select everything between commas), and it will act as a text object:
```
onoremap i, :normal T,vt,
di,    <-- works like `delete between commas`
```

So yes, vim saves the state before the plugin execution (if it was operator pending mode), and continues the command after the plugin execution regardless of the state after the plugin.

Sorry for the confusion. I'll document it a little bit clearer during the reformation for `:normal` command support.

PR #217, opened 2020-02-02
Add support for buffer list (buffers, files, ls)

angelbot:
Add support for triggering recent files dialog when ls, files, or buffers ex command is entered. Not identical to vim but close equivalent.


AlexPl292:
Hi @angelbot !
Much appreciation for the contribution. Unfortunately, since we try to emulate vim as close as possible, we'd like to have an output in a vim-like panel (`com.maddyhome.idea.vim.ui.ExOutputPanel`). What do you think about updating your PR and print the list of buffers there with the information specified in the docs?

AlexPl292:
Yep, thank you for the contribution! I've added some small changes regarding Kotlin, nothing important.

PR #218, opened 2020-02-04
Macros record input keystrokes, not mapped keystrokes

citizenmatt:
This PR fixes [VIM-1835](https://youtrack.jetbrains.com/issue/VIM-1835) by recording input keystrokes for macros instead of the mapped keystrokes.

This PR is also required as a stepping stone for `showcmd` support ([VIM-434](https://youtrack.jetbrains.com/issue/VIM-434)), as both issues require refactoring `KeyHandler#handleKey`, and I thought it best to separate the issues.

A few other things were fixed as part of this PR:
* [VIM-1899](https://youtrack.jetbrains.com/issue/VIM-1899) - support an argument to the `:registers` command
* [VIM-1900](https://youtrack.jetbrains.com/issue/VIM-1899) - improve output of `:registers`, `:marks`, `:jumps` and `:ascii`


src/com/maddyhome/idea/vim/KeyHandler.java
```diff
@@ -213,71 +213,60 @@ public void handleKey(@NotNull Editor editor,
     // If this is a "regular" character keystroke, get the character
     char chKey = key.getKeyChar() == KeyEvent.CHAR_UNDEFINED ? 0 : key.getKeyChar();
 
-    final boolean isRecording = editorState.isRecording();
-    boolean shouldRecord = true;
-
-    if (allowKeyMappings && handleKeyMapping(editor, key, context)) {
-      if (!editorState.isOperatorPending() || commandBuilder.peekCurrentCommandPartActualArgumentType() != Argument.Type.OFFSETS) {
-        return;
-      }
```
citizenmatt: I don't know what this is checking, and removing it didn't break any tests, which always makes me nervous. Perhaps there's a test scenario missing?

test/org/jetbrains/plugins/ideavim/action/CopyActionTest.java
```diff
@@ -146,6 +146,8 @@ public void testStateAfterYankVisualBlock() {
   }
 
   // VIM-476 |yy| |'clipboard'|
+  // TODO: Review this test
+  // This doesn't use the system clipboard, but the TestClipboardModel
```
citizenmatt: Testing the output of `:registers` requires mocking the system clipboard to get stable results. That changes the purpose of these tests. Previously, they tested:

a) the `*` register is the system clipboard register
b) integration with the system clipboard

Now, they simply test a) the `*` register is the system clipboard register, but the data from the clipboard comes from a mock. 

If this is ok, I'll remove the TODO comments. If we also want to test b) then we'll need new tests.

src/com/maddyhome/idea/vim/ex/handler/RegistersHandler.kt
```diff
@@ -21,15 +21,25 @@ package com.maddyhome.idea.vim.ex.handler
...
   override val argFlags = flags(RangeFlag.RANGE_OPTIONAL, ArgumentFlag.ARGUMENT_OPTIONAL, Access.READ_ONLY)
   override fun execute(editor: Editor, context: DataContext, cmd: ExCommand): Boolean {
-    val regs = VimPlugin.getRegister().registers.joinToString("\n", prefix = "--- Registers ---\n") { reg ->
-      """"${reg.name}   ${StringHelper.toKeyNotation(reg.keys)}"""
-    }
+
+    val registerGroup = VimPlugin.getRegister()
+    val regs = registerGroup.registers
+      .filter { cmd.argument.isEmpty() || cmd.argument.contains(it.name) }
+      .joinToString("\n", prefix = "Type Name Content\n") { reg ->
```
AlexPl292: Why have you changed the output template? I mean, I think your version is better, but it differs from vim, isn't it?

src/com/maddyhome/idea/vim/KeyHandler.java
AlexPl292: Yes, I guess the tests for this scenario are in the IdeaVim-EasyMotion project. I'll check this code and add tests.

src/com/maddyhome/idea/vim/ex/handler/RegistersHandler.kt
citizenmatt: Ah, I changed it to match Vim on my machine, and it looks like Vim's output was [updated in 8.1.2212](https://github.com/vim/vim/commit/3691f1ee72b68a47e6dcc75927bfa46565cf3614#diff-19aa2e6e5c0b4e89aeee2e37be194690)

AlexPl292: Oh, I have an 8.1 version, but seems like an older update. Sorry for troubles, I didn't mean to send you to vim sources :)

PR #219, opened 2020-02-07
Fix block actions (i.e ci{) in presence of quotes (VIM-1008)

fan-tom:
Finally make it to pass all tests


AlexPl292:
Wow, this is a great refactoring! Thank you for your contribution.

PR #220, opened 2020-02-11
Implement showcmd

citizenmatt:
This PR implements [VIM-434](https://youtrack.jetbrains.com/issue/VIM-434) adding support for `'showcmd'`, showing the current (partial) command in the status bar.

[Vim enables this setting by default](https://vimhelp.org/options.txt.html#%27showcmd%27) everywhere apart from Unix (in case of slow terminals), and also enables it in [`defaults.vim`](https://vimhelp.org/starting.txt.html#defaults.vim). Similarly, it is on by default in IdeaVim.


AlexPl292:
Yeah, great! Thank you for your contribution!

PR #221, opened 2020-02-15
argtextobj: support bracket pairs configuration via let g:argtextobj_pairs="..."

igrekster:
argtextobj by default only handles arguments inside parenthesis. This is
very limiting when editing C++ source files. This change allows the list
of bracket pairs to be configurable via a global IdeaVim variable. The
variable takes effect immediately.


AlexPl292:
Yes, looks great!

PR #224, opened 2020-02-18
Populate intelij navigation history together with ideavim jumplist

pmnoxx:
Currently we keep intelij navigation history separate from ideavim jumplist.
This change will cause all updates to ideavim jumplist to be propagated to intelij navigaration history as well.

https://youtrack.jetbrains.com/issue/VIM-44

### DIFF
--- a/src/com/maddyhome/idea/vim/group/MarkGroup.java
+++ b/src/com/maddyhome/idea/vim/group/MarkGroup.java
@@ -30,6 +30,7 @@ import com.intellij.openapi.editor.event.DocumentEvent;
 import com.intellij.openapi.editor.event.DocumentListener;
 import com.intellij.openapi.editor.event.EditorFactoryEvent;
 import com.intellij.openapi.fileEditor.FileDocumentManager;
+import com.intellij.openapi.fileEditor.ex.IdeDocumentHistory;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.openapi.vfs.VirtualFile;
@@ -74,6 +75,7 @@ public class MarkGroup {
   public void saveJumpLocation(@NotNull Editor editor) {
     addJump(editor, true);
     setMark(editor, '\'');
+    IdeDocumentHistory.getInstance(editor.getProject()).includeCurrentCommandAsNavigation();
   }
 
   /**


AlexPl292:
Hi @pmnoxx !
As I see, there are still some improvements needed to close this issue, but it works better now. Thank you for your PR!

pmnoxx:
Hi @AlexPl292

I also figured how to make ctrl-o, ctrl-O call back / forward actions in intellij. I can make a PR for it.

In order to fully support jumplist (:jumps operation), printing jumplist. I would need to figure out how to get jumplist from intellij. I couldn't figure it out yet. I spend a few hours trying. Then it would be possible to remove internal jumplist data structure completely.

AlexPl292:
Yep, great! Just make a PR and we'll discuss the details.

Regarding the internal structure of jumplist. Yes, this would be great if we could get rid of this structure, but before that, we should make sure that IJ jumps work the same as vim jumps. Unfortunatelly, we haven't made any investigations on this question yet.

kjyv:
So, I'm running 0.59 and don't see this working. Any settings necessary for the jumplists to be joined?

PetrusZ:
Not see this working on 0.61 too.

6a6ai:
it's broken. Please seriously consider using neovim as an embedded editor and retiring ideavim, I and a few other paying users are seriously upset by that and considering dropping the IDE in favour of other solutions.

pmnoxx:
Currently we write both events from intelij and ideavim to Idea's `jumplist`.

Some one options is to rebind U, u to point to Back, Forward inside intellij.

I can create a PR for ideavim, where pressing u/U triggers Intellij back/forward actions.
The only downside of this solution is that `:jumps` will not work properly, because I couldn't figure out how to access Idea's jump list from ideavim.

I know how to trigger the following actions.
1) add element to Idea's jump list (currently added to ideavim)
2) go forward to next place on Idea's jumplist (I can PR for this)
3) go backward to previous place on Idea's jumplist (I can add PR for this)

To implement this functionally fully, that is to support ':jumps'. I would have to figure out how to get list of jumps from Idea. I couldn't figure it out yet. Maybe I should ask someone on JetBrains forum.

pmnoxx:
If it would get accepted I can make a PR for replacing ideavim 'u', 'U' actions with Idea's Back, Forward.

pmnoxx:
I'll try to create a PR during the weekend.

6a6ai:
> I'll try to create a PR during the weekend.

why not bind to ctrl+o, ctrl+i  - the standard vim commands for jumping back and forward? The problem now is that ctrl-o/ctrl-i are functioning within the file, until you select another file and move around there - then you end up in random places if you try to use them. 

I tried rebinding them to  :action Back & :action Forward -> didn't work propertly as well.

isudfv:
On windows ideavim jumplist is not propagated to intelij navigaration history so `<Action>(Back)` doesn't work in cases like `2j`

PR #225, opened 2020-02-20
Implement cmap

sumoooru2:
I implemented c[nore]map partially. It would be ideal if I could fully, but it seems to need a lot of work.
It's partial because

- cmap can't be set in the command line unless it's written in ideavimrc or listed in ExKeyBindings.kt
- cmap only supports one-to-many keycaps, e.g. `cmap <C-a><C-b> <Home>` is impossible

### DIFF
--- a/src/com/maddyhome/idea/vim/KeyHandler.java
+++ b/src/com/maddyhome/idea/vim/KeyHandler.java
@@ -45,6 +45,7 @@ import com.maddyhome.idea.vim.handler.EditorActionHandlerBase;
 import com.maddyhome.idea.vim.helper.*;
 import com.maddyhome.idea.vim.key.*;
 import com.maddyhome.idea.vim.option.OptionsManager;
+import com.maddyhome.idea.vim.ui.ExEntryPanel;
 import com.maddyhome.idea.vim.ui.ShowCmd;
 import org.jetbrains.annotations.Contract;
 import org.jetbrains.annotations.NonNls;
@@ -53,6 +54,7 @@ import org.jetbrains.annotations.Nullable;
 
 import javax.swing.*;
 import java.awt.*;
+import java.awt.event.InputEvent;
 import java.awt.event.KeyEvent;
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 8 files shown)


AlexPl292:
Hi, thank you for the contribution! That's would nice if you could add some tests to prove the changes. Also, I didn't completely understand why it's possible to add only one-to-many mappings?

sumoooru2:
The reason for only one-to-many mappings is that the command line mode uses different way to implement key mappings from other modes. Because of the difference it's difficult to implement many-to-many mappings, but I'm trying to implement it.

sumoooru2:
Hi, I implemented cmap and removed "partially" from the title. I added unit tests.

src/com/maddyhome/idea/vim/command/CommandBuilder.kt
```diff
@@ -24,6 +27,9 @@ class CommandBuilder(private var currentCommandPartNode: CommandPartNode) {
   var expectedArgumentType: Argument.Type? = null
     private set
 
+  var prevExpectedArgumentType: Argument.Type? = null
+    private set
+
```
AlexPl292, line 32: Is this value presented to properly support some commands with digraphs? E.g. `dt<C-k>o:`.

src/com/maddyhome/idea/vim/KeyHandler.java
```diff
@@ -753,8 +763,10 @@ private void startWaitingForArgument(Editor editor,
         // the key handler when it's complete.
         if (action instanceof InsertCompletedDigraphAction) {
           editorState.startDigraphSequence();
+          setPromptCharacterEx('?');
```
AlexPl292, line 766: Is this code supposed to insert `?` character under the caret when the user inserts digraph? IIRC, this functionality already works, isn't it?

AlexPl292 (commented):
Hi @sumoooru2!
Thank you for your updates. Now I really want to merge the changes into the repository. Hope we can make it this time.
Few questions regarding the pr:
1) Do I understand correctly that this PR fixes also some issue with literals and digraphs in command mode?
2) Do I understand correctly that the main idea is to process keys from command line using KeyHandler and not `processExKey`?

sumoooru2:
> 1. Do I understand correctly that this PR fixes also some issue with literals and digraphs in command mode?

No, I re-introduced some features about literals and digraphs (e.g. show '?' or '^'). They used to be done by Actions in ExEditorKit, which is no longer used.
> 2. Do I understand correctly that the main idea is to process keys from command line using KeyHandler and not `processExKey`?

Yes, but processExKey is still used by KeyHandler.

src/com/maddyhome/idea/vim/KeyHandler.java
sumoooru2: Yes, but using KeyHandler changed the logic and I needed to implement it here

src/com/maddyhome/idea/vim/command/CommandBuilder.kt
sumoooru2: Yes, it is. Without it , e.g. testDeleteToDigraph will fail.

sumoooru2:
This page shows "All checks have failed". I run the test "DeleteVisualLinesEndActionTest" locally several times, but it succeeds 🤔

AlexPl292:
Awesome, thank you for your work! Looks like ex panel needs some polishing, but this is definitely out of this scope.

sumoooru2:
Thank you for reviewing and merging!

PR #226, opened 2020-02-20
Implement the vim-textobj-entire plugin emulation.

agrison:
The goal of this PR is to add the **vim-textobj-entire** plugin emulation.

This Vim plugin adds two useful text-objects:
* `ae` which represents the entire content of a buffer
* `ie` which represents the entire content of a buffer without the leading and trailing spaces.

You can then use `gUae` to transform the whole text buffer content in uppercase, whereas normally you would have to use `gggUG`. This text-object has the advantage that you can use it from anywhere your cursor is currently located, you don't have to be conscious of the cursor position.

### DIFF
--- a/src/com/maddyhome/idea/vim/extension/entiretextobj/VimEntireTextObjExtension.java
+++ b/src/com/maddyhome/idea/vim/extension/entiretextobj/VimEntireTextObjExtension.java
@@ -1,0 +1,153 @@
+/*
+ * IdeaVim - Vim emulator for IDEs based on the IntelliJ platform
+ * Copyright (C) 2003-2020 The IdeaVim authors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package com.maddyhome.idea.vim.extension.entiretextobj;
+
+import com.intellij.openapi.actionSystem.DataContext;
+import com.intellij.openapi.editor.Caret;
+import com.intellij.openapi.editor.Editor;
+import com.maddyhome.idea.vim.command.*;
+import com.maddyhome.idea.vim.common.TextRange;
+import com.maddyhome.idea.vim.extension.VimExtension;
+import com.maddyhome.idea.vim.extension.VimExtensionHandler;
+import com.maddyhome.idea.vim.handler.TextObjectActionHandler;
+import com.maddyhome.idea.vim.listener.SelectionVimListenerSuppressor;
+import com.maddyhome.idea.vim.listener.VimListenerSuppressor;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.EnumSet;
+
+import static com.maddyhome.idea.vim.extension.VimExtensionFacade.putExtensionHandlerMapping;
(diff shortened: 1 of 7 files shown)


agrison:
Any feedback?

src/com/maddyhome/idea/vim/extension/entiretextobj/VimEntireTextObjExtension.java
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  @Override
+  public void init() {
+    putExtensionHandlerMapping(MappingMode.XO, parseKeys("<Plug>IncludingLeadingTrailing"), getOwner(),
+      new VimEntireTextObjExtension.EntireHandler(false), false);
+    putExtensionHandlerMapping(MappingMode.XO, parseKeys("<Plug>IgnoringLeadingTrailing"), getOwner(),
+      new VimEntireTextObjExtension.EntireHandler(true), false);
+
+    putKeyMapping(MappingMode.XO, parseKeys("ae"), getOwner(), parseKeys("<Plug>IncludingLeadingTrailing"), true);
```
AlexPl292: These commands should be renamed to `<Plug>(textobj-entire-a)` and `<Plug>(textobj-entire-i)`. [link](https://github.com/kana/vim-textobj-entire/blob/master/doc/textobj-entire.txt#L68)

src/com/maddyhome/idea/vim/extension/entiretextobj/VimEntireTextObjExtension.java
```diff
@@ -0,0 +1,153 @@
...
+ * See also the reference manual for more details at:
+ * https://github.com/kana/vim-textobj-entire/blob/master/doc/textobj-entire.txt
+ *
+ * @author Alexandre Grison (@agrison)
+ */
+public class VimEntireTextObjExtension implements VimExtension {
+  @Override
+  public @NotNull
+  String getName() {
+    return "entiretextobj";
```
AlexPl292: Seems like the plugin is called `vim-textobj-entire`. Should we name this one as`textobj-entire`?

src/com/maddyhome/idea/vim/extension/entiretextobj/VimEntireTextObjExtension.java
```diff
@@ -0,0 +1,153 @@
...
+        if (ignoreLeadingAndTrailing) {
+          String content = editor.getDocument().getText();
+          for (int i = 0; i < content.length(); ++i) {
+            if (!Character.isWhitespace(content.charAt(i))) {
+              start = i;
+              break;
+            }
+          }
+
+          for (int i = content.length() - 1; i >= 0; --i) {
```
AlexPl292: Isn't iteration till `start` enough? `i >= start` ?

src/com/maddyhome/idea/vim/extension/entiretextobj/VimEntireTextObjExtension.java
agrison: Ok didn't know there was a convention for this, I'll rename it :)

src/com/maddyhome/idea/vim/extension/entiretextobj/VimEntireTextObjExtension.java
agrison: Should totally be named `textobj-entire`, I'll rename it also in the `README`

src/com/maddyhome/idea/vim/extension/entiretextobj/VimEntireTextObjExtension.java
agrison: Nice catch :+1:

src/com/maddyhome/idea/vim/extension/entiretextobj/VimEntireTextObjExtension.java
AlexPl292: No, there is no convention, but this is a command name from the original implementation. I'm pretty sure that 99.9% of users don't care about the name, but if anybody will change the default mapping (`ae`) to something else via `omap aentire <Plug>(textobj-entire-a)`, it would be nice if it would work in IdeaVim as well.

AlexPl292:
Yes, great! Thank you for the contribution, merged.

PR #227, opened 2020-03-04
Add support for buffer command

angelbot:
This adds support for the buffer command. It supports both numbered and named buffers. Named buffers work with partial match but no tab complete.


AlexPl292:
Hi @angelbot !
Thank you for your contribution to the project!

PR #228, opened 2020-03-18
fixed #VIM-570

kevin70:
When searching (using /) Chinese characters are displayed as a box surrounded by question marks.
[https://youtrack.jetbrains.com/issue/VIM-570](https://youtrack.jetbrains.com/issue/VIM-570)


AlexPl292:
Hi! Thank you for the contribution. I actually thought that I've fixed this issue in 0c77b320dbeb36122aea13382ae03e194e420662, but it might be I missed something. Where do you experience this problem with the build from the master?

kevin70:
@AlexPl292 ideaVIM-0.55

kevin70:
@AlexPl292 
Hello, I still encounter this problem in the master snapshot version.

AlexPl292:
Okay, it's strange that I've skipped this case. Thank you for the heads up!

AlexPl292:
Hi @kevin70 ! Could you please take a look at this discussion so we can use your change under the new MIT license? Thank you!

PR #229, opened 2020-03-21
VIM-921 vim-exchange plugin emulation

fan-tom:
Closes [VIM-921](https://youtrack.jetbrains.com/issue/VIM-921)
Adds emulation of [vim-exchange](https://github.com/tommcdo/vim-exchange) plugin

### DIFF
--- a/src/com/maddyhome/idea/vim/extension/exchange/VimExchangeExtension.kt
+++ b/src/com/maddyhome/idea/vim/extension/exchange/VimExchangeExtension.kt
@@ -25,6 +25,7 @@ import com.intellij.openapi.editor.LogicalPosition
 import com.intellij.openapi.editor.colors.EditorColors
 import com.intellij.openapi.editor.markup.HighlighterLayer
 import com.intellij.openapi.editor.markup.HighlighterTargetArea
+import com.intellij.openapi.editor.markup.RangeHighlighter
 import com.intellij.openapi.util.Key
 import com.maddyhome.idea.vim.VimPlugin
 import com.maddyhome.idea.vim.command.CommandState
@@ -68,8 +69,18 @@ class VimExchangeExtension: VimExtension {
     const val EXCHANGE_CLEAR_CMD = "<Plug>ExchangeClear"
     const val EXCHANGE_LINE_CMD = "<Plug>ExchangeLine"
     val EXCHANGE_KEY = Key<Exchange>("exchange");
-    class Exchange(val type: CommandState.SubMode, val start: Mark, val end: Mark, val text: String)
+    class Exchange(val type: CommandState.SubMode, val start: Mark, val end: Mark, val text: String) {
+      private var myHighlighter: RangeHighlighter? = null
+      fun setHighlighter(highlighter: RangeHighlighter) {
+        myHighlighter = highlighter
+      }
+      fun getHighlighter(): RangeHighlighter? = myHighlighter
+
+    }
     fun clearExchange(editor: Editor) {
+      editor.getUserData(EXCHANGE_KEY)?.getHighlighter()?.let {
+        editor.markupModel.removeHighlighter(it)
+      }
       editor.putUserData(EXCHANGE_KEY, null)
     }
   }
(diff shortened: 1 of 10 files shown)


fan-tom:
Seems that tests are broken because there is no vim registers cleanup in seuUp/tearDown and in one test (`test replace with empty register`) last register is not empty because `testYankEntireBufferIgnoreLeadingTrailing` fills it previously. Cannot reproduce it locally because gradle doesn't allow to set test execution order

yahya-abou-imran:
This is a very good idea! Hope it will be merged soon.

src/com/maddyhome/idea/vim/extension/exchange/VimExchangeExtension.kt
```diff
@@ -0,0 +1,331 @@
...
+    putExtensionHandlerMapping(MappingMode.N, parseKeys(EXCHANGE_LINE_CMD), owner, ExchangeHandler(true), false)
+
+    putKeyMapping(MappingMode.N, parseKeys("cx"), owner, parseKeys(EXCHANGE_CMD), true)
+    putKeyMapping(MappingMode.X, parseKeys("X"), owner, parseKeys(EXCHANGE_CMD), true)
+    putKeyMapping(MappingMode.N, parseKeys("cxc"), owner, parseKeys(EXCHANGE_CLEAR_CMD), true)
+    putKeyMapping(MappingMode.N, parseKeys("cxx"), owner, parseKeys(EXCHANGE_LINE_CMD), true)
+  }
+
+  private companion object {
+    const val EXCHANGE_CMD = "<Plug>Exchange"
```
AlexPl292, line 68: The original implementation uses parentheses around the command name (`<Plug>(Exchange)`). Could you please update the names to be the same as in the vim plugin?

src/com/maddyhome/idea/vim/group/ChangeGroup.java
```diff
@@ -1751,7 +1751,7 @@ private boolean deleteText(final @NotNull Editor editor,
       if (type != null) {
         final int start = range.getStartOffset();
         VimPlugin.getMark().setMark(editor, MarkGroup.MARK_CHANGE_POS, start);
-        VimPlugin.getMark().setChangeMarks(editor, new TextRange(start, start));
+        VimPlugin.getMark().setChangeMarks(editor, new TextRange(start, start+1));
```
AlexPl292, line 1754: I think we should add some tests to prove all the +1/-1 changes. Or does IdeaVim already have enough tests for marks?

test/org/jetbrains/plugins/ideavim/extension/exchange/VimExchangeExtensionTest.kt
```diff
@@ -0,0 +1,113 @@
...
+class VimExchangeExtensionTest: VimTestCase() {
+  @Throws(Exception::class)
+  override fun setUp() {
+    super.setUp()
+    enableExtensions("exchange")
+  }
+  // |cx|
+  fun `test exchange words left to right`() {
+    doTest(StringHelper.parseKeys("cxe", "w", "cxe"),
+      "The quick <caret>brown fox catch over the lazy dog",
```
AlexPl292, line 35: JFI, `VimTestCase` has some constants like `c` that can be used instead of `<caret> in the tests. E.g. this like can look like `"The quick ${c}brown fox catch over the lazy dog"`. It doesn't provide any features except that it's now highlighted in IJ and thus easy to recognize.

fan-tom:
> Hey, the plugin looks really useful, great!
> Could you please check some of my comments?
> Also:
>
>     1. If I'm correct, at the moment the dot command doesn't work with this plugin. You can use `isRepeatable` in handlers to enable it. Let me know if you'll experience any problems with it.
>
>     2. Do I understand it right that `<esc>` should not finish work after first motion (e.g. `cxiw<esc>`)?

2. Yes, `<esc>` doesn't leave "exchange mode". To exit "exchange mode", you have to use `cxc` command.

1. I tried to simply set `isRepeatable() = true` but realized that the `VimRepeater.lastChangeCommand` command remembered after, for example, `cxiw`, is yanking visually selected text, which is part of exchange implementation. I checked that vim doesn't remember yanking it the same case (and at all, I suppose), so I think this is a bug in ideavim.

AlexPl292:
Okay, in this case, we should fix `isRepeatable` behavior to make it work with this extension.
Yep, great, I think we can merge this PR. Thank you very much for contribution! 🚀

AlexPl292:
JFI: I've just found that the original plugin keeps the list of emulations. Added this PR to the list: https://github.com/tommcdo/vim-exchange#other-text-editors

PR #230, opened 2020-03-22
VIM-1924

fan-tom:
Closes [VIM-1924](https://youtrack.jetbrains.com/issue/VIM-1924) by returning VISUAL_CHARACTERWISE from autodetectVisualSubmode if 'Add selection for next occurrence' was performed previously. I'm however not sure if it is the right way to fix that issue


AlexPl292:
Hmm, well, yeah, doesn't look like a proper solution, but it should work :)
Thank you for the contribution!

PR #231, opened 2020-04-07
Support smooth scrolling

citizenmatt:
Instead of maintaining state to enable/disable smooth scrolling per-editor, when IdeaVim is enabled/disabled, this PR means IdeaVim supports smooth scrolling. All scrolling such as `<C-F>`  will be handled correctly, and the page is animated, as when IdeaVim is disabled. The "Enable smooth scrolling" option in _Preferences | Editor | General_ is now correctly respected by IdeaVim.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -54,9 +54,6 @@ _Available since 0.55.3 EAP:_
 * [VIM-434](https://youtrack.jetbrains.com/issue/VIM-434) Add `'showcmd'` support, on by default
 * `ReplaceWithRegister` plugin emulation ([ReplaceWithRegister](https://www.vim.org/scripts/script.php?script_id=2703))
 
-**Changes:**
-* Support IntelliJ's smooth scrolling. Use "Enable smooth scrolling" checkbox in _Preferences | Editor | General_ to disable.
-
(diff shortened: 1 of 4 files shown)


AlexPl292:
Yep, I like how does it work. Merged :)

LifeIsStrange:
@AlexPl292 will IdeaVim have soon a point stable release in order for users to benefits from this change as soon as possible ?

AlexPl292:
Hi! We don't have any plans for the stable release at the moment. You can join EAP, and you'll get this change in about a week. 
https://github.com/JetBrains/ideavim#get-early-access

LifeIsStrange:
@AlexPI292 thanks, just to be clear:
Is it recommended to be on Idea EAP in order to have ideavim EAP or does Ideavim EAP has the same "stability guarantee" with stable IntelliJ Idea?

AlexPl292:
@LifeIsStrange you can use IdeaVim EAP with the stable version of IJ. As described in "Get Early Access" section, EAP versions don't guarantee your stability of the plugin, but practically the quality of EAP builds is quite high and it can be used on a daily basis.

PR #234, opened 2020-04-16
[VIM-1991] fix >0 number register not work

runforprogram:
fix when use  >0  number register, treat it as a [count].but it is a  register.


AlexPl292:
Hi! Thank you for the contribution, the fix will be shipped with the nearest release.

AlexPl292:
Hi @runforprogram ! Could you please take a look at this discussion so we can use your change under the new MIT license? Thank you!

runforprogram:
> Hi @runforprogram ! Could you please take a look at this discussion so we can use your change under the new MIT license? Thank you!
> 
> #542

ok

PR #237, opened 2020-04-30
Reset operator-pending commands on escape

adriafarres:
Suppose we have the text "|Hello, world", with | being the position of
the cursor. Pressing `d<Esc>dw` simply moves the cursor on top of the
comma instead of leaving the text as ", world".

This fixes issue VIM-1421.

I'm creating the pull request again, since on rebasing I apparently messed up the history.

### DIFF
--- a/src/com/maddyhome/idea/vim/KeyHandler.java
+++ b/src/com/maddyhome/idea/vim/KeyHandler.java
@@ -212,13 +212,15 @@ public class KeyHandler {
     handleKeyRecursionCount++;
 
     try {
+      if (isEditorReset(key, editorState)) {
+        handleEditorReset(editor, key, context, editorState);
+      }
+
       if (!allowKeyMappings || !handleKeyMapping(editor, key, context)) {
         if (isCommandCountKey(chKey, editorState)) {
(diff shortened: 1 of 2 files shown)


AlexPl292:
Hi @adriafarres !
Thank you for your attention to the project!
At the moment it looks like your PR doesn't handle some cases where `<Esc>` is mapped to any other key, or pressing `<C-[>`. Could you take a look at `handleEditorReset` and `isEditorReset` functions and it's usages? It might be a better place for this fix, what do you think?

adriafarres:
You are absolutely right, Alex. In fact, this case was already being handled and I missed it. My apologies. Moving it outside of the condition seems to fix it. Let me know if there's a better way to handle this or if you would like me to add more tests.

AlexPl292:
This might be a proper solution. And yep, I would appreciate if you could add more test cases to prove the change. I think, also `<C-[>`, different commands and mappings like `map <C-D> <Esc>` should be tested as well.

AlexPl292:
We've got more tests here: https://github.com/JetBrains/ideavim/pull/241
Thank you for the contribution! :)

PR #238, opened 2020-05-13
Prioritise startup activity to initialise IdeaVim early

citizenmatt:
Initialise early in the post startup activities so that any open editors get Vim features as soon as possible. This is most important on Rider, which has a post startup activity that blocks until Rider's project model is ready. Unfortunately, it also blocks plugin startup activities.

All core startup activities are unordered and don't have IDs, so all we can do is go first.


AlexPl292:
Thank you for change. I hope rider will work better with it.

PR #240, opened 2020-06-21
add support for putting & editing macros

rieonke:
Add support for putting and editing a macros
https://youtrack.jetbrains.com/issue/VIM-2059

Currently we can't edit a recorded marco just like what we can do in classic vim.

What this PR does:

1. put a recorded macro to current buffer by converting key strokes to printable characters with `p` or `:put`
2. convert text to a collection of key strokes while playing back a macro

Effects :
1. can put recorded macro as printable character
2. support for editing macro (put from and yank to register)
3. can play back any register as a macro even if it's not a recorded macro, eg from yanking

I'm a idea & vim fanatic, ideavim saved my days! Thank you guys!

### DIFF
--- a/src/com/maddyhome/idea/vim/action/copy/PutTextAction.kt
+++ b/src/com/maddyhome/idea/vim/action/copy/PutTextAction.kt
@@ -25,6 +25,7 @@ import com.maddyhome.idea.vim.command.Command
 import com.maddyhome.idea.vim.group.copy.PutData
 import com.maddyhome.idea.vim.group.copy.PutData.TextData
 import com.maddyhome.idea.vim.handler.ChangeEditorActionHandler
+import com.maddyhome.idea.vim.helper.StringHelper
 
 sealed class PutTextBaseAction(
   private val insertTextBeforeCaret: Boolean,
@@ -35,7 +36,7 @@ sealed class PutTextBaseAction(
 
   override fun execute(editor: Editor, context: DataContext, count: Int, rawCount: Int, argument: Argument?): Boolean {
     val lastRegister = VimPlugin.getRegister().lastRegister
-    val textData = if (lastRegister != null) TextData(lastRegister.text, lastRegister.type, lastRegister.transferableData) else null
+    val textData = if (lastRegister != null) TextData(lastRegister.text ?: StringHelper.toKeyNotation(lastRegister.keys), lastRegister.type, lastRegister.transferableData) else null
     val putData = PutData(textData, null, count, insertTextBeforeCaret, indent, caretAfterInsertedText, -1)
     return VimPlugin.getPut().putText(editor, context, putData)
   }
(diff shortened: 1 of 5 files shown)


rieonke:
Hello , I'm stuck with the `non-printable` characters.

In classic vim, the `non-printable` characters are stored as a naive ascii code, and rendered specially in buffer. 

For example, `ESC`: stored as raw ascii code `0x1B`, but rendered as `^[` and treated as a single character(select/edit/delete them as a monomer), while yanking the `^[`, you'll get an single ascii code `0x1b` instead of two char `^` and `[` in register.

But now, what the idea vim does is treating them as separate chars by converting them to printable characters directly. It introduced an ambiguity problem. Think about what happens if you have a yanked macro in a register:
```
A^[^[p
```
Then, does it mean that:
```
1. A   # insert at last
2. ^   # insert a char '^'
3. [   # insert a char '['
4. ^[  # ESC back to normal mode
5. p   # put
```
or like this:
```
1. A    # insert at last
2. ^[   # ESC back to normal mode
3. ^    # goto the the first non-blank char of line
4. [p   # put before the cursor
```
or something else?

So  what should we do ? 
- store it as raw ascii code, render and edit it just like the classic way
- use different and unambiguous way to display and edit those non-printable characters, such as `<ESC> <BS> <C-R>` instead of `^[ ^H ^R`
- or something better

Thanks!

AlexPl292:
Hey, yep, there is an annoying problem in IJ that it doesn't render non-printable chars. So, we just can't print `^[` as a single character. So, it seems like we don't have any other opportunities than use `<Esc>`, `<BS>` form.

rieonke:
Hello, This PR is all ready.

Behaviors effected:
- ex command `reg`: display non-printable char as vim notation, `^[` => `<C-[>`

Thanks!

AlexPl292:
Yay, thank you for the contribution. We have a kind of pre-release freeze, so I'll check the PR right after the release!

AlexPl292:
Hey, I've just reviewed the changes and I'm going to merge the changes. But I just wanted to check what is the reason for changing the behaviour of `:reg` command? Do you think it would be more convenient?

rieonke:
Hello, usually we use reg command to view all the registers and select one which should be edited, it’s strange if what we saw in registers panel is totally different from the putted line. besides, the edited macro is saved as raw text (by yanking) instead of KeyStrokes, so it would look messy if we mixd the two modes (`^[` for original key strokes macro and `<Esc>` for edited macro)  . So I think it would be better to make them look the same.

Thanks!

AlexPl292:
Okay, I've ended up with merging the PR (of course), but I'll revert the behaviour of `:reg` and update it to render all characters in `^J` form. The reasons for that are: 1) At the moment not all chars are converted. At least new line char isn't converted to `<C-J>` form and it looks a bit ugly and confusing to see two forms of keystrokes in one `:reg` panel. 2) This definitely should be an ability to enable the previous behaviour. I wasn't able to find a vim option that switches between `<C-J>` and `^J` forms, so it looks like we should add a specific one for IdeaVim.

However, I agree with your point and I'll think about how to make it better. If you'll have any thoughts / PRs on this topic, you're welcome.

PR #241, opened 2020-06-25
Properly reset when pressing Esc in the middle of command

patrick-elmquist:
This is a continuation on Adriafarres pull request (#237) since that PR has been idle for the past month and this bug really annoys me. I'll just quote the previous issue description:

> Suppose we have the text "|Hello, world", with | being the position of
the cursor. Pressing d<Esc>dw simply moves the cursor on top of the
comma instead of leaving the text as ", world".
>
> This fixes issue VIM-1421

I've added some more test cases based on the last comment in the other PR.
Please let me know what you think :)

### DIFF
--- a/src/com/maddyhome/idea/vim/KeyHandler.java
+++ b/src/com/maddyhome/idea/vim/KeyHandler.java
@@ -212,13 +212,15 @@ public class KeyHandler {
     handleKeyRecursionCount++;
 
     try {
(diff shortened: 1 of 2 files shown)


AlexPl292:
Oh, I like the level of cooperation :)
We have a pre-release freeze, so I'll merge the both PRs right after the release.

AlexPl292:
Yep, merged. Thank you again for the contribution!

PR #242, opened 2020-06-26
Partial implementation of virtualedit config

i-e-b:
This does not support all config settings,
but does add the 'onemore' option.

This partly addresses https://youtrack.jetbrains.com/issue/VIM-844

### DIFF
--- a/src/com/maddyhome/idea/vim/action/motion/leftright/MotionArrowRightAction.kt
+++ b/src/com/maddyhome/idea/vim/action/motion/leftright/MotionArrowRightAction.kt
@@ -27,6 +27,8 @@ import com.maddyhome.idea.vim.command.Argument
 import com.maddyhome.idea.vim.command.MotionType
 import com.maddyhome.idea.vim.handler.NonShiftedSpecialKeyHandler
 import com.maddyhome.idea.vim.helper.StringHelper.parseKeys
+import com.maddyhome.idea.vim.helper.commandState
+import com.maddyhome.idea.vim.helper.isEndAllowed
 import java.awt.event.KeyEvent
 import javax.swing.KeyStroke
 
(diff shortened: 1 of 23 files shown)


AlexPl292:
Yep, thanks for the contribution. We have a pre-release freeze, so I'll check the PR right after the release.

Are you going to support other settings, or you're interested in this particular one? The both answers are perfectly fine, but I just want to understand your goals :)

i-e-b:
My primary purpose was to support `onemore`, mainly as it makes mouse-based copy-and-paste actions much more intuitive.
If I have time later, I might try the other settings, but there is an interaction between Vim settings and Idea virtual space settings I'd have to work out.

i-e-b:
@AlexPl292 : Updated to fix conflicts.

AlexPl292:
HI! I've just checked the PR. Unfortunately, it looks like it would be a bit more complicated. Here are some inconsistencies I was able to find:
- `D` command after the end of the line
- `D` command in the middle of the line
- `$` command and moving to a shorter line.



Could you please check these cases and write some tests to prove the implementation?

Also, it looks like there are some unnecessary files in the PR.
But overall, the implementation looks good. Seems like it can be merged after fixing these small issues.

kamolsrisan:
good

kamolsrisan:
good

i-e-b:
Thanks for the detailed feedback. I'll try to correct those issues as soon as possible.

i-e-b:
Hi @AlexPl292 .
The PR should be cleaned up and have tests that match your description above.

It turns out the interaction between `$` and visual mode are complex, and IdeaVim doesn't currently match Vim 8.x
Rather than try to re-work everything, I've tried to end up with the most reasonable behaviour for `onemore` that doesn't break existing IdeaVim behaviour.

i-e-b:
Updated against upstream. As far as I'm aware, there are no other changes required?

AlexPl292:
Hi! I was able to perform some more aggressive refactorings that allowed me to reach the desired behaviour without additional workarounds. I hope this won't give us any other difficulties.
Thank you for your contribution! This was really a big change!

PR #243, opened 2020-06-30
Fix alternative XDG path in README

strogiyotec:
README shows this path that follows XDG standart
```
$XDG_CONFIG_HOME$/ideavim/ideavimrc
```
It should be 
```
$XDG_CONFIG_HOME/ideavim/ideavimrc
```
Because be default **$XDG_CONFIG_HOME** points to $HOME/.config and I can to `cd $XDG_CONFIG_HOME` but `cd $XDG_CONFIG_HOME$`  will show this error `no such file or directory: /home/username/.config$`


AlexPl292:
Yep, thank you for the fix!

PR #244, opened 2020-07-02
Implement tabc[lose] ex command

rieonke:
Hello, 
This PR implemented the `tabc[lose]` ex command to close a specific tab.
https://youtrack.jetbrains.com/issue/VIM-2068

---
🍺 usages:
```
:tabc[lose]   Close current tab page
:tabc[lose]{count}  Close tab page {count}. just like the classic vim:
         :tabclose -2    " close the two previous tab page
         :tabclose +     " close the next tab page
         :tabclose 3     " close the third tab page
         :tabclose $     " close the last tab page
```

🔔 notices:
- If there is only one tab page on the current active tabbed pane, then this tab will be closed, which is different from the classic vim.
- If the `{count}` parameter is out of range, then the nearest head or tail pane will be closed.
- If current tab has been closed, the focus state will be transfered to next tab, otherwise the current tab remains focus state

❓ questions: 
How can I write a test for UI related functions? I can't get the `EditorWindow` object in test case, it's always null.

Thanks!


AlexPl292:
Great, thank you for the contribution! Unfortunatelly, UI tests is a complicated question for IdeaVim (and for IJ, I guess). I'll try to find a way to test it and notify you if I find something.

AlexPl292:
Also, I've updated the syntax a bit. If you're interested: 26aa753d9e4029429b7f209b2c7fab951e7d5dc8

PR #245, opened 2020-07-18
VIM-1970 | Working solution of plugin vim-highlightedyank

KostkaBrukowa:
Hey!
Before i go in I'd like to mention, that I've never contributed to any project before so if i did something wrong things please let me know :)
So, I've found issue ([VIM-1970](https://youtrack.jetbrains.com/issue/VIM-1970)) on plugin that is really simple but makes vim a little bit more pleasurable. This is vim-highlightedyank. I've been using it with Vim and it would be nice to have it in ideavim.
There was no response since that question and I decided to implement it myself.
Based on your advice (`or create a draft PR to indicate that you've started working`) I've created a draft PR to let you know that I've started working on this.

Right now I believe this draft is overall working version of the plugin. I was thinking if you have any thoughts on this solution.

It's still a draft and the next steps I would like to take are:
1. make it an extension and move this code to a different package
2. make this plugin optional (to make it possible to enable it via `set vim-highlightedyank`
3. make a duration of the highlight customizable (e.g. `let g:highlightedyank_highlight_duration = 1000`)
4. maybe make a highlight color customizable (don't know if it is possible)
5. of course write some tests to it.

What do you think about it?

### DIFF
--- a/doc/emulated-plugins.md
+++ b/doc/emulated-plugins.md
@@ -69,3 +69,14 @@ Available extensions:
 * Emulates [vim-textobj-entire](https://github.com/kana/vim-textobj-entire)
 * Additional text objects: `ae`, `ie`
 * By [Alexandre Grison](https://github.com/agrison)
+
+## highlightedyank
+
+* Setup: 
+    * `set highlightedyank`
+    * if you want to optimize highlight duration, assign a time in milliseconds:  
+       `let g:highlightedyank_highlight_duration = "1000"`  
+       A negative number makes the highlight persistent.  
+       `let g:highlightedyank_highlight_duration = "-1"`
+* Emulates [vim-highlightedyank](https://github.com/machakann/vim-highlightedyank)
+* By [KostkaBrukowa](https://github.com/KostkaBrukowa)
--- a/doc/emulated-plugins.md
+++ b/doc/emulated-plugins.md
@@ -78,5 +78,7 @@ Available extensions:
        `let g:highlightedyank_highlight_duration = "1000"`  
        A negative number makes the highlight persistent.  
        `let g:highlightedyank_highlight_duration = "-1"`
+     * if you want to change background color of highlight you can provide the rgba of the color you want e.g.  
+       `let g:highlightedyank_highlight_color = "rgba(160, 160, 160, 155)"`
 * Emulates [vim-highlightedyank](https://github.com/machakann/vim-highlightedyank)
 * By [KostkaBrukowa](https://github.com/KostkaBrukowa)
--- a/doc/emulated-plugins.md
+++ b/doc/emulated-plugins.md
@@ -70,7 +70,7 @@ Available extensions:
 * Additional text objects: `ae`, `ie`
 * By [Alexandre Grison](https://github.com/agrison)
 
-## highlightedyank
+## highlightedyank [To Be Released]
 
 * Setup: 
     * `set highlightedyank`
--- a/src/com/maddyhome/idea/vim/extension/highlightedyank/VimHighlightedYank.kt
+++ b/src/com/maddyhome/idea/vim/extension/highlightedyank/VimHighlightedYank.kt
@@ -21,7 +21,6 @@ package com.maddyhome.idea.vim.extension.highlightedyank
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.editor.colors.EditorColors
-import com.intellij.openapi.editor.colors.TextAttributesKey
 import com.intellij.openapi.editor.markup.EffectType
 import com.intellij.openapi.editor.markup.HighlighterLayer
 import com.intellij.openapi.editor.markup.HighlighterTargetArea
@@ -31,13 +30,15 @@ import com.maddyhome.idea.vim.VimPlugin
 import com.maddyhome.idea.vim.common.TextRange
 import com.maddyhome.idea.vim.ex.vimscript.VimScriptGlobalEnvironment
 import com.maddyhome.idea.vim.extension.VimExtension
+import java.awt.Color
 import java.awt.Font
 import java.util.concurrent.Executors
 import java.util.concurrent.TimeUnit
 
 private const val DEFAULT_HIGHLIGHT_DURATION: Long = 300
 private const val HIGHLIGHT_DURATION_VARIABLE_NAME = "g:highlightedyank_highlight_duration"
-private val DEFAULT_HIGHLIGHT_TEXT_ATTRIBUTES: TextAttributesKey = EditorColors.TEXT_SEARCH_RESULT_ATTRIBUTES
+private const val HIGHLIGHT_COLOR_VARIABLE_NAME = "g:highlightedyank_highlight_color"
+private val DEFAULT_HIGHLIGHT_TEXT_COLOR: Color = EditorColors.TEXT_SEARCH_RESULT_ATTRIBUTES.defaultAttributes.backgroundColor
 
 
 /**
@@ -52,6 +53,9 @@ private val DEFAULT_HIGHLIGHT_TEXT_ATTRIBUTES: TextAttributesKey = EditorColors.
 
  * A negative number makes the highlight persistent.
  * let g:highlightedyank_highlight_duration = "-1"
+
+ * if you want to change background color of highlight you can provide the rgba of the color you want e.g.
+ * let g:highlightedyank_highlight_color = "rgba(160, 160, 160, 155)"
  *
  * When a new text is yanked or user starts editing, the old highlighting would be deleted.
  */
@@ -109,8 +113,8 @@ class VimHighlightedYank: VimExtension {
 
     private fun highlightSingleRange(editor: Editor, range: ClosedRange<Int>) {
       val textAttributes = TextAttributes(
-        DEFAULT_HIGHLIGHT_TEXT_ATTRIBUTES.defaultAttributes.foregroundColor,
-        DEFAULT_HIGHLIGHT_TEXT_ATTRIBUTES.defaultAttributes.backgroundColor,
+        null,
+        extractUsersHighlightColor(),
         editor.colorsScheme.getColor(EditorColors.CARET_COLOR),
         EffectType.SEARCH_MATCH, Font.PLAIN
       )
@@ -129,7 +133,7 @@ class VimHighlightedYank: VimExtension {
     }
 
     private fun setClearHighlightRangeTimer(editor: Editor, highlighter: RangeHighlighter) {
-      val timeout = extractHighlightDuration()
+      val timeout = extractUsersHighlightDuration()
 
       //from vim-highlightedyank docs: A negative number makes the highlight persistent.
       if(timeout >= 0) {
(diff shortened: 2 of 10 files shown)


AlexPl292:
Wow, you're right on time! A few days ago I've looked at this ticket and thought that it would be nice to have this feature in one of the next releases.
I've checked your drafts and they look fine for me. The one important thing is that you embed this functionality right into the yank group, what makes it a core functionality and not the extension of IdeaVim :) (yeah, I see your first punkt, just wanted to make it clear). To convert it in IdeaVim extension, check out how other extensions are implemented (`extension` package in the project). If you'll do it this way, you'll automatically get the second punkt of your list.
However, you may still need to modify `YankGroup` because at the moment I don't know how an extension can get a "notification" that a yank was performed. That would be awesome if we could do it in the way as vim does it.

Don't hesitate to reach me out in case of questions. Another great plugin is on the way and this is awesome!

KostkaBrukowa:
Hey @AlexPl292 !
I think I've finished the plugin and you can take a look at it.
Sadly there are two things that I didn't know how to do:
1. the fourth punkt from my list that is: customisable color. The highlight function takes the `TextAttributes` object and I haven't found the way to just create that kind of object with for example color `#c3c3c3` as a parameter, so the highlight color is not customisable. Do you think that the default highlight coloring is OK?
2. I haven't managed to test the removal of the highlight after some timeout. I've tried to do this with `Thread.sleep(...)` but is does not work within the test, and test fails (it works well on the Intelij instance).
```kotlin
  fun `test highlighting for a correct default amount of time`() {
    doTest("yiw", code, code, CommandState.Mode.COMMAND, CommandState.SubMode.NONE)

    Thread.sleep(100)
    assertAllHighlightersCount(1)
    Thread.sleep(300) // even value like 10000 does not work
    assertAllHighlightersCount(0) //says that there is still 1 highlighter
  }

  fun `test highlighting for a correct user provided amount of time`() {
    configureByJavaText(code)
    typeText(StringHelper.parseKeys(":let g:highlightedyank_highlight_duration = \"500\"<CR>"))
    typeText(StringHelper.parseKeys("yiw"))

    Thread.sleep(100)
    assertAllHighlightersCount(1)
    Thread.sleep(200)
    assertAllHighlightersCount(1)
    Thread.sleep(200)
    assertAllHighlightersCount(0)
  }
```
Do you have any thoughts on it?

If you have any comments or ideas please let me know :)

kamolsrisan:
good

kamolsrisan:
good

KostkaBrukowa:
Hey! 
I've actually managed to modify background color of the highlight from .ideavimrc :)
but again I didn't know how to check highlight color in tests :(

src/com/maddyhome/idea/vim/extension/highlightedyank/VimHighlightedYank.kt
```diff
@@ -0,0 +1,184 @@
...
+
+      setClearHighlightRangeTimer(editor, highlighter)
+    }
+
+    private fun setClearHighlightRangeTimer(editor: Editor, highlighter: RangeHighlighter) {
+      val timeout = extractUsersHighlightDuration()
+
+      //from vim-highlightedyank docs: A negative number makes the highlight persistent.
+      if(timeout >= 0) {
+        Executors.newSingleThreadScheduledExecutor().schedule({
```
AlexPl292, line 140: I suggest to use `pooledThreadSingleAlarm` from `SingleAlarm.kt`.

test/org/jetbrains/plugins/ideavim/extension/highlightedyank/VimHighlightedYankTest.kt
```diff
@@ -0,0 +1,142 @@
...
+  fun `test removing previous highlight when new range is yanked`() {
+    configureByJavaText(code)
+    typeText(StringHelper.parseKeys("yyjyy"))
+
+    assertAllHighlightersCount(1)
+    assertHighlighterRange(40, 59, getFirstHighlighter())
+  }
+
+  fun `test removing previous highlight when entering insert mode`() {
+    doTest("yyi", code, code, CommandState.Mode.INSERT, CommandState.SubMode.NONE)
```
AlexPl292, line 58: May I ask you to add `@TestWithoutNeovim(reason = SkipNeovimReason.PLUGIN)` annotation to all tests that use `doTest` function? We've recently started experimenting with neovim in testing, but it doesn't support plugins at the moment.

doc/emulated-plugins.md
```diff
@@ -69,3 +69,16 @@ Available extensions:
 * Emulates [vim-textobj-entire](https://github.com/kana/vim-textobj-entire)
 * Additional text objects: `ae`, `ie`
 * By [Alexandre Grison](https://github.com/agrison)
+
+## highlightedyank
```
AlexPl292, line 73: Please add [To Be Released] mark.
```suggestion
## highlightedyank [To Be Released]
```

kamolsrisan:
good

AlexPl292:
@kamolsrisan Hi! I understand that you like the results and want it to be merged, but could you please stop posting this message? This already looks like spam.

doc/emulated-plugins.md
kamolsrisan: good

kamolsrisan:
good

doc/emulated-plugins.md
kamolsrisan: good

src/com/maddyhome/idea/vim/extension/highlightedyank/VimHighlightedYank.kt
```diff
@@ -0,0 +1,184 @@
...
+/**
+ * @author KostkaBrukowa (@kostkabrukowa)
+ *
+ * Port of vim-highlightedyank
+ * See https://github.com/machakann/vim-highlightedyank
+ *
+ * if you want to optimize highlight duration, use g:highlightedyank_highlight_duration. Assign a time in milliseconds.
+ *
+ * let g:highlightedyank_highlight_duration = "1000"
+
```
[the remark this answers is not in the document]
kamolsrisan: good

doc/emulated-plugins.md
kamolsrisan: good

src/com/maddyhome/idea/vim/extension/highlightedyank/VimHighlightedYank.kt
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+
+  override fun dispose() {
+    super.dispose()
+    highlightEnabled = false
+  }
+
+  companion object {
+    var highlightEnabled: Boolean = false
```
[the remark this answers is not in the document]
kamolsrisan: good

src/com/maddyhome/idea/vim/extension/highlightedyank/VimHighlightedYank.kt
```diff
@@ -0,0 +1,184 @@
...
+      highlightHandler.highlightYankRange(editor, range, highlightEnabled)
+    }
+
+    fun clearAllYankHighlighters() {
+      highlightHandler.clearAllYankHighlighters()
+    }
+  }
+
+  private class HighlightHandler {
+    private val yankHighlighters: MutableSet<Pair<Editor, RangeHighlighter>> = mutableSetOf()
```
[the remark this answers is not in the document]
kamolsrisan: good

src/com/maddyhome/idea/vim/extension/highlightedyank/VimHighlightedYank.kt
```diff
@@ -0,0 +1,184 @@
...
+
+      val highlighter = editor.markupModel.addRangeHighlighter(
+        range.start,
+        range.endInclusive,
+        HighlighterLayer.SELECTION - 1,
+        textAttributes,
+        HighlighterTargetArea.EXACT_RANGE
+      )
+
+      yankHighlighters.add(Pair(editor, highlighter))
```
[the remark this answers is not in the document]
KostkaBrukowa: Hey! I've tried to follow your advice, but i guess Editor doesn't implements `Disposable` interface :(
image

src/com/maddyhome/idea/vim/extension/highlightedyank/VimHighlightedYank.kt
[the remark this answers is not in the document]
KostkaBrukowa: I'm creating multiple highlighters when user is in multi cursor mode. I haven't found the way to create highlighter with multiple ranges so I'm creating multiple highlighters
On the other hand I don't need editor for every highlighter, I just need one, so I've removed editor from set

src/com/maddyhome/idea/vim/group/copy/YankGroup.kt
```diff
@@ -172,6 +173,8 @@ class YankGroup {
                         startOffsets: Map<Caret, Int>?): Boolean {
     startOffsets?.forEach { caret, offset -> MotionGroup.moveCaret(editor, caret, offset) }
 
+    VimHighlightedYank.highlightYankRange(editor, range)
```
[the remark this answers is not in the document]
KostkaBrukowa: You may take a look if this looks OK to you.

src/com/maddyhome/idea/vim/extension/highlightedyank/VimHighlightedYank.kt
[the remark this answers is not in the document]
KostkaBrukowa: Yes it could be private, but with listener pattern it's no longer needed :)

src/com/maddyhome/idea/vim/extension/highlightedyank/VimHighlightedYank.kt
KostkaBrukowa: Hmmm... this is weird because it doesn't really work at all with `pooledThreadSingleAlarm`. I've tried it with many different options and the function inside `pooledThreadSingleAlarm` its just not called at all

test/org/jetbrains/plugins/ideavim/extension/highlightedyank/VimHighlightedYankTest.kt
KostkaBrukowa: done

doc/emulated-plugins.md
KostkaBrukowa: done

src/com/maddyhome/idea/vim/group/ChangeGroup.java
```diff
@@ -421,6 +422,8 @@ private void initInsert(@NotNull Editor editor, @NotNull DataContext context, @N
 
       VisualGroupKt.updateCaretState(editor);
     }
+
+//    VimHighlightedYank.Companion.clearAllYankHighlighters();
```
KostkaBrukowa, line 426: You might have not noticed this before but I guess this it the same case as with yank ` I think this would be a bad code to inject extension-specific calls into the core code`. Should I do the same thing as I did with yank, this is listener pattern?
There is a note in original plugin that when user enters the insert mode all highlights should be removed, so originally put the line here

test/org/jetbrains/plugins/ideavim/extension/highlightedyank/VimHighlightedYankTest.kt
```diff
@@ -0,0 +1,172 @@
...
+//    doTest("yiwi", codeWithMultipleCurors, codeWithMultipleCurors, CommandState.Mode.INSERT, CommandState.SubMode.NONE)
+//
+//    assertAllHighlightersCount(0)
+//  }
+
+  @TestWithoutNeovim(reason = SkipNeovimReason.PLUGIN)
+  fun `test highlighting for a correct default amount of time`() {
+    doTest("yiw", code, code, CommandState.Mode.COMMAND, CommandState.SubMode.NONE)
+
+    assertHappened(DEFAULT_HIGHLIGHT_DURATION.toInt(), 200) {
```
KostkaBrukowa, line 130: I've followed your advice, but the precision has to be quite large :( 200ms is I think the lowest possible amount I've found to be working. 
I've also tried to measure time locally on running IDE, and it works well, the difference between set time and actual highlight duration is max. 5ms

KostkaBrukowa: I've done some more measurements and I think the reason for this is that the `typeText` takes about 200ms...

src/com/maddyhome/idea/vim/extension/highlightedyank/VimHighlightedYank.kt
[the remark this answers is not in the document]
AlexPl292: Okay, that's right. I've replaced it with `Disposer.register(project)`. Well, anyway seems like it works correct.

src/com/maddyhome/idea/vim/extension/highlightedyank/VimHighlightedYank.kt
AlexPl292: Okay, let's don't use it :)

AlexPl292:
Hey, I've added a commit with some small changes to speed up the process. Actually, I've just added insert listener and performed some small refactorings.

Thank you very much for the contribution. This is great stuff, nice to see it in IdeaVim!

KostkaBrukowa:
I'm glad that I could add something to this really great project. Thanks

citizenmatt:
I discovered highlightedyank through this PR, and it's now a firm favourite extension in both IdeaVim and my standard Vim config. Great PR! 🎉

PR #249, opened 2020-08-31
VIM-2113 Increase tag range only in visual mode

jpalus:


AlexPl292:
Hi! Thank you for your contribution and sorry for the delay in reviewing 🎉

PR #250, opened 2020-08-31
VIM-2114 Do not override registers when deleting empty range

jpalus:
I'm not all that confident about the fix -- it's quite hard to follow Vim code here -- but it results in behavior matching Vim and all tests still pass.

Note that it depends on #249 due to another bug in this simple scenario.


AlexPl292:
Taking this one as well. Thank you!

PR #251, opened 2020-09-07
VIM-1756: startSel works in insert mode

shaunpatterson:
With ideavimrc settings

set selectmode+=key
set keymodel^=startsel

arrow keys and home/end allow selection in insert mode.  

This appears to be the correct fix to me.  Open to suggestions


AlexPl292:
Hi, thank you for your contribution!

AlexPl292:
Hi @shaunpatterson ! Could you please take a look at this discussion so we can use your change under the new MIT license? Thank you!

PR #253, opened 2020-09-23
Add CJK characterType for delimiting caret movement

yaohui-wyh:
- Add unicode CJK Unified Ideographs type
- Stop caret movement for CJK chars (as VIM does). Similar to [VIM-58](https://youtrack.jetbrains.com/issue/VIM-58)


AlexPl292:
Hi! Nice PR, thank you for the work!

PR #254, opened 2020-11-08
VIM-1475: Add an option to use block caret in insert mode

antekone:
Hi,

This is my first pull request here, and for an Idea plugin in general, and I admit I don't know the API very well. So, please tell me what do you think first, and if something's not right, feel free to point it out or close the PR.

Currently IdeaVim always uses block caret for normal mode, and bar caret for insert mode. This commit introduces a new `.ideavimrc` option named "ideacursor", which can hold an array of "bar" and "block" properties. User can choose which caret style can be used for normal and insert mode by specifying 'bar' or 'caret' in this option:

    set ideacursor=<insert caret style>,<normal caret style>

For example:

    set ideacursor=block,bar

This option will instruct IdeaVim to display block caret in insert mode, and bar caret in normal node. The second argument is optional and will be defaulted to 'block' in case it's missing. If the first argument is missing as well, it will be defaulted to 'bar'. So, not specifying the 'ideacursor' option is equivalent to setting it as:

    set ideacursor=bar,block

(use bar in insert mode, and block in normal mode, this is the standard behavior)

This commit should fix VIM-1475.

Example video: https://www.youtube.com/watch?v=QtcAQvX-a-8

YouTrack link to the issue: https://youtrack.jetbrains.com/issue/VIM-1475

### DIFF
--- a/src/com/maddyhome/idea/vim/KeyHandler.java
+++ b/src/com/maddyhome/idea/vim/KeyHandler.java
@@ -325,7 +325,7 @@ public class KeyHandler {
       }
     }
     reset(editor);
-    ChangeGroup.resetCaret(editor, false);
+    ChangeGroup.resetCaret(editor, VimPlugin.getEditor().isBarCursor());
   }
 
   private boolean handleKeyMapping(final @NotNull Editor editor,
--- a/src/com/maddyhome/idea/vim/group/EditorGroup.java
+++ b/src/com/maddyhome/idea/vim/group/EditorGroup.java
@@ -203,6 +203,10 @@ public class EditorGroup implements PersistentStateComponent<Element> {
     }
   }
 
+  public boolean isBarCursor() {
+    return !isBlockCursor;
+  }
+
   public void editorCreated(@NotNull Editor editor) {
     isBlockCursor = editor.getSettings().isBlockCursor();
     isRefrainFromScrolling = editor.getSettings().isRefrainFromScrolling();
@@ -217,7 +221,7 @@ public class EditorGroup implements PersistentStateComponent<Element> {
       VimPlugin.getChange().insertBeforeCursor(editor, new EditorDataContext(editor, null));
       KeyHandler.getInstance().reset(editor);
     }
-    editor.getSettings().setBlockCursor(!CommandStateHelper.inInsertMode(editor));
+    editor.getSettings().setBlockCursor(!CommandStateHelper.inInsertMode(editor) || isBlockCursor);
     editor.getSettings().setRefrainFromScrolling(REFRAIN_FROM_SCROLLING_VIM_VALUE);
   }
 
--- a/src/com/maddyhome/idea/vim/group/visual/VisualGroup.kt
+++ b/src/com/maddyhome/idea/vim/group/visual/VisualGroup.kt
@@ -163,7 +163,7 @@ fun updateCaretState(editor: Editor) {
 
 fun CommandState.Mode.resetShape(editor: Editor) = when (this) {
   CommandState.Mode.COMMAND, CommandState.Mode.VISUAL, CommandState.Mode.REPLACE -> ChangeGroup.resetCaret(editor, false)
-  CommandState.Mode.SELECT, CommandState.Mode.INSERT -> ChangeGroup.resetCaret(editor, true)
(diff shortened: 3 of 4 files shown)


AlexPl292:
Hi @antekone! Thank you for your interest in the project. I've checked the PR and I think that the overall implementation is good, but I have only one concern about the option itself. Even I like your approach, we're trying to introduce new idea-specific options in rare cases only. I've found that there is `guicursor` (http://vimdoc.sourceforge.net/htmldoc/options.html#'guicursor') options that can be used to define the caret shape for insert or normal modes. However, this option is really powerful, so at the moment we can add support for normal/insert modes and bar/block shapes.
Going through the possibilities of this option, I see that we can possibly support other combinations in the future.
So, what's your opinion about `guicursor` options?

antekone:
From my personal POV, `guicursor` seems like a good place, but one question is that it also can be used to specify the blink rate; should this option be ultimately respected (not necessarily in this PR, but in a future one)? And if yes, will it override global IDEA settings related to cursor blink rate? But also I’m not really sure if IDEA could support the same flexibility as `guicursor` provides, as it can be used to configure nonregular blinks, as well as custom thickness.

Another option I’ve considered is to reuse the “use block caret” setting in IDE. With IdeaVim installed, this option seems to be unused. I’m not sure if there’s a person who would want to have a bar cursor in navigation mode; if we would assume there is no such person, then maybe it would be a good idea to just wire up the “use block caret” option setting to set the IdeaVim’s cursor only in insert mode?

I actually think the second option would maybe be slightly better, because it seems to be intuitive. I’ve chosen another approach only because I wasn’t sure how to read this configuration option from IDE (but I think I do now, though I haven’t verified it).

I think all three approaches (ideacursor, guicursor, “use block caret”) will work, so it’s really a matter of your choice :), I’m fine with all of them, though also I agree that maybe it’s not the best idea to introduce any new configuration option when it’s not really needed, so I’ll still be happy to work on this PR to change the approach to a different one. Even parsing guicursor doesn’t seem to involve that much work, especially if the initial implementation would be limited. So my personal preferences would be something like this:

1. “use block caret” IDE option,
2. guicursor
3. ideacursor

AlexPl292:
Yes, fully agree with your arguments. That's would be cool if we could reuse "use block caret" setting for insert mode. +1 about being more intuitive than the creepy `guicursor`. What have you found about getting the value of this option? My small research gave me `EditorSettingsExternalizable.getInstance().isBlockCursor()`.

As for the `guicursor` option, I don't think that we would be able to support all the options, but we can support some of them. (including thickness). But this is definitely not the goal of the current PR. Regarding the conflicting of `~/.ideavimrc` with IDE settings: at the moment we respect IDE settings unless they are explicitly defined in `~/.ideavimrc`. So, we can use the value of `guicursor` from `~/.ideavimrc` and the value of  "use block caret" if `guicursor` is not present anywhere.

antekone:
OK, I've changed the approach in the PR.

> What have you found about getting the value of this option?

I'm not sure I did it correctly, but I've noticed that in the `editorCreated` method (`EditorGroup.java`), the plugin stores current `getSettings().isBlockCursor()` setting into a field. This field is later used to restore the proper cursor style in case IdeaVIM is disabled in current buffer. So it seems to me that `isBlockCursor` field contains an actual boolean setting cache for the "use block caret" IDE option. So I've used this field when deciding whether IdeaVIM should change the cursor to bar or leave it as block.

The drawback of this is that when changing this option in the IDE settings, a buffer recreation (close + reopen) is needed. But it seems it works this way without IdeaVIM plugin installed as well; in order to change the style of the cursor without IdeaVIM, buffer reopen is needed as well, so it's kind of consistent.

I've changed the title of the PR so that it doesn't mention `ideacursor` anymore, and in case that you will accept it as a PR, before the actual merge I'd like to perform a git squash into 1 commit and force push into this branch, so that the `ideacursor` commit will disappear.

AlexPl292:
> But it seems it works this way without IdeaVIM plugin installed as well; 

Actually, I see that in some cases the caret changes its shape and in other, it doesn't. Not sure what it depends on. I think, that this PR can be merged and I'll check if there is some kind of listener for that.

> git squash into 1 commit and force push into this branch

👍 Looks like the rebase of the branch is also needed because currently there are some conflicts with the master.

antekone:
Rebased to master, squashed and force-pushed. :)

AlexPl292:
Yep, taken into the project. Thank you for your work! :)

PR #255, opened 2020-11-10
Fix various scrolling issues

citizenmatt:
- [x] Fix [VIM-2158](https://youtrack.jetbrains.com/issue/VIM-2158): scrolling jumps around when `scrolloff` is greater than half the screen height, but less than the total screen height
- [x] Reposition the cursor when `scrolloff` changes. E.g. `:set so=999` will reposition the cursor in the middle of the screen immediately, rather than waiting for the next motion
- [x] Limit the height of inlays shown above/below a line when scrolling up/down. Normally, IdeaVim will try to show the associated block inlay (e.g. Rider's Code Vision links for usages, source control, etc.) above a line, when moving that line to the top of the screen, and vice versa - inlays attached below a line when moving that line to the bottom. However, [multiline rendered doc comments](https://blog.jetbrains.com/idea/2020/03/intellij-idea-2020-1-eap8/#in_editor_javadocs_rendering) are treated as block inlays rather than lines, and can take up a lot of space. Trying to show all of the inlay can cause the visible area to bounce around and can even force the cursor off screen. IdeaVim will now limit the block inlay height to 3 standard line heights. If the inlay is taller than this, it will be cut off.
- [x] Workaround an issue with lots of block inlays/doc comments causing the actual number of text lines to be very small. This forces the scrolling algorithm to treat a small movement as a large scroll and so position the cursor in the middle of the screen. This can cause scrolling to "stick", with the cursor unable to scroll to the proper location. A fudge factor has been added to try to catch when there are significantly less text lines than expected, and avoid the scroll to centre loop.
- [x] Tests

Top tip: if you're looking for a good example file to test rendered doc comments, then `String.java` is a good place to go.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -46,6 +46,7 @@ usual beta standards.
 * Add option to disable IdeaVim in dialogs / single line editors. [VIM-765](https://youtrack.jetbrains.com/issue/VIM-765)  
 Use `set ideaenabledbufs=` to disable IdeaVim in dialog editors.  
 _Note for EAP users: the option name can be changed for the stable release_
+* Reposition cursor when `scrolloff` changes
 
 ### Fixes:
 * [VIM-2150](https://youtrack.jetbrains.com/issue/VIM-2150) `Shift-D` should not delete an empty line
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -52,6 +52,7 @@ _Note for EAP users: the option name can be changed for the stable release_
 * [VIM-2157](https://youtrack.jetbrains.com/issue/VIM-2157) Fix tab with an active template
 * [VIM-2156](https://youtrack.jetbrains.com/issue/VIM-2156) Correct up/down motions with inlays
 * [VIM-2144](https://youtrack.jetbrains.com/issue/VIM-2144) Correct text position after block insert with inlays
+* [VIM-2158](https://youtrack.jetbrains.com/issue/VIM-2158) Fix scrolling when `scrolloff` is over half screen height, but less than full height
 
 ## 0.60, 2020-10-09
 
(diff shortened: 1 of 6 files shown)


AlexPl292:
Hey, I've just tried to use this PR with `String` and I'm afraid that the scroll is broken a bit 😞. I don't know, are these issues are the part of this PR, or they are completely independent?

AlexPl292:
Matt, I really appreciate your help. I just see how difficult it is and I'm happy that someone can handle it properly.

<details>
<summary>Offtopic</summary>

Matt Ellis trying to fix all scroll issues

image


</details>

citizenmatt:
Ah. This issue isn't due to the PR - you'll see the same behaviour in a stable version.

The problem is that `so=5` will try to keep five lines of context at the bottom of the screen when moving down with `j`. But the 5th line has a ridiculously large doc comment above it. Scrolling to show the actual 5th line includes this massive inlay, and that pushes the cursor off the top of the screen. And then subsequent `j` movements are trying to catch up.

This PR does have a mitigation for a similar problem - if we were moving up (`k`) instead, or moving down and the 5th line had an inlay attached to the _bottom_ of the line, then IdeaVim makes sure only a max of 3 * line height of the doc comment is shown (try moving the `class String` declaration off the top of the screen and moving up with `k`). But this fix doesn't help in this situation, and it also doesn't help when the line with the massive inlay is line 3 in a 5 line offset.

This needs a separate PR to fix, although I don't know what the fix is yet. First guess is to normalise the scroll offset height to be `expected line height of offset + total acceptable inlay height`, limiting it so that the offset will only include say 3 line heights of inlay. Of course, it's all complicated by the fact that we scroll by positioning a line at the top/bottom of the screen, rather than specifying a pixel offset...

AlexPl292:
Okay, I see. Looks like IdeaVim gets troubles every time IJ devs decide to draw something else than text in the editor :)

PR #257, opened 2020-11-27
Extract SearchHighlightsHelper from SearchGroup

citizenmatt:
I want to refactor state management in `SearchGroup` to allow search state to be per-project, which will give us a chance to correctly handle the [`h` in `'vminfo'`](https://vimhelp.org/options.txt.html#viminfo-h) ([see VIM-2173](https://youtrack.jetbrains.com/issue/VIM-2173)). 

However, `SearchGroup` is a large god class, which makes it hard to reason about. This PR is the first step, extracting most of the search highlight code out of `SearchGroup` and introducing `SearchHighlightsHelper` as a Kotlin helper. It pulls out static methods for adding/removing/updating the search highlights for `hlsearch` and `incsearch`. The methods that require current state are still in `SearchGroup` but defer to `SearchHighlightsHelper`.

It also introduces a cleaner API for setting the search and command registers which does not require a non-null `Editor` that doesn't get used.

A further PR(s) will build on this to continue refactoring and extract state management.


AlexPl292:
Nice work, thank you!

src/com/maddyhome/idea/vim/group/SearchGroup.java
```diff
@@ -115,15 +110,13 @@ public int getLastDir() {
   public void resetState() {
     lastSearch = lastPattern = lastSubstitute = lastReplace = lastOffset = null;
     lastIgnoreSmartCase = false;
-    lastDir = 0;
+    lastDir = Direction.UNSET;
```
AlexPl292, line 113: @citizenmatt Don't you think we can remove UNSET value? It's used only here (I've checked `getLastDir` and "This method is used in AceJump integration plugin", it's okay to remove it). Wouldn't it make usage of this enum more convenient (no need to process UNSET value)?

citizenmatt: Sounds good. I'll remove `UNSET` and `getLastDir` as part of the next round of changes.

AlexPl292: Nono! Don't remove `getLastDir`! Sorry, I've written it incorrectly, I meant that it's okay to remove UNSET because `getLastDir` usage in AceJump integration plugin doesn't use `0` value. It would be safe just to have `1` and `-1`.
But `getLastDir`is still needed :)

citizenmatt: Ha ha! I read that wrong! I'll make sure to keep it. 😁

PR #258, opened 2020-11-30
Show the correct handler class in :map

citizenmatt:
Small fix to show the correct handler class name in the output of `:map`, instead of all handlers showing `call com.maddyhome.idea.vim.key.ToHandlerMappingInfo`


AlexPl292:
Oh, haven't even noticed that. Thank you!

PR #260, opened 2020-12-10
Refactor SearchGroup

citizenmatt:
Continuing the groundwork for [VIM-2173](https://youtrack.jetbrains.com/issue/VIM-2173) that began in #257, this PR refactors `SearchGroup`. There are no code changes, only moving methods around. This is a separate PR to make the next PR easier to review - it will be making code changes without moving things around :smile: 

* Static "find" methods are moved to `SearchHelper`. The methods in `SearchGroup` are now the "high level" APIs to implement Vim's search and replace functionality. They maintain state, parse commands and call into `SearchHelper` to do the actual searching.
* Rearrange and document the APIs in `SearchGroup` to better understand what the methods are doing (there are quite a few overloads simply called `search` that do different things with state and carets, etc.)
* Remove the unnecessary `Direction.UNSET` introduced in #257


AlexPl292:
So much comments. Awesome, thank you!

PR #265, opened 2021-01-09
VIM-1756: Fix escape key when moving from insert mode to select mode

shaunpatterson:
Bug fix for VIM-1756. Original PR: #251 

When moving from Insert mode to Select mode, the escape key would only exit select mode.

Steps to reproduce:

With ideavimrc settings

set selectmode+=key
set keymodel^=startsel

From anywhere in an editor, enter insert mode, select+arrow left, escape. Your cursor will still be in insert mode. This fix escapes Insert mode as well.

Not sure if this is the BEST fix. Open to suggestions.


AlexPl292:
Hi! Sorry for such a big delay. Actually, I'm not sure that the current behaviour of IdeaVim is incorrect. I've checked this in vim and neovim and it looks like they both exit only select mode in this case and leave the editor in insert mode.

PR #275, opened 2021-02-25
Fix edge cases for H, L and M and introduce 'startofline' option

citizenmatt:
This PR fixes a number of edge cases for `H`, `L` and `M` motions:

* Correctly place cursor at midpoint of visible lines, rather than midpoint of screen. Fixes issue when number of visible lines is less than a full screen, such as short files or scrolled up with virtual space. [VIM-2238](https://youtrack.jetbrains.com/issue/VIM-2238)
* Handle op pending mode. e.g. `d100H` will try to delete from current line to 99 lines from top of the screen. In normal mode, this will be capped at the visible screen length, but in op pending, it will use the full `{count}` value and then apply `scrolloff` if set.
* Other minor fixes for `scrolloff`
* Adds a load of tests

It also introduces the [`'startofline'`](https://vimhelp.org/options.txt.html#%27startofline%27) option. It positions the cursor at the first non-blank char when on (default) and tries to keep the same column when disabled. This applies to `H`, `L` and `M`, and also `G`, `gg` and `i_CTRL-Home`, the scrolling motions `<C-D>`, `<C-U>`, `<C-F>` and `<C-B>`, the shift operators `<<` and `>>` and also `{count}%` and `:{range}`.


AlexPl292:
Okay, this great guy goes directly into master. Thank you for updates! (And now I should update my `~/.ideavimrc` with this nice option :)

citizenmatt:
Ah, I've just realised that I didn't update the docs. Not entirely sure how I do that as part of a PR, sorry.

AlexPl292:
You can't. GitHub doesn't support PRs to wiki pages, this is a downside of the current approach :|
Should I update anything?

PR #276, opened 2021-02-26
More refactoring of SearchGroup

citizenmatt:
This PR continues the work started in #260. It refactors `SearchGroup` to provide a more appropriate API that handles user initiated search and substitution. `SearchGroup` maintains (global) state and highlights results. The `SearchHelper` class is intended as a lower level API that does not require state, and is an appropriate API to use from extensions.

This PR will:

- [x] Remove unnecessary APIs from `SearchGroup` and update usages to use `SearchHelper` or other appropriate APIs.
- [x] Remove `CommandFlags.FLAG_SEARCH_FWD` and `FLAG_SEARCH_REV`. The commands simply pass direction into the `SearchGroup` APIs
- [x] Fixes bugs with last used search or substitution patterns. Various bugs with updating the correct pattern at the correct time, and using the correct pattern. Behaviour matches Vim, with added tests
- [x] Fixes to the multiple-cursors extension to use the appropriate search APIs that don't interact with state or highlights. Also fixes a few issues with compatibility with the Vim plugin:
    * Matches no longer update state
    * Search highlighting is not updated for the matched occurrences, and stays at last search/substitute pattern
    * Searches are now always case sensitive, like the plugin
    * Asterisks, etc. in selected text are no longer treated as regex wildcards
    * "Select all occurrences" works correctly with `'nowrapscan'`, so only finds to the end of the file
    * "Select next occurrence" from an existing selection no longer uses word boundary flags
    * Show "no more matches" at correct time
- [x] Update tests


AlexPl292:
Matt, I've rebased your PR on the latest master and found several things:
1) CommandParser.java was transferred to kotlin, so it's now CommandParser.kt, so I've updated this class in the latest commit.
2) One of your previous tests fails with this PR. Exactly because of changes in CommandParser. Could you please check it? `test goto line with scrolloff`

citizenmatt:
Oh wow. Saved by another PR! I did merge latest and reapplied the changes to `CommandParser.kt`, but it was a bad fix even in the Java version! The tests added in #275 found it. At least it means I wasn't guilty of pushing without running tests 😄

PR #289, opened 2021-04-13
VIM-2276 ls returns "absolute path" for files on Microsoft Windows.

MichalPlacek:


AlexPl292:
Hi! Thank you for your fix. File paths are always complicated. 
Unfortunately, I don't have a windows machine at the moment. Does any of existing tests fail under windows without this fix, or we should add a new one? I just want to make sure that this fix is tested.

MichalPlacek:
> Hi! Thank you for your fix. File paths are always complicated.
> Unfortunately, I don't have a windows machine at the moment. Does any of existing tests fail under windows without this fix, or we should add a new one? I just want to make sure that this fix is tested.

All tests are performed correctly (with and without a fix). Tonight, when I have some time, I will add a test for it.

MichalPlacek:
I have a problem to write a test. 
All the tests in BufferListHandlerTest use files in a temporary directory (not belonging to the project). 

For example "testLsActionWithLongFileName" test.
myFixture.project.basePath is "/tmp/unitTest_lsActionWithLongFileName_1rDuizuCEpCGTwpMe9xPeItkeOr"
but file created in this test:  vFile=temp:///src/aaaaaaaaaaaaaaaaaaaaaaaaaaaaa.txt

I need time to find a solution (it's not easy because I don't know this framework)

AlexPl292:
Oh, I see. Yes, this might be a problem.
Okay, in this case could you please check for me again that everything work as expected on windows and I make me a note that I need to do something with that during the next test organization reiteration?

MichalPlacek:
I checked the fix for windows again .It works correctly.
Maybe "organization" knows how to test this case (how to create file in project structure for tests).
I tried to find information about it on the internet but with no success. 
Cutting file path is not possible to tested now.

if you know of any other "simple" bug in ideavim then I can take a look at it.

AlexPl292:
Great, thank you!
We have a list of "patch_welcome" issues, you can take a look at it :)
https://youtrack.jetbrains.com/issues/VIM?u=1&q=tag:%20patch_welcome%20

MichalPlacek:
> Great, thank you!
> We have a list of "patch_welcome" issues, you can take a look at it :)
> https://youtrack.jetbrains.com/issues/VIM?u=1&q=tag:%20patch_welcome%20

I was looking at https://youtrack.jetbrains.com/issue/VIM-1476 during weekend.
I think i'm able to "fix" it (i made a fix but i have to still test it and write test ).

PR #295, opened 2021-04-19
VIM-1476

MichalPlacek:


MichalPlacek:
@AlexPl292 I don't know why TeamCity build failed.  I have not worked with this tool before so I don't know why it didn't work.

AlexPl292:
We have some issues with configuration, sorry for that.

MichalPlacek:
This is ready to merge (the problem is only with "TeamCity build failed")

AlexPl292:
Hi! Sorry for the delay. Actually, it looks like this case is a bit more complicated than expected. Because at the moment `lastKeys` contain a list of "raw" commands, so in case if user types `ab` then presses left arrow and types `c` (acb in the result), we should also process this left arrow for correct `.` register representation. But since in IdeaVim dot register has only representation purpost, I guess this small inconsistency can be ignored at the moment. 
Thank you for your contribution!

PR #300, opened 2021-04-23
Fix scrolling at end of file with folds

citizenmatt:
Fixes [VIM-2291](https://youtrack.jetbrains.com/issue/VIM-2291). If a file has any folds (most Java or Kotlin files) and `scrolloff` > 0 and the caret is within `scrolloff` lines of the bottom of the file, the screen is incorrectly scrolled to put the last line at the bottom of the screen. This is due to mixing up logical lines and visual lines.


AlexPl292:
Nice! Thank you!

PR #308, opened 2021-05-13
Fix unnecessary caret movement when switching tabs

citizenmatt:
Changes `gt` and `gT` actions from `MotionActionHandler` to `VimActionHandler`, as they're not motions. This was causing a problem because IdeaVim would switch tabs then try to move the caret to the current offset, which we don't optimise away because it might result in a different visual position thanks to inlays. Because the editor was no longer visible, the visible area is effectively zero, and repositioning the caret to fit in this area would move it to the top left of the editor. Fixes [VIM-2308](https://youtrack.jetbrains.com/issue/VIM-2308)


test/org/jetbrains/plugins/ideavim/RegisterActionsTest.kt
```diff
@@ -79,7 +79,7 @@ class RegisterActionsTest : VimTestCase() {
     val after = "I f${c}ound it in a legendary land"
     var motionRightAction: ActionBeanClass? = null
     doTest(keys, before, after, CommandState.Mode.COMMAND, CommandState.SubMode.NONE) {
-      motionRightAction = VIM_ACTIONS_EP.extensions().findAny().get()
+      motionRightAction = VIM_ACTIONS_EP.extensions().filter { it.actionId == "VimPreviousTabAction" }.findFirst().get()
```
AlexPl292, line 82: The PR is good, but what is this change about?

citizenmatt: It makes the action we're using explicit, rather than relying on the order of the actions in `VimActions.xml`. The `findAny` function is documented as returning "any" item from the collection, but in practice it was returning the first item, which just happened to be the "previous tab action". I moved that action from the `motion.tabs` package to `windows.tab`, and moved the registration in `VimActions.xml` to match, and that broke the assumption on the order of the actions.

PR #332, opened 2021-06-17
VIM-1357+VIM-1566: Use OS shell to run filter command

citizenmatt:
This PR fixes a few issues with the filter (`:!`) command:

- [x] Use OS shell to execute the given command. On Unix, this is `$SHELL` or `sh`, on Windows, this is `cmd.exe`. Fixes [VIM-1566](https://youtrack.jetbrains.com/issue/VIM-1566) and [VIM-1357](https://youtrack.jetbrains.com/issue/VIM-1357)
- [x] Add `'shellcmdflag'`, `'shellxescape'` and `'shellxquote'` options, required to properly run, quote and escape parameters
- [x] Errors are displayed or substituted into text, like Vim
- [x] Long running tasks can be cancelled
- [x] Adds support for `%` inside a filter command, mapping to the current file name

### DIFF
--- a/resources/dictionaries/ideavim.dic
+++ b/resources/dictionaries/ideavim.dic
@@ -16,6 +16,9 @@ relativenumber
 scrolljump
 scrolloff
 selectmode
+shellcmdflag
+shellxescape
+shellxquote
 showcmd
 showmode
 sidescroll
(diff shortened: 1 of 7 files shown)


AlexPl292:
That's great! Thank you for the fixes!

keithn:
thanks for this PR, really love my shell filters in Vim!

PR #333, opened 2021-06-24
VIM-1062

MichalPlacek:
https://youtrack.jetbrains.com/issue/VIM-1062
added:
:bn select the next editor tab, circling from the last to the first
:bp select the previous editor tab, circling from the first to the last
:bd or bdX

### DIFF
--- a/resources/META-INF/includes/VimExCommands.xml
+++ b/resources/META-INF/includes/VimExCommands.xml
@@ -32,10 +32,10 @@
     <vimExCommand implementation="com.maddyhome.idea.vim.ex.handler.MarkHandler" names="ma[rk],k"/>
     <vimExCommand implementation="com.maddyhome.idea.vim.ex.handler.MarksHandler" names="marks"/>
     <vimExCommand implementation="com.maddyhome.idea.vim.ex.handler.MoveTextHandler" names="m[ove]"/>
-    <vimExCommand implementation="com.maddyhome.idea.vim.ex.handler.NextFileHandler" names="n[ext]"/>
+    <vimExCommand implementation="com.maddyhome.idea.vim.ex.handler.NextFileHandler" names="n[ext],bn"/>
     <vimExCommand implementation="com.maddyhome.idea.vim.ex.handler.NoHLSearchHandler" names="noh[lsearch]"/>
     <vimExCommand implementation="com.maddyhome.idea.vim.ex.handler.OnlyHandler" names="on[ly]"/>
-    <vimExCommand implementation="com.maddyhome.idea.vim.ex.handler.PreviousFileHandler" names="N[ext],prev[ious]"/>
+    <vimExCommand implementation="com.maddyhome.idea.vim.ex.handler.PreviousFileHandler" names="N[ext],prev[ious],bp"/>
     <vimExCommand implementation="com.maddyhome.idea.vim.ex.handler.PromptFindHandler" names="pro[mptfind]"/>
     <vimExCommand implementation="com.maddyhome.idea.vim.ex.handler.PrintHandler" names="p[rint],P[rint]"/>
     <vimExCommand implementation="com.maddyhome.idea.vim.ex.handler.PromptReplaceHandler" names="promptr[epl]"/>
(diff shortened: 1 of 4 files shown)


resources/META-INF/includes/VimExCommands.xml
```diff
@@ -32,10 +32,10 @@
     <vimExCommand implementation="com.maddyhome.idea.vim.ex.handler.MarkHandler" names="ma[rk],k"/>
     <vimExCommand implementation="com.maddyhome.idea.vim.ex.handler.MarksHandler" names="marks"/>
     <vimExCommand implementation="com.maddyhome.idea.vim.ex.handler.MoveTextHandler" names="m[ove]"/>
-    <vimExCommand implementation="com.maddyhome.idea.vim.ex.handler.NextFileHandler" names="n[ext]"/>
+    <vimExCommand implementation="com.maddyhome.idea.vim.ex.handler.NextFileHandler" names="n[ext],bn"/>
```
AlexPl292, line 35: This is not the full command. It should be `bn[ext]`. The same works for other commands.

src/com/maddyhome/idea/vim/ex/handler/BufferCloseHandler.kt
```diff
@@ -0,0 +1,40 @@
...
+import com.maddyhome.idea.vim.VimPlugin
+import com.maddyhome.idea.vim.ex.CommandHandler
+import com.maddyhome.idea.vim.ex.ExCommand
+import com.maddyhome.idea.vim.ex.flags
+
+class BufferCloseHandler : CommandHandler.SingleExecution() {
+  override val argFlags = flags(RangeFlag.RANGE_OPTIONAL, ArgumentFlag.ARGUMENT_OPTIONAL, Access.READ_ONLY)
+  override fun execute(editor: Editor, context: DataContext, cmd: ExCommand): Boolean {
+    val arg = cmd.argument.trim()
+    if (arg.isNotEmpty() && arg.matches(Regex("^\\d+$"))) {
```
AlexPl292, line 32: You can use `toIntOrNull` instead of regex check

src/com/maddyhome/idea/vim/group/FileGroup.java
```diff
@@ -161,6 +160,24 @@
...
 
+  /**
+   * Closes editor.
+   */
+  public void closeFile(int number, @NotNull Editor editor, @NotNull DataContext context) {
+    final Project project = PlatformDataKeys.PROJECT.getData(context);
+    final FileEditorManagerEx fileEditorManager = FileEditorManagerEx.getInstanceEx(project);
+    final EditorWindow window = fileEditorManager.getCurrentWindow();
+    VirtualFile[] editors = fileEditorManager.getOpenFiles();
+    if (number == 99) {
```
AlexPl292, line 171: Why 99 is a special number?

src/com/maddyhome/idea/vim/group/FileGroup.java
```diff
@@ -161,6 +160,24 @@
...
+      number = editors.length - 1;
+    }
+    if (number >= 0 && number < editors.length) {
+      fileEditorManager.closeFile(editors[number], window);
+    }
+    if (number < 0 || number >= editors.length) {
```
AlexPl292, line 177: This if is empty

AlexPl292 (changes requested):
Thank you for your PR! Could you please check some comments?

resources/META-INF/includes/VimExCommands.xml
MichalPlacek: fixed it

src/com/maddyhome/idea/vim/ex/handler/BufferCloseHandler.kt
MichalPlacek: fixed it.

src/com/maddyhome/idea/vim/group/FileGroup.java
MichalPlacek: it was by mistake. i removed it.

src/com/maddyhome/idea/vim/group/FileGroup.java
MichalPlacek: it was by mistake. i removed it.

AlexPl292:
That's great, thank you for your work!

PR #334, opened 2021-06-29
vim-2335 Support for navigation to previous buffer with :b#

MichalPlacek:


AlexPl292:
Great, thank you for updates!

PR #343, opened 2021-07-09
Emulate matchit.vim

myzeiri:
[Addresses VIM-539](https://youtrack.jetbrains.com/issue/VIM-539)

Hello! This is my attempt at emulating [matchit.vim](https://github.com/chrisbra/matchit), a plugin which extends the `%` motion to match whole words and patterns.

My version currently supports jumping with `%` and `g%` on matching HTML/XML tags (including HTML found in JavaScript/TypeScript and React files), and Ruby keywords, including Embedded Ruby `erb` files. It handles comments and strings like the original plugin.

I plan on adding support for more languages in the future. But I think that the HTML jumps alone make this plugin worth adding to IdeaVim. 🙂  

I took care to make sure this doesn't break the default `%` motion, and I've included a lot of unit tests.

**There are a couple of things I'm uncertain about:**

The original plugin uses Vim syntax groups to check for comments, strings, and other language features it wants to ignore. I've done the same but with PSI elements. However, I'm not sure how to check the types of Ruby elements since they appear to be defined outside of IntelliJ. I've resorted to checking the debugNames.

Similarly, I wasn't sure how write unit tests just for Ruby files. I added a new `doTest()` method which takes a filetype parameter -- that makes it easy to deal with `HTMLFileType` or `JavaFileType`. But there's no `RubyFileType`, so I needed another `doTest` that takes a filename parameter. If a better solution for these Ruby features exists then I'd love to hear it.

Thank you for your time, Alex! I look forward to your feedback.

### DIFF
--- a/resources/META-INF/includes/VimExtensions.xml
+++ b/resources/META-INF/includes/VimExtensions.xml
@@ -109,6 +109,13 @@
         <alias name="vim-indent-object"/>
       </aliases>
     </vimExtension>
+
+    <vimExtension implementation="com.maddyhome.idea.vim.extension.matchit.Matchit" name="matchit">
+      <aliases>
+        <alias name="vim-matchit"/>
+        <alias name="chrisbra/matchit"/>
+      </aliases>
+    </vimExtension>
   </extensions>
 
   <!--  IdeaVim extensions-->
--- a/src/com/maddyhome/idea/vim/command/CommandBuilder.kt
+++ b/src/com/maddyhome/idea/vim/command/CommandBuilder.kt
@@ -187,6 +187,10 @@ class CommandBuilder(private var currentCommandPartNode: CommandPartNode<ActionB
     prevExpectedArgumentType = null
   }
 
+  fun resetCount() {
+    count = 0
+  }
+
   fun resetInProgressCommandPart(commandPartNode: CommandPartNode<ActionBeanClass>) {
     count = 0
     setCurrentCommandPartNode(commandPartNode)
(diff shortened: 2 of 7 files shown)


AlexPl292:
Hi! Thank you for your PR! I looked a bit in the code, so far it looks cool. I really like the comments, they simplify the review process.
At the moment, I have to review another PR because it's a bit older, but after that I'll get back to your code to dive into it.

AlexPl292:
Hello, thank you for the great implementation. I think it worth merge this change into the repository.
Regarding the questions:
> However, I'm not sure how to check the types of Ruby elements since they appear to be defined outside of IntelliJ. I've resorted to checking the debugNames.

I think, using the debugNames if they are ruby-specific, is okay. However, you can specify an [optional dependency](https://plugins.jetbrains.com/docs/intellij/plugin-dependencies.html#optional-plugin-dependencies) to ruby plugin (ruby plugin id: `org.jetbrains.plugins.ruby`) to access ruby classes. However, this would complicate the structure because the plugin should work fine without ruby plugin.

> Similarly, I wasn't sure how write unit tests just for Ruby files.

You can add ruby plugin id to `intellij {` section of `build.gradle.kts` in the same way as it made with `java` plugin. In this way this won't be a ruby dependency, but the ruby plugin will be accessible during the tests.

Please keep in mind that I haven't tried these options, this is just how I think this should work.

AlexPl292:
Sure, no problems. However, this specific request about `region` comment is applicable to all languages because it's a "custom folding" feature: https://blog.jetbrains.com/idea/2012/03/custom-code-folding-regions-in-intellij-idea-111/

PR #391, opened 2021-10-24
fix(VIM-2432): handle scrolloff with <C-D> and <C-U>

citizenmatt:
Fixes [VIM-2432](https://youtrack.jetbrains.com/issue/VIM-2432). `<C-U>` and `<C-D>` would not place the caret on the very first or very last line of the file when `'scrolloff'` was set. Also, `<C-D>` would introduce virtual space rather than keep the last line of the file at the bottom of the screen (unless virtual space was already there, in which case it wouldn't scroll at all).


AlexPl292:
Great, thank you for the change!

PR #401, opened 2021-11-09
Reset caret shape when disabling plugin

citizenmatt:
Fixes [VIM-2462](https://youtrack.jetbrains.com/issue/VIM-2462)


AlexPl292:
Oh, I've seen this bug. Thank you for the fix!

PR #418, opened 2021-11-30
Fix compile errors in unit tests on Windows

chylex:
JVM on Windows doesn't use UTF-8, so compilation of unit test sources with special characters fails on Windows:
```
assertIsKeyword('Ā');
// unclosed character literal
// unclosed character literal
// not a statement
```

Discussion: https://github.com/JetBrains/ideavim/discussions/389

Unfortunately I don't know how to fix the `doc` folder containing files with quotes, which are illegal on Windows. In case anyone finds this, all I can recommend is this command to remove the `doc` folder altogether, so that `git status` doesn't constantly complain about missing files:
```
git submodule deinit -f doc
```

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -13,3 +13,6 @@ publishToken=token
 publishChannels=eap
 
 slackUrl=
+
+# Gradle settings
+org.gradle.jvmargs='-Dfile.encoding=UTF-8'


AlexPl292:
Nice!
I wanted to get rid of doc directory as a submodule for some time already, looke like I can do it now.

chylex:
I think removing the submodule but keeping the .md files will make it worse, removing the submodule would effectively hide the illegally named files from git so it wouldn't even try to pull them into Windows :P

Could you please rename these two .md files
- https://github.com/JetBrains/ideavim/blob/master/doc/%22ideajoin%22-examples.md
- https://github.com/JetBrains/ideavim/blob/master/doc/%22set%22-commands.md
so they don't have quotes in the filename?

AlexPl292:
Oops, sorry. Actually github wiki forced me to use the quotes in the name, so now it's fine to remove them. Already done.

PR #421, opened 2021-12-05
Add Matchit support for single file Vue-components, *.vue files

DanEEStar:


AlexPl292:
Hi! I'd say that "JavaScript" item here should cover "vue.js" as well. Doesn't this plugin work for you in `vue.js` files? Does it work in other `.js` files?

DanEEStar:
The plugin works in normal `.js` files. 

This feature adds support for single file [Vue components](https://vuejs.org/v2/guide/components.html) which have its own `.vue`-extension and are a "mixture" of a HTML-Template, JavaScript- and CSS-code. (A bit similar to JSX-files...)
With this added "Vue.js" fileType they work now as expected.

So the title of the Pull-Request is a bit misleading when I think about it. It should say "Add Matchit support for single file Vue-components".

The "hardest" part of the implementation was to find out what the `fileTypeName` is for these Vue-components.
The vue-plugin seems to be a paid feature plugin for IntelliJ.
With some file output writing in the IdeaVim plugin I could find it out by trial and error -> `Vue.js`
Or is there an easier way to "Start IJ with IdeaVim" and then have this bare IntelliJ load paid plugins?

TroAlexis:
Any update on this? Would be really cool to have matchit in vue files, real pity it's not working there 😞 
Good job with the PR, @DanEEStar 
And thanks for your efforts with the plugin @AlexPl292 !

PR #436, opened 2022-01-02
WIP Use ::back() ::forward() from intelij instead of jump list

pmnoxx:
Right now, jump list doesn't work properly. We can switch to using `back()`, `forward()` from `IdeDocumentHistory`.

https://youtrack.jetbrains.com/issue/VIM-44

TODO:
- [ ] Test the feature


pmnoxx:
In order to do the testing. I need to figure out how to get the debug output from `logger`.
```
    if (logger.isDebugEnabled()) {
      logger.debug("width=" + width);
    }
```

PR #440, opened 2022-01-12
Add matchit support for handlebars

ksrb:
Right now there's only support for HTML and JSX it would be nice to add support for Handlebars/Mustache as well.


AlexPl292:
Thank you for your contribution!

PR #441, opened 2022-01-13
Add Matchit support for Vue.js files

DanEEStar:
Rebased the changes from #421 to master.

This feature adds support for single file Vue components which have its own .vue-extension and are a "mixture" of a HTML-Template, JavaScript- and CSS-code. (A bit similar to JSX-files...)
With this added "Vue.js" fileType they work now as expected.

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/extension/matchit/Matchit.kt
+++ b/src/main/java/com/maddyhome/idea/vim/extension/matchit/Matchit.kt
@@ -194,7 +194,7 @@ private object MatchitPatterns {
 
   // Files that can contain HTML or HTML-like content.
   // These are just the file types that have HTML Matchit support enabled by default in their Vim ftplugin files.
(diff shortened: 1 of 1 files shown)


DanEEStar:
> Do you mean that IJ returns a special vue.js value in virtualFile?.fileType?.name for vue.js file? And not JavaScript?

Yes, exactly that! `virtualFile?.fileType?.name` will have the value `"Vue.js"` for `*.vue` vue-component files. 
I checked that with some added logging in a custom build of the IdeaVim plugin.

[Vue Single File Components](https://v3.vuejs.org/guide/single-file-component.html) do also contain HTML-code. With this added `"Vue.js"` value in `htmlLikeFileTypes` matchit also works in them! (At least in my custom build, not in the current IdeaVim version... 😄)

AlexPl292:
Oh, I see now. Okay, great, let's merge it :)

DanEEStar:
👍 Thanks for your great work with IdeaVim!

PR #448, opened 2022-01-20
Support custom element names in vim-surround

cravay:
Hi there!

This PR adds support for custom elements (e.g. `<custom-element></custom-element>`) to the vim-surround extension. Let me know if I should change something.

Thanks a lot for creating this plugin. I like it a lot!


AlexPl292:
Hi! Do I understand it right that is was more like a bug that a tag that contians `-` character wasn't inserted correctly?

cravay:
> Hi! Do I understand it right that is was more like a bug that a tag that contians `-` character wasn't inserted correctly?

Yes, your understanding is correct.

AlexPl292:
Great, thank you for the contribution!

PR #453, opened 2022-01-31
fix(VIM-2470): Fix incorrect reset of cursor shape

citizenmatt:
Fixes [VIM-2470](https://youtrack.jetbrains.com/issue/VIM-2470). Setting the `'guicursor'` option to include an `a:` mode ("all") without specifying a cursor shape would overwrite existing cursor shapes for other modes with "block".

After this change, a missing cursor shape will no longer override. If all modes are missing, then the default block is used.


AlexPl292:
Great, thank you for the fix!
First PR for this release =D

PR #483, opened 2022-03-19
Fix(VIM-2553): Substitute consecutive matches

Vvalter:
We need to set newend after the content of match is final, otherwise we
can end up in an endless loop. This allows us to remove the
startoff != latestOff check that previously prevented this endless
loop and caused this bug.


AlexPl292:
@lippfi It seems that you've added a new condition into this if. Could you please take a look at this PR when you have time?

lippfi:
@Vvalter, thank you for the fix

PR #493, opened 2022-04-18
Improvements to Commentary extension

citizenmatt:
This PR adds a number of improvements and fixes to the Commentary extension:

- [x] Fix [VIM-1687](https://youtrack.jetbrains.com/issue/VIM-1687) - `gcc` does not take a count
- [x] Fix [VIM-2381](https://youtrack.jetbrains.com/issue/VIM-2381) - fall back to line comment if block comment is not available
- [x] Fix [VIM-2589](https://youtrack.jetbrains.com/issue/VIM-2589) - add `gc` text object. Can now use comment blocks as text objects, e.g. delete a comment block with `dgc`, change a comment block with `cgc`, yank with `ygc`, etc. Also allows for `gcgc` to uncomment the current comment block ([VIM-2125](https://youtrack.jetbrains.com/issue/VIM-2125))
- [x] Add `gcu` mapping to uncomment the current comment block (mapped to `<Plug>Commentary<Plug>Commentary` with is the same as `gcgc`)
- [x] Add `:Commentary` command, which works great for commands such as `:%s/fun/Commentary`
- [x] Correctly reposition caret to the range start after adding comments
- [x] Uses the same mappings as Commentary, so custom bindings for the Vim plugin will work with IdeaVim, too. E.g. `<Plug>Commentary` instead of `<Plug>(CommentMotion)`. Old mappings are maintained for compatibility

Some implementation notes:

* The Commentary extension was converted to Kotlin
* The extension no longer uses the `Commenter` services, but calls the `CommentByLineComment` and `CommentByBlockComment` actions directly. This is an important step to get this working with Rider, but it doesn't yet work because Rider's actions are async. The current implementation invokes the actions and then immediately resets selection and caret position. This causes Rider to terminate the async processing of the actions. I will follow up with a separate PR for this ([VIM-1758](https://youtrack.jetbrains.com/issue/VIM-1758), [RIDER-38359](https://youtrack.jetbrains.com/issue/RIDER-38359))
* Simplifying the implementation of the Normal and Visual operators meant introducing Visual mode support for `g@`. THis means the Commentary handler can now set an operator func and invoke `g@` for both Normal and Visual mode.
* Adding a test for [VIM-2381](https://youtrack.jetbrains.com/issue/VIM-2381) requires including YAML as a dependency. All of the file types in the default set of references support both block and line comments. YAML only supports line comments. The YAML plugin is already part of the IdeaIC SDK, but needs to be imported as a reference


src/main/java/com/maddyhome/idea/vim/extension/commentary/CommentaryExtension.kt
```diff
@@ -0,0 +1,199 @@
...
+
+    putKeyMappingIfMissing(MappingMode.NXO, parseKeys("gc"), owner, parseKeys("<Plug>Commentary"), true)
+    putKeyMappingIfMissing(MappingMode.N, parseKeys("gcc"), owner, parseKeys("<Plug>CommentaryLine"), true)
+  }
+
+  private class CommentMotionHandler : VimExtensionHandler {
+    override fun isRepeatable() = true
+
+    override fun execute(editor: Editor, context: DataContext) {
+      val commandState = getInstance(IjVimEditor(editor))
```
AlexPl292: `editor.commandState` can be used here. But this is not important, of course

AlexPl292 (approved):

src/main/java/com/maddyhome/idea/vim/extension/commentary/CommentaryExtension.kt
```diff
@@ -0,0 +1,138 @@
...
+import com.maddyhome.idea.vim.extension.VimExtensionHandler
+import com.maddyhome.idea.vim.helper.StringHelper.parseKeys
+import com.maddyhome.idea.vim.key.OperatorFunction
+import com.maddyhome.idea.vim.newapi.IjVimEditor
+
+class CommentaryExtension : VimExtension {
+  override fun getName() = "commentary"
+
+  override fun init() {
+    putExtensionHandlerMapping(MappingMode.N, parseKeys("<Plug>Commentary"), owner, CommentMotionHandler(), false)
```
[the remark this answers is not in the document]
citizenmatt: Fixed. Added mappings from `<Plug>(CommentMotion)`, etc. to the new `<Plug>…` mappings

src/main/java/com/maddyhome/idea/vim/extension/commentary/CommentaryExtension.kt
citizenmatt: Done

citizenmatt:
Added a couple of new commits. This adds a backwards compatibility mapping so anyone using `<Plug>(CommentMotion)` or any of the other older mappings will still work.

Also fixes [VIM-2381](https://youtrack.jetbrains.com/issue/VIM-2381), by adding a fallback. If the block comment action isn't available, then we fall back to the line comment (and vice versa).

AlexPl292:
Yes, this looks awesome! Thank you for the updates.

PR #504, opened 2022-05-11
Minor bug fixes

citizenmatt:
This PR fixes a couple of tiny bugs:

- [x] [VIM-2622](https://youtrack.jetbrains.com/issue/VIM-2622) If `'shell'` contains `powershell` on Windows, then `'shellcmdflags'` will default to `-Command`
- [x] Fix incorrect animation when scrolling one line, e.g. with `j`. If the caret was inside `'sidescrolloff'` columns of the left of the screen, then IdeaVim would try to put the caret column in the middle of the screen, asking the scrolling system to scroll to a negative horizontal offset. This was capped at 0, but the animation system would include the negative value in its calculation of distance scrolled. This puts the distance over a threshold and the scroll is animated. If the caret is in any column greater than `'sidescrolloff'` then it is not animated.


AlexPl292:
Merged, Thank you!

PR #519, opened 2022-06-18
Fix(VIM-2227): Wrong behavior when deleting / changing surround with invalid character

chylex:


AlexPl292:
This looks correct. Thank you for the contribution!

PR #525, opened 2022-07-05
Improve handling of fractional width fonts

citizenmatt:
The default font for the `corretto-11` JDK (at least on my Mac) is Menlo 12 point. This results in nice 8 pixel character widths and integer maths. If `jbr-11` is used, the font is JetBrains Mono 13 point, which has characters that are 7.8 pixels wide (I don't know why a larger font has a smaller pixel side, but that's not really the problem here). 

Most APIs to work with editor positions uses integers, which can cause rounding errors. I don't think this caused any obvious issues in production - possibly some minor off-by-one (or two) errors in column positioning, but it does cause issues with tests, where we need reproducibility.

This PR updates `EditorHelper` to overcome rounding errors in scrolling, and updates tests to correctly calculate expected offsets. Tests now run on both `corretto-11` and `jbr-11`.

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/helper/EditorHelper.java
+++ b/src/main/java/com/maddyhome/idea/vim/helper/EditorHelper.java
@@ -37,6 +37,7 @@ import org.jetbrains.annotations.Nullable;
 import org.jetbrains.annotations.Range;
 
 import java.awt.*;
+import java.awt.geom.Point2D;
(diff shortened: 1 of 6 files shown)


AlexPl292:
Fonts:
Me previously: It's about how fancy you can draw a letter.
Me now: MATH

Thank you for the fix :)

PR #528, opened 2022-07-20
Implement partial code completion support in macros

chylex:
Works ok with insertions (Enter, Ctrl+Enter) but not with replacements (Tab). It also breaks if the code completion modifies imports.

I did this a while back to improve the situation with https://youtrack.jetbrains.com/issue/VIM-913. I hoped there would be a better solution that would cover all possible cases including replacements and imports, but this half-fix is better than nothing :P


AlexPl292:
Yes, it seems okas as a start. Thank you!

PR #531, opened 2022-08-02
Consolidate doTest methods

citizenmatt:
Updates `VimTestCase.kt` to consolidate the various `doTest` functions so everything eventually goes through one `doTest` function. `@JvmOverloads` are added for Java tests, and default arguments means it is no longer necessary to specify the default modes of `COMMAND`/`NONE` at all call sites (not removed in existing tests). Also fixes a minor bug in one of the deleted `doTest` overloads that would pass keys to neovim twice (was fortunately never called).

This is groundwork for extending `doTest` to support testing wrapped and folded text without having to use offsets.


AlexPl292:
Yay, thank you for cleanup :D

PR #550, opened 2022-10-28
Fix(VIM-2778) Remove override of editor scroll setting 

citizenmatt:
Previously, IdeaVim would enable the "refrain from scrolling" setting in each editor. This appears to be historical, as all IdeaVim scrolling is explicit and to an exact location. This setting now only affects the scroll location when IntelliJ navigates, such as with Go To Member when the target is already visible.

The default setting is "keep the caret in place, scroll editor canvas" (refrain from scrolling disabled - hooray for double negatives!), which will scroll the editor and place the caret about 1/3rd of the way down the editor. IdeaVim would previously override this to "move caret, minimise editor scrolling", which will move the caret without scrolling if the target is already visible.

Even though this overridden behaviour is more Vim-like (Vim prefers to move the caret if the target is visible, and scroll to centre otherwise - see `gv`), this setting does not affect IdeaVim navigation, and only applies to IntelliJ navigation.

This fixes [VIM-2778](https://youtrack.jetbrains.com/issue/VIM-2778), so the Find Usages preview window will now correctly use the setting.

This PR also replaces non-exact scrolling (`scrollTo(CENTER)`) in the `gv` handlers. The default `scrollCaretIntoView` behaviour is more Vim-like, and will scroll only if necessary. This removes some unnecessary code from `VimEditor`.

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/group/EditorGroup.java
+++ b/src/main/java/com/maddyhome/idea/vim/group/EditorGroup.java
@@ -59,10 +59,8 @@ import static com.maddyhome.idea.vim.helper.CaretVisualAttributesHelperKt.update
  */
 @State(name = "VimEditorSettings", storages = {@Storage(value = "$APP_CONFIG$/vim_settings.xml")})
 public class EditorGroup implements PersistentStateComponent<Element>, VimEditorGroup {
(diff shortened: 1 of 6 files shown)


AlexPl292:
Awesome, thank you!

PR #553, opened 2022-11-14
Rearrange and rename some code in engine

citizenmatt:
This PR is mostly refactoring, with very little change in logic. The changes are related to a forthcoming PR with fixes for soft wrap support, and is pulled into a separate PR to try to reduce the noise for the next PR.

The main changes in this PR:

- [x] Rearrange `VimMotionGroup` and `VimCaret` to group related functions, to make it easier to see what functions are available
- [x] Rename various "screen" related functions in `VimMotionGroup` to "display". This is to try to start using more Vim terminology and less IntelliJ terminology in the engine API. In Vim, a "screen" is a collection of all visible windows, as well as the command line and status bar. It would be more correct to use "window", but to avoid confusion with GUI windows, I've chosen to use "display", which the Vim help uses in places to mean "window" lines.
- [x] Rename various line motion functions `moveCaretTo…` to clarify if it moves to buffer line (e.g. `100G`), moves relative to current line (`2j`) or moves the caret to the start/end of the current line

These changes only affect the engine API in `VimMotionGroup`, and don't (yet) rename actions or implementations in the editor helper classes.

The forthcoming changes for soft wrap make it more important to use the correct terminology, because we currently work in a mix of IntelliJ logical lines, which are equivalent to Vim's "buffer" lines, and visual lines, which don't really have a Vim equivalent. E.g. motions such as `j` and `k` work in visual lines, but should be in "Vim logical" lines, which are buffer lines with folding applied. Scrolling is also currently in visual lines, but should also be in "Vim logical" lines. We should be doing much more in terms of "Vim logical" lines.

(Vim does have "display" or "window" lines, which map to the static window character grid and don't change as the text content is scrolled, modified or wrapped, unlike IntelliJ's visual lines. Vim has also recently introduced a `'smoothscroll'` option that will scroll by display line when using `<C-E>` and `<C-Y>`, but it's off by default. And to make things even more interesting, Vim only allows folding complete lines, but IntelliJ can fold arbitrary lines, like in a Java getter or setter method, which means a "Vim logical" line should map to a range of start/end buffer lines 😁)


vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimMotionGroup.kt
```diff
@@ -15,116 +15,117 @@ import com.maddyhome.idea.vim.common.TextRange
...
+  // TODO: These should move to the Vim logical line at the top/bottom/middle of the display
+  fun moveCaretToFirstScreenLine(editor: VimEditor, caret: VimCaret, count: Int, normalizeToScreen: Boolean): Int
+  fun moveCaretToLastScreenLine(editor: VimEditor, caret: VimCaret, count: Int, normalizeToScreen: Boolean): Int
+  fun moveCaretToMiddleScreenLine(editor: VimEditor, caret: VimCaret): Int
+
+  // Move caret to buffer column
+  fun moveCaretToColumn(editor: VimEditor, caret: VimCaret, count: Int, allowEnd: Boolean): Motion
+
+  // Move caret to column relative to the bounds of the display (aka window)
+  fun moveCaretToCurrentDisplayLineStart(editor: VimEditor, caret: VimCaret): Motion
```
AlexPl292, line 60: Just sharing plans: I also would like to get rid of the "move" inside such method because it looks like this method does some modification while it's not

src/main/java/com/maddyhome/idea/vim/newapi/IjVimCaret.kt
```diff
@@ -93,10 +92,6 @@ class IjVimCaret(val caret: Caret) : VimCaretBase() {
     this.caret.moveToLogicalPosition(LogicalPosition(logicalPosition.line, logicalPosition.column, logicalPosition.leansForward))
   }
 
-  override fun offsetForLineStartSkipLeading(line: Int): Int {
```
AlexPl292, line 96: Removing methods from our Vim*** classes makes me so happy

AlexPl292 (approved):

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimMotionGroup.kt
citizenmatt: I haven't been able to come up with a good name for this though - `calculateMotionToCurrentDisplayLineStart` or `motionToCurrentDisplayLineStart`? Or even more simple, because it's always called through `motion` - `injector.motion.toCurrentDisplayLineStart`?

citizenmatt:
Be careful what you wish for 😁

PR #554, opened 2022-11-15
Refactor "last column" calculations

citizenmatt:
This PR refactors how IdeaVim handles tracking the "last column" value. This value is the intended location of the caret, when that value might be adjusted before actually moving the caret. For example, if the caret is on a long line and moved to a shorter line, the original column value is stored as the "last column" (intended) value, so that the original column can be used when moving back to a longer line. Similarly, it can be set to the `LAST_COLUMN` magic value by the `$` command to always put the caret at the end of the line, regardless of length.

This PR:
- [x] Introduces a `Motion.AdjustedOffset` that can be returned from motion action handlers and helpers. This allows the handlers and helpers to calculate the adjustment, which will be important for soft wrap support. The base motion action handler understands this new type and will move the caret and update the saved value
- [x] Removes the `preOffsetComputation`, `preMove` and `postMove` hooks in the action handlers. These were mostly used to set and reset the `LAST_COLUMN` magic value. It is simpler to return these as part of the motion, which reduces the places that set the "last column" value
- [x] Simplify invalidating and resetting the value. Moving the caret invalidates the set value (handling any changes external to IdeaVim) and the value is only set if it is different to the current location

Only handlers that use `Motion` (i.e. motion action handlers) get this automatic setting of "last column". Other handlers such as operators still have to set the value manually, as before. However, they need to be careful to reset the value if the caret is moved as part of the operation (e.g. delete will move the caret, so the value needs to be stored in a variable before deleting, and reset before moving).

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/group/ChangeGroup.java
+++ b/src/main/java/com/maddyhome/idea/vim/group/ChangeGroup.java
@@ -43,6 +43,7 @@ import com.maddyhome.idea.vim.common.TextRange;
 import com.maddyhome.idea.vim.ex.ranges.LineRange;
 import com.maddyhome.idea.vim.group.visual.VimSelection;
 import com.maddyhome.idea.vim.group.visual.VisualModeHelperKt;
+import com.maddyhome.idea.vim.handler.Motion;
 import com.maddyhome.idea.vim.helper.*;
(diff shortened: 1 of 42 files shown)


citizenmatt:
Getting closer! Maybe one more refactor… 😁

PR #555, opened 2022-11-17
Rename VimLogicalPosition

citizenmatt:
This PR renames `VimLogicalPosition` to `BufferPosition` and updates related function, parameter and variable names in the engine API. This is to avoid confusion between Vim's "logical lines", which are required for soft wrap support, and IntelliJ's `LogicalPosition`, which is an IntelliJ concept to represent a location in the text buffer.

Vim understands several line types, including "buffer lines" and "logical lines" (see [`:help definitions`](https://vimhelp.org/intro.txt.html#definitions)). Buffer lines are the lines of the text buffer, as stored on disk, and map exactly to IntelliJ's `LogicalPosition`. Vim logical lines are buffer lines with folds applied. A folded range is represented by a single Vim logical line, as is a soft wrapped line. 

Vim motions should be in terms of Vim logical lines. E.g. `j` and `k` should move down or up by one Vim logical line, which is important for wraps and folds. The current implementation moves in terms of IntelliJ visual lines, which works for folds, but not soft wraps. To make wraps work, it will be necessary to introduce a "Vim logical line" type, but naming could be confusing with both `VimLogicalPosition` and `VimLogicalLine`. It makes sense to use Vim's concepts rather than IntelliJ's in the engine API, so we'll use Vim's terminology too.

This PR also tries to make buffer lines the default line type in the API. Unless otherwise stated in the function or parameter name, any API with a name that contains only "line" is a buffer line. Any other line type or a scenario which is ambiguous will use an explicit name, e.g. `visualLine`. (Visual lines are also IntelliJ concepts that don't have Vim equivalent. It should be a goal to remove them from the engine API, and swapping visual lines for Vim logical lines in motions will help with this)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -270,7 +270,7 @@ tasks {
         dependsOn("generateGrammarSource")
     }
     named("compileTestKotlin") {
-        dependsOn("generateGrammarSource")
+        dependsOn("generateTestGrammarSource")
     }
 
     // Add plugin open API sources to the plugin ZIP
--- a/src/main/java/com/maddyhome/idea/vim/action/internal/AddInlineInlaysAction.kt
+++ b/src/main/java/com/maddyhome/idea/vim/action/internal/AddInlineInlaysAction.kt
@@ -16,7 +16,7 @@ import com.intellij.openapi.actionSystem.DataContext
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.editor.VisualPosition
 import com.maddyhome.idea.vim.api.lineLength
-import com.maddyhome.idea.vim.api.visualLineToLogicalLine
+import com.maddyhome.idea.vim.api.visualLineToBufferLine
 import com.maddyhome.idea.vim.helper.VimNlsSafe
 import com.maddyhome.idea.vim.newapi.vim
 import java.util.*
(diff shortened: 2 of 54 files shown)


AlexPl292:
Why do you think this might be "controversial"? I don't see any problems with this one :)

Do you believe that we can merge IJ folding with vim folding concepts? I mean, the fact that IJ may have only a part of the line that is folded, while in vim you always have the whole line folded. Or this is not a part of your investigation?

AlexPl292 (approved):

citizenmatt:
Possibly controversial because I'm making unilateral sweeping changes to the engine and asking forgiveness, not permission :grin:

I think folds will be fine. We already work _pretty_ well with them, although there are some issues. When I've been making changes for soft wraps, I've been trying to also make it work with folds, but there are other places that I'm not touching and will need fixing in the future. The main change is to remember that everything should be done in terms of "Vim logical lines" and thanks to IntelliJ's folding model, one Vim logical line might span multiple buffer lines (e.g. Java getter and setters). My `VimLogicalLine` is actually a range of start/end buffer line, and that seems to work nicely, but currently all APIs are in terms of buffer lines. As long as we use these sensibly, then it's ok - if we need to get the end offset of the current line, we should get the end offset of the end buffer line of the current Vim logical line. But I'm concerned that it's too easy to forget this and just get the current caret line and ignore this. It might be better to change the APIs to expect `VimLogicalLine` so it can handle that logic for you, I'll decide that later, once everything's working 😄

PR #560, opened 2022-11-26
Fix(VIM-2577) paste not working at end of notebook cell

Runinho:
Improves the situation at [VIM-2577](https://youtrack.jetbrains.com/issue/VIM-2577/PyCharm-2021.3-EAP-IdeaVim-1.10.0-Cannot-paste-at-end-of-cell) with the same "fix" that is done at the end of the file.

Technically that behavior is not the same as in vim. Pasting at the end of the file should not result in an empty line bellow the inserted text.

### DIFF
--- a/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/put/VimPutBase.kt
+++ b/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/put/VimPutBase.kt
@@ -25,6 +25,7 @@ import com.maddyhome.idea.vim.command.VimStateMachine
 import com.maddyhome.idea.vim.command.isBlock
 import com.maddyhome.idea.vim.command.isChar
 import com.maddyhome.idea.vim.command.isLine
+import com.maddyhome.idea.vim.common.Offset
 import com.maddyhome.idea.vim.common.TextRange
 import com.maddyhome.idea.vim.common.offset
 import com.maddyhome.idea.vim.diagnostic.vimLogger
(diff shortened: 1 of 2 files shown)


Runinho:
Sorry, I messed up the issue numbers. Code and commit should mention the right one now.

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/put/VimPutBase.kt
```diff
@@ -433,6 +434,11 @@ abstract class VimPutBase : VimPut {
         SelectionType.LINE_WISE -> {
           startOffset =
             min(vimEditor.text().length, injector.motion.moveCaretToLineEnd(vimEditor, line, true) + 1)
+          // At the end of a notebook cell the next symbol is a guard,
+          // so we add a newline to be able to paste. Fixes VIM-2577
+          if (vimEditor.document.getOffsetGuard(Offset(startOffset))!= null) {
+            application.runWriteAction { (vimEditor as MutableVimEditor).insertText((startOffset-1).offset, "\n") }
```
AlexPl292, line 440: Do we need a `startOffset > 0` check here?

AlexPl292 (approved):
Hi there! Thank you for looking into this!
This looks fine for me generally, but I believe it would be needed to add at lease one test to be sure that everything works as expected.

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/put/VimPutBase.kt
Runinho: startOffset is only 0 if vimEditor.text().length is 0.
We only get in trouble if there is also a guard at that 0 position.
I think this case should not exist? But let's be safe instead of sorry.

Runinho:
@AlexPl292 I added a test case.
I think it would be better if we could do the test in a real notebook editor instance.
But these are not available during test time right?

My workaround is creating a guard similar to the ones created by the notebook editor.

AlexPl292:
> But these are not available during test time right?

Strictly speaking, it's possible, but it'll be needed to change the SDK dependency to PyCharm and setup all of that stuff, so it's way easier to do it in the way you did.

Thank you for the fix!

PR #575, opened 2022-12-20
Multicaret marks

AlexPl292:


vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimMarkServiceBase.kt
```diff
@@ -0,0 +1,605 @@
...
+      else -> false
+    }
+  }
+
+  override fun setMark(caret: ImmutableVimCaret, mark: Mark): Boolean {
+    val markChar = mark.key
+
+    if (!markChar.isValid(VimMarkService.Operation.SET, caret)) return false
+
+    if (markChar.isGlobalMark()) {
```
AlexPl292, line 179: What about replacing this huge if with `when`?

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimMarkServiceBase.kt
```diff
@@ -0,0 +1,605 @@
...
+    val selectionType = SelectionType.fromSubMode(editor.subMode)
+    editor.carets()
+      .forEach {
+        val start = editor.offsetToBufferPosition(it.vimSelectionStart)
+        val end = editor.offsetToBufferPosition(it.offset.point)
+        it.lastSelectionInfo = SelectionInfo(start, end, selectionType)
+      }
+  }
+
+  override fun setVisualSelectionMarks(caret: ImmutableVimCaret, range: TextRange) {
```
AlexPl292, line 245: Please move all such "helper" method to the extension function on the interface. This will make an implementation a way cleaner.

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimMarkServiceBase.kt
```diff
@@ -0,0 +1,605 @@
...
+    return getMarksRange(caret, CHANGE_START_MARK, CHANGE_END_MARK)
+  }
+
+  private fun getMarksRange(caret: ImmutableVimCaret, startMark: Char, endMark: Char): TextRange? {
+    val editor = caret.editor
+    val startOffset = getLocalMark(caret, startMark)
+      ?.offset(editor) ?: return null
+    val endOffset = getLocalMark(caret, endMark)
+      ?.offset(editor) ?: return null
+    return TextRange(startOffset, endOffset + 1)
```
AlexPl292, line 269: Are you sure about this `+1` magic constant? Do we always keep the end offset as exclusive? Do we always need it inclusive? Does `endOffset + 1` always < file size? Is it possible that `startOffset > endOffset`?

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimMarkServiceBase.kt
```diff
@@ -0,0 +1,605 @@
...
+      } else if (markChar == SELECTION_END_MARK) {
+        val selectionInfo = caret.lastSelectionInfo
+        val endPosition = selectionInfo.end
+        if (endPosition != null) caret.lastSelectionInfo = SelectionInfo(selectionInfo.start, null, selectionInfo.type)
+      } else {
+        val path = caret.editor.getPath() ?: return
+        getLocalMarks(path).remove(markChar)
+      }
+    } else {
+      caret.markStorage.removeMark(markChar, caret)
```
AlexPl292, line 301: Shouldn't we have the logic above about the SELECTION_START_MARK also inside of markStorage.removeMark?

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimMarkServiceBase.kt
```diff
@@ -0,0 +1,605 @@
...
+  override fun updateMarksFromInsert(editor: VimEditor, insertStartOffset: Int, insertLength: Int) {
+    val caretToMarks = getAllMarksForFile(editor)
+    if (caretToMarks.isEmpty()) return
+
+    val insertEndOffset = insertStartOffset + insertLength
+    val insStart = editor.offsetToBufferPosition(insertStartOffset)
+    val insEnd = editor.offsetToBufferPosition(insertEndOffset)
+    logger.debug { "mark insert. insStart = $insertStartOffset, insEnd = $insertEndOffset" }
+    val lines = insEnd.line - insStart.line
+    if (lines == 0) return
```
AlexPl292, line 321: Why don't we update marks if the amount of lines was not changed? What about the colum-dependent marks?

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimMarkServiceBase.kt
```diff
@@ -0,0 +1,605 @@
...
+    for (editor in injector.editorGroup.getAllEditors()) {
+      editor.carets().forEach {
+        resetAllMarksForCaret(it)
+      }
+    }
+    filepathToLocalMarks.clear()
+    globalMarks.clear()
+  }
+
+  override fun isValidMark(char: Char, operation: VimMarkService.Operation, isCaretPrimary: Boolean): Boolean {
```
AlexPl292, line 423: Name is unclear, probably `isOperationValidOnMark`?

src/main/java/com/maddyhome/idea/vim/group/EditorGroup.java
```diff
@@ -257,6 +255,14 @@ public void notifyIdeaJoin(@NotNull VimEditor editor) {
     notifyIdeaJoin(((IjVimEditor) editor).getEditor().getProject());
   }
 
+  @NotNull
+  @Override
+  public Collection<VimEditor> getAllEditors() {
```
AlexPl292, line 260: Why don't you use `localEditors`? I suppose the method you've created may return remote editors in case of CWM.

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/mark/VimMarkGroup.kt
```diff
@@ -10,7 +10,10 @@ package com.maddyhome.idea.vim.mark
 
 import com.maddyhome.idea.vim.api.VimEditor
 import com.maddyhome.idea.vim.common.TextRange
+import org.jetbrains.annotations.ApiStatus
 
+@Deprecated("Please use VimMarkService instead")
+@ApiStatus.ScheduledForRemoval(inVersion = "2.3")
 interface VimMarkGroup {
```
AlexPl292, line 17: You can execute `Find external usages` action on every of the method. If there are no local and no external usages, the method is safe to delete.

src/main/java/com/maddyhome/idea/vim/group/MotionGroup.java
```diff
@@ -608,47 +608,41 @@ public boolean scrollLines(@NotNull VimEditor editor, int lines) {
...
+      for (Caret carett : markEditor.getCaretModel().getAllCarets()) {
+        new IjVimCaret(carett).moveToOffset(toLineStart
+                                            ? moveCaretToLineStartSkipLeading(new IjVimEditor(markEditor), line)
+                                            : markEditor.logicalPositionToOffset(
+                                              new LogicalPosition(line, mark.getCol())));
       }
     }
-    return -2;
+
+    return -1;
```
AlexPl292, line 639: Why it's `-1` now?

src/main/java/com/maddyhome/idea/vim/group/MotionGroup.java
```diff
@@ -608,47 +608,41 @@ public boolean scrollLines(@NotNull VimEditor editor, int lines) {
...
-      for (Caret caret : selectedEditor.getCaretModel().getAllCarets()) {
-        new IjVimCaret(caret).moveToOffset(toLineStart
-                                           ? moveCaretToLineStartSkipLeading(new IjVimEditor(selectedEditor), line)
-                                           : selectedEditor.logicalPositionToOffset(
-                                             new LogicalPosition(line, mark.getCol())));
+    final Project project = ((IjVimEditor) caretEditor).getEditor().getProject();
+    final Editor markEditor = selectEditor(project, mark);
+    if (markEditor != null) {
+      // todo should we move all the carets or only one?
+      for (Caret carett : markEditor.getCaretModel().getAllCarets()) {
```
AlexPl292, line 631: `carett`

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/mark/Marks.kt
```diff
@@ -14,14 +14,11 @@ import org.jetbrains.annotations.NonNls
 
 interface Mark {
   val key: Char
-  val line: Int
-  val col: Int
-  val filename: String
+  val line: Int // 0-based
```
AlexPl292, line 17: Respect for comment

PR #579, opened 2022-12-28
VIM-2799: Add Matchit support for cshtml files

myzeiri:
This enables HTML jumps in [Razor templates][razor] to fix [VIM-2799][issue]. I confirmed this works in the latest version of Rider.

[issue]: https://youtrack.jetbrains.com/issue/VIM-2799/matchit-support-for-cshtml-files

[razor]: https://learn.microsoft.com/en-us/aspnet/core/mvc/views/razor?view=aspnetcore-7.0


AlexPl292:
Looks good for me, thank you!

PR #583, opened 2023-01-05
Introduce ScrollGroup

citizenmatt:
This PR extracts the scrolling functionality from `MotionGroup` and `VimMotionGroup` and introduces the `VimScrollGroup` interface and the `ScrollGroup` service. The methods on `VimMotionGroup` were checked for external usages before moving. The code has also been converted into Kotlin.

Furthermore, the `scrollCaretIntoView` functions in the `ScrollViewHelper.kt` file are now licensed under the Vim license. These functions control how the window is scrolled to place the caret at least `'scrolloff'` lines away from the top/bottom of the window, or place the caret to the middle of the window. The current implementation is based on reading Vim source, but is a different implementation that does not support soft wrapped lines. Scrolling is a complex area in Vim, and to accurately mimic Vim's behaviour with soft wraps, this implementation will need to be rewritten to match Vim's implementation.

PS. I migrated `MotionGroup` to Kotlin too, because who doesn't like a bit of Kotlin?


src/main/java/com/maddyhome/idea/vim/newapi/IjVimCaret.kt
```diff
@@ -145,10 +145,10 @@ class IjVimCaret(val caret: Caret) : VimCaretBase() {
   override fun hashCode(): Int = this.caret.hashCode()
 }
 
-val VimCaret.ij: Caret
+inline val VimCaret.ij: Caret
```
AlexPl292, line 148: Just curious, any advantages from making this field inlined?

AlexPl292:
Awesome as always. Thank you!

src/main/java/com/maddyhome/idea/vim/newapi/IjVimCaret.kt
citizenmatt: Possibly not :) I didn't measure anything, but it felt like an overhead to call a function that simply does a cast and property access, especially when we've got that same pattern repeated in the code. It just means there's no difference to putting the pattern in the calling code or using this helper function, which is a lot nicer.

I've always wondered if it's worth putting `IjVimEditor` into `Editor`s user data, to prevent allocating too frequently, too.

PR #586, opened 2023-01-09
Fixed number references in Tutor

patrick-elmquist:
Fixed some minor reference issues in the Tutor


AlexPl292:
Nice catch! Thank you.

PR #589, opened 2023-01-12
Refactor option usage and introduce accessor API

citizenmatt:
First things first - sorry about the size of this PR. It's a bit of a monster, but I don't think I can make it any smaller :blush:. It would be possible to mostly split the PR into production code changes and test code changes, but there are some renames that affect all files, so this wouldn't reduce the number of files to review, only the number of commits. Fortunately, the changes are mostly mechanical refactorings, with no significant changes to business logic.

This PR is primarily about refactoring and minimising usages of `OptionService`, so that a subsequent PR can refactor that service with less impact, which will make that PR easier to review. These subsequent changes are to support `:set wrap` - an option that needs to be backed by IDE APIs rather than a dictionary value, as there doesn't appear to be an event listener for editor settings to sync the dictionary value.

Specifically, this PR will:

* Introduce a friendly API to consume options values in code, e.g. `getIntValue` to work with `'scrolloff'`. This `OptionValueAccessor` is accessed via `injector.options(editor)` or `injector.globalOptions()`, and decouples the majority of production code from `OptionService`. It also simplifies consuming options. Code that needs to administer options, such as in response to `:set something+=whatever` will still use `OptionService`.
* Introduce the same accessor into tests via the helper functions `VimTestCase.options()` and `VimTestCase.optionsNoEditor()` (for accessing global options). This again simplifies the calling code, and reduces coupling to `OptionService`.
* Replace test usage of `OptionService.setOptionValue`, `setOption`, `unsetOption`, etc. by using the `:set` command. Tests now set options in the same way that users do - `:set scrolloff=10`, `set hlsearch ignorecase`, etc.
* Some other small refactorings


vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimInjector.kt
```diff
@@ -50,7 +50,26 @@ interface VimInjector {
...
-  // TODO We should somehow state that [OptionServiceImpl] can be used from any implementation
+  /**
+   * Gets an API for consuming only global options
+   *
+   * This function is intended to retrieve global options, not global values for options that are local to buffer, local
+   * to window or global-local - for that, use [OptionService.getOptionValue]. Typical option access should use
+   * [VimInjector.options] and pass in a [VimEditor] for context. This will return local or global values as
+   * appropriate.
+   */
+  fun globalOptions() = optionService.getValueAccessor(null)
```
AlexPl292, line 61: I'd prefer to have these methods as extensions on VimInjector. In this way they won't be a part of the interface and there is no risk that some implementation will override it.

AlexPl292:
Well, this looks for me :)

Also, I'd like @lippfi to take a look as he worked on this OptionService a lot

AlexPl292 (approved):

AlexPl292:
Also, just a fancy idea: what if we'll change our current option listeners to coroutine API with flow and stuff. Not sure if this will bring any actual value, but this may be interesting plus, it's a movement into the common direction "use kt coroutines".

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimInjector.kt
citizenmatt: I did do that originally, but moved it to an interface method to keep Java consumers simple. This felt too clumsy to write, when the aim of adding the accessor was to simplify the API:

```
final int scrolloff = VimInjectorKt.options(injector, editor).getIntValue(OptionConstants.scrolloff);
```

But we could use static import for the function:

```
final int scrolloff = options(injector, editor).getIntValue(OptionConstants.scrolloff);
```

What do you think?

AlexPl292: Static imports seem fine to me. Also, taking the fact that we slowly move the code to kt, this inconvenience should be eliminated.

AlexPl292: We have actually like 10% of java in the sources. We can make a move and convert everything to kt

citizenmatt: I've just pushed a new commit that replaces the default interface methods with extension methods.

citizenmatt:
I think we should definitely look at coroutines for other listeners, but I don't really think there's a need for option change listeners. If anything, I'd be happy to get rid of them! They look like an API that can be used to get notification that options have changed, but they're really an implementation detail to refresh the UI when the option changes. I'd be tempted to move them to the `Option` class itself.

they're a good fit for option change listeners. If anything, I'd like to get rid of them if possible! They look like an API, but they're really an implementation detail, for refreshing behaviour in response to `:set`

lippfi (approved):

PR #603, opened 2023-02-25
VIM-2773: Prevent viewport displacement on undo

dominicpalmer:
Small bug fix for [VIM-2773](https://youtrack.jetbrains.com/issue/VIM-2773/Undoing-text-deletion-forces-the-editor-view-upwards-when-deleted-from-Visual-Line-or-via-dd) which I raised on YouTrack.

Fixes viewport displacement when undoing deletion of lines or blocks that are immediately followed by a space character.

Before:

before_scrollback

After:

after_scrollback

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/helper/UndoRedoHelper.kt
+++ b/src/main/java/com/maddyhome/idea/vim/helper/UndoRedoHelper.kt
@@ -42,6 +42,9 @@ class UndoRedoHelper : UndoRedoBase() {
     val fileEditor = TextEditorProvider.getInstance().getTextEditor(editor.ij)
     val undoManager = UndoManager.getInstance(project)
     if (undoManager.isUndoAvailable(fileEditor)) {
+      val scrollingModel = editor.getScrollingModel()
+      scrollingModel.accumulateViewportChanges()
+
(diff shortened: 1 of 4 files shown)


AlexPl292:
This looks good for me. @citizenmatt Could you please also take a look, I know you like scrolling?

citizenmatt:
LGTM. I'd like to understand why it's necessary, to see if we're doing something wrong, or if it's a quirk of IntelliJ's undo system. More importantly, I'd like to know if it's something we should be guarding against elsewhere, but I can look into that separately. 

(It appears to be related to caret positioning. The behaviour shows if the caret is moved to a column when it's moved away from the deleted line. If we `:set nostartofline`, so the caret column is maintained, and the caret column on the deleted line was 0, then we don't see the scrolling behaviour. If the caret is moved to any other column, then we see scrolling on undo)

citizenmatt:
Hmm. I've just seen the same behaviour if you use `Ctrl+z` to undo a `dd` operation, so we should probably look deeper into what's going on here. I think this is still a worthwhile change, though.

citizenmatt:
Oh, and forgot to say thanks @dominicpalmer! Nice work 😁

PR #604, opened 2023-02-26
VIM-2773: Include empty lines in the selection when dragging backwards from EOL

dominicpalmer:
A fix for an [issue](https://youtrack.jetbrains.com/issue/VIM-2773/Undoing-line-block-deletion-of-text-thats-immediately-followed-by-a-space-forces-the-editor-to-scoll-upwards) I raised on YouTrack.

In Vim, when starting a selection from an empty line and dragging backwards (or vertically upwards), the entirety of the empty line (including the EOL character) should be included as part of the selection. Hence when deleting the block, the empty line should also get removed. IdeaVim didn't adhere to this behaviour. This PR fixes it.

Before:

before

After: 

after


AlexPl292:
I used to think I caught all mouse-selection related bugs. Thank you for the fix!

PR #606, opened 2023-03-01
VIM-1735: fix too many carets being added when in visual line mode

rickythefox:
Closes #VIM-1735 (https://youtrack.jetbrains.com/issue/VIM-1735/IdeaVim-multiple-cursors-visual-block-to-multi-cursor-ads-one-caret-too-much).

There might be a better way to determine whether the last selected line is the last line of the file, but I'm not familiar enough with the codebase. Feel free to suggest a better solution.


rickythefox:
Maybe the problem needs an explanation:

Before:
Kapture 2023-03-03 at 12 31 30

After:
Kapture 2023-03-03 at 14 05 37

AlexPl292:
Hi there! Thank you for your interest. However, I've decided to go with a bit different approach: instead of removing unnecessary caret, we just don't add the one that is not needed here: [VIM-1735](https://youtrack.jetbrains.com/issue/VIM-1735)

PR #611, opened 2023-03-09
VIM-2887: Fix:Include escaped character in search target

kawagh:
A fix for an issue I reported on YouTrack.
The Issue link is:
https://youtrack.jetbrains.com/issue/VIM-2887/action-leads-to-unexpected-destination-in-codes-containing-escaped-parentheses

I did `%` several times on parentheses below moives.

## before

https://user-images.githubusercontent.com/45124565/223920790-4fd2bfd5-e974-4e48-9fc0-ac3fb55007d4.mp4

## after

https://user-images.githubusercontent.com/45124565/223920819-d58b4501-108c-4d75-aa4d-8f5798a8a4b0.mp4


AlexPl292:
Hi there. Not sure why it was `false` in that method, but looks like it really should be `true`. Could you please write some tests to secure this behavior?

kawagh:
Thank you for reviewing. I added tests.

AlexPl292:
Awesome, thank you for your contribution!

PR #612, opened 2023-03-11
Fix quick-scope setup documentation

danisevas:
It was pointing to the wrong plugin to install


AlexPl292:
Oh, classic copy-paste issue. Thank you for the fix!

PR #615, opened 2023-03-26
Introduce strongly typed properties for options

citizenmatt:
After the last set of refactorings on the options code, we had a number of places where we were doing upcasts from `VimDataType` to e.g. `VimString` or `VimInt` with helper methods such as `getKnownStringOption`. This is because while we know the type of the option when we create it, the option is stored by name as an `Option<VimDataType>`. We need to upcast to get the value of the option to consume. This is mostly handled by the `OptionsValueAccessor`, but doesn't help when trying to set the option in code.

This PR introduces strongly typed properties for each option, and that type information flows through the setters and getters in the `VimOptionGroup` interface, so we can get and set strongly typed option values without needing to cast.

There is special case handling for the `'clipboard'` option. The default value is overridden by the IntelliJ implementation of IdeaVim to include the `ideaput` value, which uses the platform's paste processors when pasting text.


src/main/java/com/maddyhome/idea/vim/group/OptionGroup.kt
```diff
@@ -8,124 +8,33 @@
...
-      StringOption(
-        OptionConstants.clipboard,
-        OptionConstants.clipboardAlias,
-        "ideaput,autoselect,exclude:cons\\|linux",
-        isList = true,
-      ),
-    )
-
-    addOption(ToggleOption("unifyjumps", "unifyjumps", true))
+    IjOptions.initialise()
```
AlexPl292, line 15: Is there a chance that `Options.clipboard` will be accessed before the `OptionGroup` is initialized?

AlexPl292:
I have a weird feeling that we already had field-per-option in the form `val option = addOption(Option(....))`.
Looks like options API makes some kind of journey XD

citizenmatt:
Yes, we've gone back to the future! 😁 

I missed some extra context of this PR as I was trying to get everything sorted before travelling. This was following a discussion with @lippfi about some of the problems with the previous refactor, as mentioned in the main post, mainly that while we ended up with a good story for getting values, setting values was messy. It's true that the majority of usages are getters, but we can still make the API for setting values better.

And there is (perhaps unsurprisingly) a part 2 to this PR, where I started to have some fun. Unfortunately, it took me a while to realise that I can't create a draft PR based on a PR, because there isn't a base branch in the target repo, so I've had to create it in my fork - see [citizenmatt/ideavim#92](https://github.com/citizenmatt/ideavim/pull/92).

The draft PR adds to this PR by providing delegated properties that give read/write access to the underlying value of the option, which I think ends up with a simpler API for getting and setting option values, although it means working with `VimDataType` instances rather than native types. More details over in my fork.

So, the point of this PR was to try and improve the story for setting option values, and the only real way to do that is to make the options strongly typed, and therefore expose them as properties (again). The draft PR is a bit of an experiment that takes it a step further and provides a facade similar to the previous accessor that makes it easier to get and set values. Both are intended to be a request for comments.

AlexPl292:
After taking a look at the second PR, I've got a second flash of "I've seen this before" :)

I'd really appreciate it if you could write down the "decisions" in KtDoc of some options-related file. So we could see something like "we did it like this and not like that because this is good *list of arguments* and that is bad *list of arguments*. And if in the next N years someone will think "hmmm, looks like I know how to make an options API better", they'll be able to check our journey with options API.
The questions I'd love to see at least:
- StringOptions with `isList` property vs separate StringListOption (from the second PR)
- Access option by string name vs access by field
- Value of option as java data type (boolean, string) vs vim data types (VimBool, VimString)

citizenmatt:
Yep, I can add that kind of comment. Just quickly here and now:

1. `StringOption` with `isList` vs `StringListOption`. In the second PR this is necessary so it can be an `Option<VimList>`, which means we can treat the underlying value as a list of strings, rather than having to parse it whenever it's accessed. We have previously had `StringListOption`, but I don't really know why it was refactored into `StringOption`.
2. Access by field is better than by name because we can be strongly typed. By name only gives us `Option<VimDataType>`, but the field allows the definition (e.g. `StringOption` or `ToggleOption`) to flow type information through `getObjectValue` and `setObjectValue` and we can get/set strongly typed options without requiring helper functions to cast to the appropriate types
3. The use of `VimDataType` is probably best answered by @lippfi, but AIUI it allows us to easily use options as VimScript variables, because they have the right type. Although this means I've got a problem with the second PR, because Vim treats e.g. `'clipboard'` as a string, not as a list (`type(&clipboard)` returns 1, which means string)

src/main/java/com/maddyhome/idea/vim/group/OptionGroup.kt
citizenmatt: This is trickier than I realised. The idea of storing the option definitions inside `VimOptionGroup` would create a circular dependency - when `Options` is initialised, it would call into `VimOptionGroup` which might need to initialise and which calls `IjOptions.initialise`, which adds option definitions via `Options`, which would try to call into `VimOptionGroup` again.

We could make `Option.defaultValue` internal, so no-one could access it except via `VimOptionGroup`, but internal just means internal to vim-engine, so it doesn't offer much protection.

I think the next best alternative is for `Options.init` to notify _someone_ that options have been initialised, at which point, _someone_ could override `Options.clipboard` (and probably set up the IntelliJ specific options too). I'm not sure who _someone_ is though. It might be an implementation interface for `VimOptionGroup`, or even a new service that does nothing in vim-engine, but can be overridden in the IntelliJ build. What do you think?

citizenmatt: Or, we take `ideaput` out of the defaults and have a notification like we do for `ideajoin`. The first time we put without `ideaput` set, show a notification, offer to modify `.ideavimrc` and set a flag to not show the notification again.

citizenmatt:
OK, rebased against master and force pushed. Added some comments in `Option.kt` and `Options.kt` and added an extra commit that makes option initialisation deterministic with an explicit `initialiseOptions` call in `VimPlugin.turnOnPlugin`.

lippfi:
`StringOption with isList vs StringListOption. In the second PR this is necessary so it can be an Option<VimList>, which means we can treat the underlying value as a list of strings, rather than having to parse it whenever it's accessed. We have previously had StringListOption, but I don't really know why it was refactored into StringOption.`
@citizenmatt  It was done this way because I guess it's the way it is done in Vim. It looks like there are no list options, but only string ones. 
As you said, "type(&clipboard) returns 1" and I guess some scripts (plugins) might depend on option type and fail for lists e.g.
`let newOptionValue = &clipboad .. ",unnamed"` will throw an exception for list type

But
I don't think that my refactoring was the best solution, I mean using option as a wrapper for `VimDataType` is kinda staightforward and not the smartest way to deal with options. 
Lately I've been thinking more and more about Kotlin configurations as an alternative to Vim Script configurations. 
And by using VimDataType I've tightly tied the architecture to VimScript and now we're deprived of the flexibility I'd like to have. 
Now I think it's a good idea to abstract from Vim Script, i.e. have StringOption, NumberOption, ListOption (DictionaryOption, or any option we need) which will turn into VimDataType only when needed and accessed from VimScript (say, by calling Option.toVimDataType method)
I would love to hear your thought about this

lippfi (approved):

citizenmatt:
As long as each option has its correct type, I don't think I have a strong opinion about storing that as a Vim data type or a native type, tbh. It's working nicely as it is right now, but I see the point that if you're going to access it from Kotlin, and it's being written in Kotlin, then it makes more sense to use Kotlin types and convert to Vim types as necessary. I was wrong to use `VimList`, though, as a string list is a string, and I've fixed that in my updated second PR (and the accessor properties are working with Kotlin types now too)

AlexPl292:
Matt, thank you for your work! Looking forward for the next PR :D

citizenmatt:
You won't have to wait long 😆

PR #638, opened 2023-05-09
VIM-2615 add support to sort u command

samabcde:
1. support `sort u`
2. fix natural sort issue when both string not contain number
3. refactored `SortCommandTest`


lippfi (approved):
Thank you for this PR!

AlexPl292:
Hi there, thank you for your contribution.
I'm not sure, will it work fine with the `i` flag and two lines with the same word of different cases? I guess, `distinct` won't eliminate such lines, will it?
I've tried to find this case in test cases, but looks like it's not tested. Something like:

```
content = """
data
DATA
""",
command = "sort ui",
expected = "data",
```

AlexPl292:
Okay, this seems to be a better solution. Glad to see this supported, thank you for your PR!

PR #646, opened 2023-05-31
VIM-1990 fix repeat delete find or till wrong MotionType

samabcde:
When doing repeat delete of find/till,
MotionType is determined by actually motion direction
- `t`,`f` -> Forward -> Inclusive
  `T`,`F` -> Backward -> Exclusive
- `,` will reverse direction 
   `;` will not.


AlexPl292:
Very nice, definitely worth merging!
Thank you!

PR #647, opened 2023-06-01
Annotations

AlexPl292:


vim-engine/src/test/resources/engine_ex_commands.yaml
```diff
@@ -0,0 +1,129 @@
+# This file was automatically generated by [com.intellij.vim.processor.EXCommandProcessor].
```
AlexPl292, line 1: Why do we generate commands for test module?

annotation-processors/src/main/kotlin/com/intellij/vim/model/LazyInstance.kt
```diff
@@ -0,0 +1,22 @@
...
+ * license that can be found in the LICENSE.txt file or at
+ * https://opensource.org/licenses/MIT.
+ */
+
+package com.intellij.vim.model
+
+import java.lang.invoke.MethodHandles
+import java.lang.invoke.MethodType
+
+open class LazyInstance<T>(private val className: String, private val classLoader: ClassLoader) {
```
AlexPl292, line 14: Why do we extract this class from `LazyVimscriptFunction`?

build.gradle.kts
```diff
@@ -118,6 +133,9 @@ dependencies {
 
     api(project(":vim-engine"))
 
+    ksp(project(":annotation-processors"))
+    implementation(project(":annotation-processors"))
```
AlexPl292, line 137: Could you please note why do we use `implementation` here and not `compile`?

vim-engine/src/test/resources/engine_ex_commands.yaml
lippfi: My bad, I thought you can't access  main resources from tests

annotation-processors/src/main/kotlin/com/intellij/vim/model/LazyInstance.kt
lippfi: I'm not sure if I understood you correctly. If you're asking "Why do we have `LazyVimscriptFunction` when we have a `LazyInstance`?", then the reason is that VimScript functions have a lateinit name field, which should be handled.

Surely, I could use only the `LazyInstance` and register all the functions in some kind of a map, such as `nameToFunction`. However, it's more convenient to have a class that carries out all the required initialization on its own, rather than doing half of the initialization with the help of `LazyInstance` and manually setting the function name later.

build.gradle.kts
lippfi: Because the `LazyInstance` class is a part of "annotation-processors". I can move it somewhere and make it compile, but I'm not sure where to move it and if it is a good idea

lippfi: Well, I've just made it compileOnly to avoid transitive dependency to kotlinx.serialization

annotation-processors/src/main/kotlin/com/intellij/vim/FileWriter.kt
```diff
@@ -0,0 +1,19 @@
...
+ * license that can be found in the LICENSE.txt file or at
+ * https://opensource.org/licenses/MIT.
+ */
+
+package com.intellij.vim
+
+import java.io.File
+import java.nio.file.Path
+
+class FileWriter {
```
AlexPl292, line 14: This class can be inlined, I guess?

PR #654, opened 2023-06-26
Allow invoking alt+enter intentions as well as actions

citizenmatt:
Useful for mapping alt+enter menu items to an IdeaVim shortcut (e.g. the alt+enter intention action might show a popup of actions, but that popup doesn't have it's own action that can be invoked by shortcut). The intention can already be assigned an IntelliJ shortcut, but this allows mapping to IdeaVim, which can be handy if several shortcuts share the same start key, e.g. `<leader>`.

Uses the same `<action>(...)` syntax as invoking actions, but requires knowing the intention class/implementation class name.

The "easiest" way to discover the intention name is to use the alt+enter menu, expand the options sub-menu, and assign a shortcut. Then enable "Track Action IDs" and invoke the intention with the shortcut. The notification will show the temporary action created for the shortcut - `IntentionActionAsAction_{intention-class-name}`. Strip off the `IntentionActionAsAction_` prefix and use the intention's class name in the mapping.

E.g.:

```
nmap <leader>a <Action>(com.intellij.ml.llm.intentions.chat.AIAssistantIntention)
```

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/helper/IjActionExecutor.kt
+++ b/src/main/java/com/maddyhome/idea/vim/helper/IjActionExecutor.kt
@@ -8,12 +8,17 @@
 
 package com.maddyhome.idea.vim.helper
 
+import com.intellij.codeInsight.intention.IntentionActionDelegate
+import com.intellij.codeInsight.intention.IntentionManager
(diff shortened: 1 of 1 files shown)


AlexPl292:
We discussed this in private conversation and I don't think this is a good idea to merge this PR. So, I'll close it.

PR #659, opened 2023-07-02
VIM-2920 fix select block expansion when enclose boundary is line break

samabcde:
1. Fix block select expansion when there is line break after open(e.g "(") or before close (e.g. ")") 
2. Add test to cover `findBlockRange` function, so we have more confidence in later refactoring
3. Add Javadoc for easier understanding `findBlockRange`


AlexPl292:
This looks good, thank you for the great tests!

PR #668, opened 2023-07-24
Handle unicode grapheme clusters

ludwig-jb:
Some characters that render as a single symbol can span over a sequence of several unicode code points (e.g., flag emojis, combination of a letter and a diacritic, Hangul syllables, etc.).

Such composites are called grapheme clusters in the unicode standard, and this patch introduces recognition of extended grapheme cluster boundaries, allowing to iterate over rendered characters. Without this, user may observe the cursor being "stuck" inside a character for several keystrokes, while it's making its way through each code point in the grapheme cluster.

The implementation follows the boundaries search algorithm outlined in the technical report 29 of the Unicode standard[1]. The implementation was tested against the set of test cases provided by the unicode character database[2].

Additionally to the grapheme cluster boundaries search itself, this patch adds `isExtendedPictographic` function, that answers whether the given code point has a unicode "Extended_Pictographic" property, which is required to correctly determine grapheme cluster boundaries. This method is implemented natively in the JDK 21 and can be removed once we start targeting that version.

Extended_Pictographic property is stored as a bitmap. I was considering making a similar map for the code point classification in the grapheme cluster boundary search implementation, which could yield better performance, but that would require adding another half a megabyte (at least) of data into the JAR and I've settled for the bunch of `if`s way.

That is something that can be reconsidered and shouldn't be difficult to change if the impact on performance would be noticeable (in my simple tests it didn't show).

A few functions in the vim-engine were adjusted to handle grapheme clusters (such as getting the horizontal offset and adjusting the cursor to not reach over the end of the line).

[1]: https://www.unicode.org/reports/tr29/#Grapheme_Cluster_Boundaries
[2]: https://www.unicode.org/Public/UCD/latest/ucd/auxiliary/GraphemeBreakTest.txt


vim-engine/src/main/kotlin/com/maddyhome/idea/vim/common/Graphemes.kt
```diff
@@ -0,0 +1,255 @@
+package com.maddyhome.idea.vim.common
+
+/**
+ * Move over unicode extended grapheme cluster boundaries.
+ *
+ * https://www.unicode.org/reports/tr29/#Grapheme_Cluster_Boundaries
+ */
+public object Graphemes {
+  public fun next(charSeq: CharSequence, start: Int): Int? {
+    if (start < 0 || start >= charSeq.length) return null
```
AlexPl292, line 10: I think it's better to throw an error in this case

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/common/Graphemes.kt
```diff
@@ -0,0 +1,255 @@
...
+ */
+private fun classify(codePoint: Int): CodePointType {
+  when (codePoint) {
+    0xD -> return CodePointType.CR
+    0xA -> return CodePointType.LF
+    in 0 until 0x20 -> return CodePointType.CONTROL
+    in 0 until 0x80 -> return CodePointType.OTHER
+  }
+
+  if (isExtendedPictographic(codePoint))
```
AlexPl292, line 181: Please add a brackets for the two-lines `if`

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/common/ExtendedPictographics.kt
```diff
@@ -0,0 +1,81 @@
...
+  val bit = codePoint % 64
+  val bucket = bitmap[idx]
+
+  return (bucket and (1L shl bit)) != 0L
+}
+
+// A bitmap that maps a code point into whether it has the Extended_Pictographic property.
+// The code points go in increasing order by index and in reverse order by bit in a specific long.
+// This way a simple divmod is enough to compute both indices.
+private val bitmap = longArrayOf(
```
AlexPl292, line 22: Where this bitmap comes from?

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/common/Graphemes.kt
```diff
@@ -0,0 +1,255 @@
...
+}
+
+private inline fun CharSequence.nextBoundary(
+  start: Int,
+  crossinline next: (Int, Int) -> Int,
+  crossinline nextCode: CharSequence.(Int) -> Int?,
+  crossinline prevCode: CharSequence.(Int, Int) -> Int?,
+  swap: Boolean,
+): Int {
+  var current = start
```
AlexPl292, line 45: I see a kind of a fancy logic here. Don't you think there is a library for that by java or unicode?

AlexPl292 (commented):
Hi, I like this change! I've tried to do something similar, but I didn't know this thing is so good defined by unicode.

Also, you mentioned that the code is tested against test cases provided by unicode. Was this a manual testing? Is there is chance to inlude these test cases to the IdeaVim tests?

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/common/Graphemes.kt
ludwig-jb: The second half of the condition (`start >= charSeq.length`) is like that, so we can call `next` in a loop and have `null` as a break condition instead of tracking the offset. `start < 0` makes zero sense here though, and for sure should throw. Would you prefer to make the return type non-nullable and move the check up to the caller's responsibility?

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/common/ExtendedPictographics.kt
ludwig-jb: I've generated it with a script from emoji-data.txt from the unicode character database. I could add the source file into the repository and add some code that would generate this bitmap on demand, is there any infrastructure of that in idea-vim or what would be the preferred way to do this?

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/common/Graphemes.kt
ludwig-jb: There is the library by the unicode organisation to handle unicode called ICU4J. But it is a pretty big library and we only need a tiny fraction of its functionality, I got a sense that idea-vim avoids unnecessary dependencies, so decided to implement it myself, given that the ruleset is fairly straightforward and compact (around a dozen rules total). Also, their API isn't great for this, the way they expose the functionality is through the implementation of the `java.text.BreakIterator` interface, which is a little awkward to use for this. 

We could technically switch to ICU4J, but I'm not sure of the merits of this switch.

*from the top-level comment*
> Also, you mentioned that the code is tested against test cases provided by unicode. Was this a manual testing? Is there is chance to inlude these test cases to the IdeaVim tests?

Yes, the technical report that introduces the notion of the extended grapheme cluster has a file with the test cases. I'll add it to the repo and write some tests that make use of it.

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/common/ExtendedPictographics.kt
AlexPl292: This won't be needed with JDK 21, right? If so, it's okay to just not where this map comes from so the people didn't wounder about these magic numbers.

ludwig-jb: Yes, JDK21 contains this functionality. Very well, I'll add a link to the file in the comment then.

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/common/Graphemes.kt
AlexPl292: Regarding ICU4J: let's keep your implementation then, but please add a note with the rationale about implementing this code by hand.

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/common/Graphemes.kt
AlexPl292: I see, thank you for the explanation. I think we can add this contract to the KDoc and keep it like this.

PR #669, opened 2023-08-01
Do not enter INSERT mode for read-only editors

citizenmatt:
When initialising a new editor, IdeaVim will automatically start INSERT mode for an editor that is not a primary editor - that is, it's not file-backed, so it's not a main editor in the content area of the IDEA. This provides a more intuitive "just start typing" experience for editors that are more likely to be part of the UI than editing files, e.g. the VCS commit message editor.

IdeaVim would check that the editor was not file-backed, and that the document was writable, and then start INSERT mode. However, it is not enough that the document is writable, the editor needs to be non-read-only, too. Log and console output for run configurations and tests will have a writable document (because the app is updating it while the file is being externally modified), but the editor is in "viewer" mode. This PR will not start INSERT mode if the editor is read-only ("viewer" mode, but also "rendered" mode, which is read-only with hidden caret).

Fixes [VIM-2313](https://youtrack.jetbrains.com/issue/VIM-2313), fixes [VIM-2318](https://youtrack.jetbrains.com/issue/VIM-2318), fixes [VIM-2666](https://youtrack.jetbrains.com/issue/VIM-2666), fixes [VIM-2951](https://youtrack.jetbrains.com/issue/VIM-2951)

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/group/EditorGroup.java
+++ b/src/main/java/com/maddyhome/idea/vim/group/EditorGroup.java
@@ -207,8 +207,18 @@ public class EditorGroup implements PersistentStateComponent<Element>, VimEditor
     VimPlugin.getKey().registerRequiredShortcutKeys(new IjVimEditor(editor));
 
     initLineNumbers(editor);
-    // Turn on insert mode if editor doesn't have any file
-    if (!EditorHelper.isFileEditor(editor) &&
+
+    // We add Vim bindings to all opened editors, even read-only editors. We automatically start INSERT mode for editors
+    // that are not the traditional file-backed editor to provide a more intuitive "just start typing" experience for
+    // editors that are more likely to be used as part of the UI, such as the VCS commit message editor, as opposed to
+    // an editor for a source file in the project.
+    // We do not start INSERT mode if the editor is in read-only "viewer" mode (which includes "rendered" mode, which
+    // is read-only and also hides the caret). This means we start INSERT mode for in-memory editors such as the VCS
+    // commit message, but start in the usual COMMAND mode for read-only editors with writable documents, such as log
+    // or console output for run configurations and tests (the document is writable because the app is updating it. It's
+    // not user-writable.)
+    if (!editor.isViewer() &&
+        !EditorHelper.isFileEditor(editor) &&
         editor.getDocument().isWritable() &&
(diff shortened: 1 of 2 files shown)


citizenmatt:
Note that the linked YouTrack issues report seeing a tooltip that states "This view is read-only". This is because `InsertExitModeAction` is marked as `Command.Type.INSERT`. This seems unexpected, but `<Escape>` can insert text when exiting insert mode after visual block mode, e.g. `<C-V>jjjI{some text}<Escape>`. This PR adds an explanatory comment to `InsertExitModeAction`.

AlexPl292:
This is impressive that such a small change fixes so many highly voted issues.
Awesome!

neshkeev:
@AlexPl292 , @citizenmatt , the issue seems to came back
image

It also happens sometimes when an external library's source file is opened in debug mode, for example, I hit "Step Out" in the debugger, an exception is thrown and IDEA automatically opens the nearest `catch` section.

It can occur when executing an invalid SQL statement via JDBC

citizenmatt:
What version are you using? This has been refactored recently by @lippfi, perhaps this is related? Could you create a new issue on youtrack.jetbrains.com with steps to reproduce, please?

lippfi:
I guess it could happen after focusing console after being in Editor in INSERT mode. Now the modes are global for all editors, so it remains in INSERT
We need to implement some listener that will go to NORMAL after focusing read only editors

neshkeev:
One of the recent updates switched from the NORMAL mode by default to INSERT mode by default, so when I open a Library's source code I can't navigate over the file, which is extremely annoying, so I have to use the mouse to enter the VISUAL mode and only then I can hit ESC to get back to the NORMAL mode.

Could you please bring back the NORMAL mode by default?

IDEA Info:
```
IntelliJ IDEA 2024.1.1 (Community Edition)
Build #IC-241.15989.150, built on April 29, 2024
Runtime version: 17.0.10+1-b1207.14 aarch64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
macOS 14.5
GC: G1 Young Generation, G1 Old Generation
Memory: 16384M
Cores: 16
Metal Rendering is ON
Non-Bundled Plugins:
  idea.plugin.protoeditor (241.15989.49)
  Docker (241.15989.111)
  IdeaVIM (2.15.0)
Kotlin: 241.15989.150-IJ
```

PR #672, opened 2023-08-03
Fix issue adding back hidden status bar widget

citizenmatt:
The 2023.2 platform requires `statusBarWidgetFactory` to have an `id` attribute in `plugin.xml`. This must be set to the same value returned by `StatusBarWidgetFactory.getId()`. If this value is missing, then hiding the status bar widget through the right click menu and then trying to show it again will fail.


AlexPl292:
Thank you!

PR #675, opened 2023-08-11
VIM-3042: Don't adjust caret with startsel and don't push multiple insert modes

shaunpatterson:
See VIM-3042. 
https://youtrack.jetbrains.com/issue/VIM-3042/Select-and-delete-from-insert-mode-moves-cursor-position


AlexPl292:
Hi, can you please write a few tests for this fix?

shaunpatterson:
I think the mode stack is correct.  If I were to start a selection in insert mode with Shift+arrows, then delete, I would expect to still be in insert mode.  I think this is correct... at least from logical user perspective.  It may not be exactly what Vim would do but it's more of a hybrid vim/normal editing.

PR #715, opened 2023-09-29
Add VimShowHoverInfo action

citizenmatt:
The `ShowHoverInfo` action has just been added to the 2023.3 wave of releases (233.8745 and above). This action will show a combined tooltip of highlighting info and quick documentation, exactly the same as shown when hovering over a code element. This can be mapped to keys, such as `nmap gh <Action>(ShowHoverInfo)`.

This PR adds a safe workaround for 2023.1 and 2023.2. It adds a `VimShowHoverInfo` action that will defer to `ShowHoverInfo` if it exists, and calls an appropriate platform API if not. This API is intended to show the tooltip in response to an `EditorMouseEvent`, so this action simply creates a fake instance for the location of the text caret.

It also bumps the `since-build` value in `plugin.xml` to match 2023.1 RTM, to allow using this platform API (it was previously on an earlier 231 build).

Fixes [VIM-2106](https://youtrack.jetbrains.com/issue/VIM-2106/Go-hover)

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/action/VimShowHoverInfoAction.kt
+++ b/src/main/java/com/maddyhome/idea/vim/action/VimShowHoverInfoAction.kt
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2003-2023 The IdeaVim authors
+ *
+ * Use of this source code is governed by an MIT-style
+ * license that can be found in the LICENSE.txt file or at
+ * https://opensource.org/licenses/MIT.
+ */
+
+package com.maddyhome.idea.vim.action
+
+import com.intellij.codeInsight.hint.HintManagerImpl.ActionToIgnore
+import com.intellij.openapi.actionSystem.ActionUpdateThread
+import com.intellij.openapi.actionSystem.AnAction
+import com.intellij.openapi.actionSystem.AnActionEvent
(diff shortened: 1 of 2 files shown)


citizenmatt:
Yep. It's to allow hover to work in 231 + 232, but it will be an action that we'll keep forever to maintain compatibility. Or we could introduce a mapping so that the `VimShowHoverInfo` action is automatically mapped to `ShowHoverInfo`, but we'd have some kind of code like this forever.

AlexPl292:
I don't think we should add any tricks here. This is a feature of IntelliJ IDEA and not of IdeaVim, so we can just write to the docs that there is such action available, but only if you use IJ >= 233.

Also, there is a very good chance that we'll increase the minimal version of IJ for IdeaVim to 233 because of code compatibility issues. So, this code won't cover 231 and 232 versions.

citizenmatt:
OK. Happy to go with whatever you decide.

As a side note, if we did take this and then want to deprecate it when we move to 233, we could add a default/builtin mapping of:

```
map <Action>(VimShowHoverInfo) <Action>(ShowHoverInfo)
```

which is a nice workaround to invoke the `ShowHoverInfo` action from an existing map that's intending to invoke `VimShowHoverInfo`.

PR #726, opened 2023-10-16
Fix range for fall back comment mode

citizenmatt:
Callback to reset caret was always called, even if action was unavailable. Fall back action would therefore act on current line only.


AlexPl292:
Looks good for me. Thank you!

PR #729, opened 2023-10-19
Add operating system type to `has()` function

chylex:
I tried to replicate the logic from https://github.com/vim/vim/blob/0f058d13206665bad37c7d42834cfa0075f50239/src/evalfunc.c#L5511, not 100% sure if I got it right (and some special cases like `win32unix` are missing) but hopefully it's ok.


AlexPl292:
I see there is a pretty hairy set of ifs there, but I think we can have a simplified version. Thank you!

chylex:
Yea, wasn't sure if there's a better way to write it considering all the quirks of how vim decides which OS features to use, like how `win32` being included in 64-bit Windows, or how MacOS is based on BSD but isn't supposed to have the `bsd` feature. I initially considered a `switch` over `OS.CURRENT`, but that doesn't include all the options (Solaris).

PR #730, opened 2023-10-19
Fix(VIM-3116): Cannot return to Normal mode in read-only file with ideaglobalmode

chylex:
Implemented using a focus listener, which has a disadvantage that it fires on every focus and not just when switching focus from a different editor. I considered `FileEditorManagerListener.selectionChanged`, but that event doesn't fire for editors outside the main editor tab pane, such as diff previews.


chylex:
The most recent `master` commit the PR is based off has unit test failures.

AlexPl292:
Hi, thank you for your changes. However, this option is really experimental and we've noticed a lot of other issues when it's enabled. Currently, we don't have plans to invest in this option, so I'll close the PR.
I've added a note to the code saying that this option is experimental.

PR #731, opened 2023-10-20
`gx` command

leonspohr:
I have gone ahead and added the missing `gx` command. 

Concerning issue:
https://youtrack.jetbrains.com/issue/VIM-1341/gx-should-open-selection-in-browser


AlexPl292:
Hi there! I see that there are 116 changed files and most of them contain formatting of them. Even if this might be a good thing to format the files, I wouldn't add such a huge change to the pull request. Could you please revert the formatting changes and leave only changes with the `gx` command?

leonspohr:
Well, I agree that those formatting shouldnt be included here. But those changes were not done by me, but by ideavim bot

AlexPl292:
This is unexpected, as I understand, the actions should be explicitly enabled on forks. Anyway, to avoid such issues, I've added a condition for actions to work on `JetBrains/ideavim` repository.

Regarding the change: Unfortunatelly, with this change the `gx` command will not only open links, but also jump to the declarations or usages. This is undesired behavior, so it looks like we need some verification that we're working on the URL.

PR #733, opened 2023-10-25
Add support for ShowHoverInfo action to 2023.1 and 2023.2

citizenmatt:
The 2023.3 IntelliJ Platform introduces an action called `ShowHoverInfo` that can be used to show the same tooltip which is displayed when the mouse hovers over a code symbol. This PR adds an extension point that will register an action with the same name for IntelliJ Platform versions that don't support this action. Once the base required version is raised to 2023.3, then this customisation can be removed, and all existing mappings will still work.

The base required version is raised to 2023.1 RTM to allow using the required platform API for the new action. It was on a preview version of 2023.1.

Users can invoke the action from a mapping, such as:

```vimL
nmap gh <Action>(ShowHoverInfo)
```

This PR fixes [VIM-2106](https://youtrack.jetbrains.com/issue/VIM-2106/Go-hover)

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/action/VimActionConfigurationCustomizer.kt
+++ b/src/main/java/com/maddyhome/idea/vim/action/VimActionConfigurationCustomizer.kt
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2003-2023 The IdeaVim authors
+ *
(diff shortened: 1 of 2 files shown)


AlexPl292:
Thank you!

PR #734, opened 2023-10-27
Support `~/` on Windows

citizenmatt:
Vim supports `~/` and `~\` on all platforms. This PR helps my muscle memory so I can type `:e ~/.ideavimrc` on Windows and it works the same in both IdeaVim and Vim.

Vim supports backslashes on Unix (tested on Vim) but both forward- and back-slashes on Windows. VFS supports both forward- and back-slashes on all platforms.


AlexPl292:
Great, thank you!

PR #755, opened 2023-11-16
VIM-3176 fix restore selection after pasting before/in/after selection

samabcde:


samabcde:
Close as already fixed in https://github.com/JetBrains/ideavim/commit/8fcca05565fddc18f2189dbe65de6386d9d19f28

PR #763, opened 2023-11-25
Fix(VIM-3176) add test for restore selection after pasting in/below s…

samabcde:
…election


lippfi:
Thanks again for identifying the exact place in the code causing the bug.
And thank you for this tests!

PR #778, opened 2023-12-11
Showmode

lippfi:
Add showmode & macro recording widgets


vim-engine/src/main/kotlin/com/maddyhome/idea/vim/common/VimListenersNotifier.kt
```diff
@@ -0,0 +1,38 @@
...
+
+package com.maddyhome.idea.vim.common
+
+import com.maddyhome.idea.vim.api.VimEditor
+import com.maddyhome.idea.vim.state.mode.Mode
+
+public class VimListenersNotifier {
+  public val modeChangeListeners: MutableList<ModeChangeListener> = mutableListOf()
+  public val editorFocusListeners: MutableList<EditorFocusListener> = mutableListOf()
+  public val macroRecordingListeners: MutableList<MacroRecordingListener> = mutableListOf()
```
AlexPl292, line 17: All these fields **must** be encapsulated. Only `add`/`remove` method should be available for the plugin writers. Otherwise, we may get a mess when the plugin writers will modify the listeners as they want.

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/common/VimListenersNotifier.kt
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+package com.maddyhome.idea.vim.common
+
+import com.maddyhome.idea.vim.api.VimEditor
+import com.maddyhome.idea.vim.state.mode.Mode
+
+public class VimListenersNotifier {
+  public val modeChangeListeners: MutableList<ModeChangeListener> = mutableListOf()
+  public val editorFocusListeners: MutableList<EditorFocusListener> = mutableListOf()
```
AlexPl292: These listeners should be migrated to kotlin flows at some moment, as they're more effective.

src/main/java/com/maddyhome/idea/vim/group/IjOptionProperties.kt
```diff
@@ -29,6 +29,7 @@ public open class GlobalIjOptions(scope: OptionAccessScope) : OptionsPropertiesB
   public val lookupkeys: StringListOptionValue by optionProperty(IjOptions.lookupkeys)
   public var trackactionids: Boolean by optionProperty(IjOptions.trackactionids)
   public var visualdelay: Int by optionProperty(IjOptions.visualdelay)
+  public var showmodewidget: Boolean by optionProperty(IjOptions.showmodewidget)
```
AlexPl292, line 32: Is this a final name for this option? Are we going to keep the option if we decide to have this widget?

src/main/java/com/maddyhome/idea/vim/ui/StatusBarModeFactory.kt
```diff
@@ -0,0 +1,145 @@
...
+import com.maddyhome.idea.vim.state.mode.Mode
+import com.maddyhome.idea.vim.state.mode.SelectionType
+import com.maddyhome.idea.vim.state.mode.mode
+import java.awt.Component
+
+private const val ID = "IdeaVim::Mode"
+
+public class StatusBarModeFactory : StatusBarWidgetFactory {
+  private companion object {
+    private const val NO_MODE = "" // for cases were no editors are focused
```
AlexPl292: Why not to use getStatusString function?

src/main/java/com/maddyhome/idea/vim/listener/VimListenerManager.kt
```diff
@@ -344,7 +373,9 @@ internal object VimListenerManager {
     }
 
     override fun editorReleased(event: EditorFactoryEvent) {
-      injector.markService.editorReleased(event.editor.vim)
+      val vimEditor = event.editor.vim
```
AlexPl292, line 376: Are we sure this function is called when the editor is released? I remember there were some details about it.

src/main/java/com/maddyhome/idea/vim/ui/MacroWidgetFactory.kt
```diff
@@ -0,0 +1,101 @@
...
+  override fun getId(): String {
+    return ID
+  }
+
+  override fun getDisplayName(): String {
+    return "IdeaVim Macro Recording Widget"
+  }
+
+  override fun createWidget(project: Project): StatusBarWidget {
+    injector.listenersNotifier.macroRecordingListeners.add(macroRecordingListener)
```
AlexPl292: Does widget disapper when we turn off the plugin via status bar icon?

AlexPl292 (changes requested):

src/main/java/com/maddyhome/idea/vim/group/IjOptionProperties.kt
lippfi: Nope, this widget is a successor of the showmode widget that we used to have in status bar (which is not visible with the new UI)
I plan to collect feedback and replace the old widget with a new one, so it's a temporary option. In future the `showmode` option will be used.

src/main/java/com/maddyhome/idea/vim/ui/StatusBarModeFactory.kt
lippfi: For several reasons:
1. We use a different format for names (e.g., 'INSERT' instead of '-- INSERT --').
2. This widget is only needed for the 'showmode' option. It should be defined in the location where it's needed. There's no need to include this method in the VimStateMachine interface, and I would like to remove it along with the old widget.

src/main/java/com/maddyhome/idea/vim/listener/VimListenerManager.kt
lippfi: It is definitely called when the editor is released (I've just rechecked)

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/common/VimListenersNotifier.kt
```diff
@@ -0,0 +1,38 @@
...
+ * https://opensource.org/licenses/MIT.
+ */
+
+package com.maddyhome.idea.vim.common
+
+import com.maddyhome.idea.vim.api.VimEditor
+import com.maddyhome.idea.vim.state.mode.Mode
+
+public class VimListenersNotifier {
+  public val modeChangeListeners: MutableList<ModeChangeListener> = mutableListOf()
```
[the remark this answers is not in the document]
AlexPl292: How is this one resolved?

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/common/VimListenersNotifier.kt
AlexPl292: Why this one was marked as resolved?

src/main/java/com/maddyhome/idea/vim/ui/widgets/mode/VimModeWidget.kt
```diff
@@ -0,0 +1,131 @@
...
+
+  private fun getModeForeground(mode: Mode?): Color {
+    // TODO make it customizable via settings (color picker) or .idevimrc
+    return UIUtil.getPanelBackground()
+  }
+
+  private fun getModeBackground(mode: Mode?): Color? {
+    // TODO make it customizable via settings (color picker) or .idevimrc
+    return when (mode) {
+      Mode.INSERT -> Color(134, 174, 213)
```
AlexPl292, line 122: BTW, I see the colors don't adjust depending on the light/dark theme. How does this look with different themes?
Dark/light theme is a must-have modification, but we should probably even make a per-theme default value for these colors.
It is also interesting how a per-theme config can be set up in .ideavimrc.

lippfi: I've tried these colors with both light and dark themes, and they look good enough (at least to my taste), but of course it's more like a proof of concept and we most likely will change default colors after receiving feedback from users.
We could add Vimscript functions like getCurrentLAF and isDarkLAF, but I doubt that anyone will ever discover or use these functions. I think it would be better to make the colors customizable via a GUI after right-clicking on the widget. We could also provide two fields for each mode (for dark and light themes), but I don't think that using Vimscript is a good idea for several reasons:
1. This exact status bar and its colors do not exist in the original Vim, and there is no need to invent a Vimscript approach to change colors (especially when the LAF in the IDE is managed by a GUI).
2. Not many people are familiar with Vimscript; a right-click option would be much more intuitive.
3. It's counterintuitive to customize such a thing with Vimscript when most of Vim's built-in functions are not supported.

PR #780, opened 2023-12-11
Remove CommandFlags.FLAG_EXIT_VISUAL and FLAG_MULTIKEY_UNDO

citizenmatt:
This PR removes the `CommandFlags.FLAG_EXIT_VISUAL` and `CommandFlags.FLAG_MULTIKEY_UNDO` enum values.

The `FLAG_EXIT_VISUAL` enum value was used to decide if a visual operator should exit visual mode. All visual operators derive from `VisualOperatorActionHandler` and all had `FLAG_EXIT_VISUAL`, so the conditional behaviour can be made default, and the flag removed. Note that visual motions (such as `v_w`) are not visual operators, don't use the same codepath, and don't require the flag.

The visual operator handler also included logic to exit visual mode _after_ executing the operator, as well as before. This second mode reset is not required - we are no longer in visual mode at this time, unless the operator puts us there. By making it the responsibility of the visual operator implementation to ensure the correct mode, we can remove this check.

This simplifies `OperatorFunction`, which no longer needs `postProcessSelection` to help decide if we should exit visual mode. (This was required for the Commentary extension to allow keeping the selection while waiting for Rider's async line/block comment actions to complete).

Removing the check also removes the only usage of `FLAG_MULTIKEY_UNDO` outside of command declarations. It is unclear what this flag was intended for, but since it is no longer checked anywhere, it can be removed.

Finally, because all visual operators had `FLAG_EXIT_VISUAL`, the `editor.vimKeepingVisualOperatorAction` flag set as part of the visual operator execution would always be `false`. Since the flag is not set anywhere else, it can be removed too.

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/action/change/OperatorAction.kt
+++ b/src/main/java/com/maddyhome/idea/vim/action/change/OperatorAction.kt
@@ -18,9 +18,7 @@ import com.maddyhome.idea.vim.api.injector
 import com.maddyhome.idea.vim.api.setChangeMarks
 import com.maddyhome.idea.vim.command.Argument
 import com.maddyhome.idea.vim.command.Command
-import com.maddyhome.idea.vim.command.CommandFlags
 import com.maddyhome.idea.vim.command.OperatorArguments
-import com.maddyhome.idea.vim.state.mode.SelectionType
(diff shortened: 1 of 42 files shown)


AlexPl292:
Thank you for the code simplification!

PR #785, opened 2023-12-27
Introduce 'operatorfunc' option for custom operators

citizenmatt:
This PR introduces the `'operatorfunc'` option. This option contains the name of a function that can act like a custom operator, and is called by the `g@` command, such as in this example from the #702 discussion:

```vimL
nnoremap gx :set opfunc=Redact<cr>g@
function! Redact(type)
    execute "normal `[v`]rx"
endfunction
```

Using a command such as `gxiw`, this map will invoke the custom operator function on a given word, and replace it with `x` characters.

The `'opfunc'` option can be the name of a function, a reference from `function('…')` or `funcref('…')` or a lambda.

This PR also allows an extension to export a script function that can be called from the VimScript executor, which is a simple wrapper function that invokes extension code. A helper function has been added to create a script function with the right signature for operator functions. All extensions have been updated to export a function, and to use the `'operatorfunc'` option.

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/action/change/OperatorAction.kt
+++ b/src/main/java/com/maddyhome/idea/vim/action/change/OperatorAction.kt
@@ -14,6 +14,7 @@ import com.maddyhome.idea.vim.VimPlugin
 import com.maddyhome.idea.vim.api.ExecutionContext
 import com.maddyhome.idea.vim.api.VimCaret
 import com.maddyhome.idea.vim.api.VimEditor
+import com.maddyhome.idea.vim.api.globalOptions
 import com.maddyhome.idea.vim.api.injector
 import com.maddyhome.idea.vim.api.setChangeMarks
 import com.maddyhome.idea.vim.command.Argument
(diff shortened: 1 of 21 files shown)


AlexPl292:
Oh, wow

AlexPl292 (approved):

citizenmatt:
Rebased onto current master, fixing conflict.

citizenmatt:
Rebased onto current master and force pushed.

Also fixes `SetCommandTests` and friends by being more consistent with the temporary/hidden flag. Options that are feature flags for work-in-progress or diagnostic behaviour are hidden from normal use and unit tests. They are always visible for internal mode.

@AlexPl292 @lippfi are we happy to merge this?

AlexPl292:
Hi, yes, sorry for the delays. We haven't yet reviewed this, but the plan is to first merge the changes about the regex, then make a release, and then handle your changes to avoid too many refactorings in one release.

AlexPl292:
Merged! Thank you!

PR #787, opened 2023-12-29
Added "Which-Key" to Plugins

Infonautica:
- Mentioned a common and very useful plugin in "Plugins" document


AlexPl292:
Nice catch, thank you!

PR #788, opened 2024-01-03
Refactor VimOptionGroupBase

citizenmatt:
The `VimOptionGroupBase` class has become a bit of a god class/kitchen sink class. It has too many responsibilities, and is confusing to reason about. It handles setting/getting option values, initialising options for new windows and buffers, notifying listeners of changes and maintaining a cache of parsed effective values (e.g. parsing the contents of `'guicursor'`).

This PR splits the responsibilities into separate classes that `VimOptionGroupBase` uses. The classes are all implementation details, so remain in the same `VimOptionGroupBase.kt` file, but at least encapsulate responsibilities.

The majority of the changes are renames, extracting classes, and other similar refactoring. There is very little behaviour change, although a couple of minor bugs have been fixed.

This refactoring is the basis for the next PR, which will modify the option storage to allow custom options that are powered by IntelliJ settings (e.g. `'wrap'`, based on IntelliJ's soft wrap settings).

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/group/OptionGroup.kt
+++ b/src/main/java/com/maddyhome/idea/vim/group/OptionGroup.kt
@@ -40,10 +40,6 @@ internal class OptionGroup : VimOptionGroupBase(), IjVimOptionGroup {
   override fun getGlobalIjOptions() = GlobalIjOptions(OptionAccessScope.GLOBAL(null))
   override fun getEffectiveIjOptions(editor: VimEditor) = EffectiveIjOptions(OptionAccessScope.EFFECTIVE(editor))
 
-  private fun updateFallbackWindow(fallbackWindow: VimEditor, targetEditor: VimEditor) {
-    copyPerWindowGlobalValues(fallbackWindow, targetEditor)
-  }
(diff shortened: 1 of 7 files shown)


AlexPl292:
Thank you for the refactoring!

> This refactoring is the basis for the next PR, which will modify the option storage to allow custom options that are powered by IntelliJ settings (e.g. 'wrap', based on IntelliJ's soft wrap settings).

 Can I hear about your ideas on how this should work?

Also, I see a lot of changes around listeners. JFI, I think at some moment they should be moved from the old observer pattern to the Kotlin flows.

citizenmatt:
The changes around listeners here are only about separating responsibilities. I have no other plans for them right now, so don't see any issues with using flow in the future.

I'll have a PR ready very soon for `'wrap'` and hopefully a few other options. I started to describe it here, but it was becoming a description of the PR, so it's probably going to be easier just to pull together a draft PR as soon as possible - hopefully by tomorrow. I'm happy with the approach - this was always the end goal of all the options changes recently, and I think it's ended up both easy and intuitive 🤞 Will share ASAP.

PR #802, opened 2024-01-22
Add Matchit support for Java statements

chylex:
Adds matchit support for `if/else`, `do/while`, and `try/catch/finally` statements in Java.


AlexPl292:
Great, thank you!

chylex:
FYI the new unit tests don't test jumping `from try to catch to finally`, might be worth testing too. I can add the test if you want, but for you it might be faster to just add it than deal with another PR.

AlexPl292:
Thank you, added this one as well

PR #806, opened 2024-01-24
Enforce LF line separator in project code style

chylex:
The project code style defaults to "System-dependent" line separator, and as I develop on Windows it means IDEA is giving me this warning and wants me to convert every file to CRLF, which imo doesn't make any sense. So this PR sets the code style to LF to make developing on Windows less annoying.

image


AlexPl292:
I didn't know about this issue, thank you!

PR #819, opened 2024-02-05
Bug fixes for IdeaVim and Code With Me

citizenmatt:
This PR fixes a number of bugs in IdeaVim when being used in Code With Me scenarios, mainly when installed on the host, but also fixes issues when installed on both host and guest (JetBrains Client)

Primarily, it updates the API used to get the editors that are in use by the host, correctly ignoring the non-local, hidden editors used by remote guests, and ensures that any editor listeners are only working with local editors. It fixes the following issues:

- [x] Search highlights and selection from host transferred to the guest Client editors ([VIM-3268](https://youtrack.jetbrains.com/issue/VIM-3268))
- [x] File/tab selection callback was not ignoring guest editors, so "switch to alternate file" (`<C-^>`) on the host could switch to the last selected file on the guest ([VIM-3274](https://youtrack.jetbrains.com/issue/VIM-3274))
- [x] Line numbers would be drawn in guest editors, including relative line numbers - relative to the host's caret ([VIM-3275](https://youtrack.jetbrains.com/issue/VIM-3275))
- [x] Stop IdeaVim actions appearing in the Client. This would mean that that Client could disable IdeaVim for the host. If the Client also had IdeaVim installed, trying to disable IdeaVim on the Client would disable on the host ([VIM-3283](https://youtrack.jetbrains.com/issue/VIM-3283))
- [x] Fixes an unrelated bug that stopped the output pane from updating when changing the theme, due to an incorrect listener registration ([VIM-2167](https://youtrack.jetbrains.com/issue/VIM-2167))

There is a minor change in behaviour when a new editor is created. Previously, IdeaVim would incorrectly register listeners and initialise options, etc. for all windows. Now it only initialises editors that are supported by IdeaVim. This means local editors, but it's also dependent on the `'ideavimsupport'` option to decide if it should handle single line editors or editors inside dialogs. Previously, because the listeners were always registered, changing the `'ideavimsupport'` option would allow these editors to immediately start working. Now, they are only possibly initialised at creation time. If the option is changed, the editor would need to be closed and reopened. I do not expect this option to be changed during an interactive session, so I believe this is reasonable.

There are other issues outstanding for remote development scenarios, where IdeaVim is installed in the Client. This is currently blocked waiting a means to debug this scenario. I'll take a look when I get debugging working.


AlexPl292:
Cool, merged! CWM should be happy

PR #824, opened 2024-02-16
Fleet

lippfi:


vim-engine/src/main/kotlin/com/maddyhome/idea/vim/command/CommandBuilder.kt
```diff
@@ -17,11 +17,10 @@ import com.maddyhome.idea.vim.key.CommandPartNode
 import com.maddyhome.idea.vim.key.Node
 import com.maddyhome.idea.vim.key.RootNode
 import org.jetbrains.annotations.TestOnly
-import java.util.*
 import javax.swing.KeyStroke
 
-public class CommandBuilder(private var currentCommandPartNode: CommandPartNode<LazyVimCommand>) {
-  private val commandParts = ArrayDeque<Command>()
+public class CommandBuilder(private var currentCommandPartNode: CommandPartNode<LazyVimCommand>): Cloneable {
```
AlexPl292, line 22: https://stackoverflow.com/questions/26398951/why-is-cloneable-not-deprecated

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/KeyHandler.kt
```diff
@@ -729,3 +729,54 @@ public class KeyHandler {
...
+      private val logger = vimLogger<KeyProcessResult>()
+      private val lock = Object()
+    }
+
+    // TODO add concurrency to other places
+    public fun processKey(key: KeyStroke, editor: VimEditor, context: ExecutionContext, allowKeyMappings: Boolean, mappingCompleted: Boolean) {
+      synchronized(lock) {
+        val keyHandler = KeyHandler.getInstance()
+        if (keyHandler.keyHandlerState != originalState) {
+          logger.warn("Unexpected editor state. Aborting command execution.")
```
AlexPl292, line 408: Do we miss `return` here?

AlexPl292:
Review notes:
- Review each commit and add a description about WHY this change was made
- `clearMessage` from tests can be moved to the `tearDown` of the `VimTestCase`
- Refactor and split this huge commit: https://github.com/JetBrains/ideavim/pull/824/commits/8514fe1d22cfb59462c2399f3e70149b1306b0de. Here are some hints what can be extracted:
  - Making `handleKey` argument not-nullable
  - Wrapping `handleKey` with a `synchronized`
  - Step-by-step extraction of the key processors instead of extracting all processors in one commit
-Commits @lippfi
[Attempt to add annotation-processors as an artifact to vim-engine](https://github.com/JetBrains/ideavim/pull/824/commits/8f00cfcfce3de73b9c07036fd11882104e38df43) and [Avoid using annotation-processors in vim-engine](https://github.com/JetBrains/ideavim/pull/824/commits/c7c0c650fda1c91c5d0822a12af635f9efee4ef0) should be squashed or even removed from this branch. Is this related to this refactoring?
- Clean up `merge branch master` and `Fix merge` commits.


-----
Additional notes:
- Global state can be removed at all. This was an experiment, there is no need to handle it separatelly
- `VimStateMachine.getInstance()` can be inlined and removed. `Any` argument can be replaced with `VimEditor`.

PR #832, opened 2024-02-23
Only notify option change to initialised editors

citizenmatt:
This PR ensures that IdeaVim only works with editors that have been initialised. This is important for various callbacks, such as notifications when setting editor options. 

For example, the `~/.ideavimrc` file is evaluated during startup, before other editors are initialised. It can set various option values which in turn notify all affected editors. If the IDE opens editors at startup, these editors would be notified of changes before they've been initialised. [VIM-3256](https://youtrack.jetbrains.com/issue/VIM-3256) is an example of an issue caused by this - the uninitialised editors are notified, and try to check local-to-buffer options. These local-to-buffer options haven't yet been initialised, but trying to get a value will create the storage map for them, without setting any values. Subsequent attempts to use local-to-buffer options (such as `'keyword'`) can result in an assert.


AlexPl292:
Awesome, thank you!

PR #846, opened 2024-03-14
Fleet

lippfi:


vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimPsiService.kt
```diff
@@ -0,0 +1,36 @@
...
+ * Use of this source code is governed by an MIT-style
+ * license that can be found in the LICENSE.txt file or at
+ * https://opensource.org/licenses/MIT.
+ */
+
+package com.maddyhome.idea.vim.api
+
+import com.maddyhome.idea.vim.common.TextRange
+
+public interface VimPsiService {
```
AlexPl292, line 13: PSI is an intellij-specific term. As vim-engine is unbound from intellij, this would be better not to use such terms as well.
Inside of IntelliJ IDEA, use of PSI is a more implementation-specific thing. Looking at the methods of this class, it seems that it answers the questions about the language syntax, so probably `VimLanguageSyntaxService`?

AlexPl292 (approved):
A lot of functions in the SearchGroup have no usages in the code. The reason for that is actually stored in the master commit 06e1af37.
The combination of unused code + "revert of revert" significantly slows down the speed of the review as I have to spend more time understanding what changed and the reason behind the change.

In addition, the unused code by itself also leads to the slow down of the review as I have to review this unnecessary code plus understand why there are no usages. And since we have a [code quality tool](https://ideavim.teamcity.com/buildConfiguration/Ideavim_Qodana), this will cause extra unnecessary reports from it.

Please handle the comment about the `VimPsiService` naming in future pull requests.

PR #847, opened 2024-03-15
[VIM-1503]Adding option to "fix" the usability problem involving international keyboards and dead keys.

bruno-ortiz:
Hi! First time contributing here. Please tell me if something could have been done better and I'll try to fix it.

Trying to fix this error: https://youtrack.jetbrains.com/issue/VIM-1503

The best that I could think of was to create an option for those that use this types of keyboard.

### DIFF
--- a/src/test/java/org/jetbrains/plugins/ideavim/ex/implementation/commands/SetCommandTest.kt
+++ b/src/test/java/org/jetbrains/plugins/ideavim/ex/implementation/commands/SetCommandTest.kt
@@ -164,10 +164,11 @@ class SetCommandTest : VimTestCase() {
     assertCommandOutput("set all",
       """
         |--- Options ---
-        |noargtextobj        noincsearch           selectmode=       notextobj-indent
-        |nocommentary        nomatchit             shellcmdflag=-x     timeout
-        |nodigraph             maxmapdepth=20      shellxescape=@      timeoutlen=1000
-        |noexchange            more                shellxquote={     notrackactionids
+        |noargtextobj        noignorecase          scrolloff=0       notextobj-entire
+        |nocommentary        noincsearch           selectmode=       notextobj-indent
+        |nodigraph           nomatchit             shellcmdflag=-x     timeout
+        |noexchange            maxmapdepth=20      shellxescape=@      timeoutlen=1000
+        |nofixdeadkeys         more                shellxquote={     notrackactionids
         |nogdefault          nomultiple-cursors    showcmd             undolevels=1000
         |nohighlightedyank   noNERDTree            showmode            virtualedit=
         |  history=50          nrformats=hex       sidescroll=0      novisualbell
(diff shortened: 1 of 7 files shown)


citizenmatt:
Thanks for the PR! I'm not entirely sure what it's doing - could you give some scenarios (inserting text, motions, etc.) and how it fixes the problems, please?

bruno-ortiz:
Hey @citizenmatt, sure! I'll try to explain:
**Assume that in this examples the cursor is represented as  the '*' char.**
Whats happens without the PR:
```kotlin
val test = "Lorem ipsum do*lor sit amet"
// For some reason, if I try to select the entire sentence using the 'vi"' command, 
// it erases the character under the cursor and then selects the sentence. 
// In the example above, the letter 'o' will be erased.
val test = "*Lorem ipsum dlor sit amet*"
```
The code in this PR, tries to keep track of what "text" is under the cursor in the moment that an action is being executed. Thats what the `preExecute` method does in the **EditorActionHandlerBase** class.
And then I created a parent class for all textObject motions, that overrides the `preExecute` method to reinsert the text where it was previously deleted.
Example with the PR:
```kotlin
val test = "Lorem ipsum do*lor sit amet"
// Now when I try to select the entire sentence using the 'vi"' command, 
// it selects the entire sentence correctly.
val test = "*Lorem ipsum dolor sit amet*"
```

citizenmatt:
Hi Bruno. Apologies for the silence - there's not much bandwidth to look at stuff right now. Thanks for the contribution, but I must admit that I have a few reservations with this approach. I tried it locally, and after enabling `:set fixdeadkeys`, I can't get `vi"` to work directly. I still need to type something like `vi"<Space>` - the `"` is still highlighted as a composing character until I do something to finish it, such as `<Space>` or `<Escape>`. However, once I press the extra key, it does work as expected.

My main issue is that this is a workaround for a specific problem, rather than fixing the root cause. So it means that `vi"` can work (with additional keypress) and it doesn't replace the text, and this is great. However, it doesn't help with something like `vf"`, or other dead keys from other keymaps (`<A-N>`) which will still overwrite the text. Unfortunately, I think we need to go a bit more low level in Swing to handle this properly, working with input methods to try and prevent the editor from drawing the composing character when in normal or visual mode.

This approach would require further investigation though, possibly using `EditorComponent.enableInputMethods` when the mode changes, or adding a custom `IdeEventQueue` event dispatcher that consumes input method events based on target and mode, but I'm not sure on the implications of these appraoches. 

There are simpler workarounds, though, such as using a different keymap, if possible. The ABC keymap on the Mac doesn't have dead keys for the quote symbols, and something like [Ukulele](https://software.sil.org/ukelele/) can create new keymaps that don't have any dead keys.

Thanks again for the contribution. I'll leave it up to @AlexPl292 to decide what to do with this PR as it stands. It might be worth taking on as a workaround until someone can take a look at other Swing based implementations (invitation is open to all! 😁)

bruno-ortiz:
Hey @citizenmatt thanks for the reply! I understand that this is a "hacky" solution, and you guys may have reservations against merging this PR(no hard feelings). About the need to press the extra key <Space> or <Esc>, unfortunately this is how it works for the international keyboard, even in Vim or NeoVim I have to issue the command vi"<Space>, so I think that this is not a problem.
About the other commands, maybe we could tackle in other PRs?

AlexPl292:
Hi there. Sorry, but I can't accept this PR. In addition to Matts response, I don't like adding new options for fixes like that. The problem is almost no one knows about such options, and they contribute to combinations under which IdeaVim may work, so it starts to be more complicated to make sure everything is properly tested.

I agree that pressing "space" after `"` is good behavior, so if it will be implemented at some lower level, this would be the solution we'd like to have.

In IJ, the initial `"` replaces everything that was selected before, which causes the problem.
I've checked other approaches:
- Vim doesn't print `", `so it doesn't replace anything.
- Obsidian seems to select everything right on `"` press, so it doesn't wait for the "space".

Not sure how this should be fixed, but I think we need a more common approach.


Thank you for your interest in the question, though!

PR #870, opened 2024-05-05
Add NERDTree action to toggle excluded files

chylex:
NERDTree uses 'I' to toggle hidden files ([source](https://github.com/preservim/nerdtree/blob/6acfc48d80f83b7c23c4e6f9faa93c3defdb150b/doc/NERDTree.txt#L286)), which I think maps nicely to excluded files in IntelliJ, so I added support for it.


AlexPl292:
Hi, yep, looks smooth. Still, I'd name the action `NERDTreeMapToggleHidden` to be consistent with the NERDTree.

chylex:
Yep makes sense, done.

AlexPl292:
Merged, thank you!

PR #871, opened 2024-05-07
Fix clearing highlighted yank in editors within modal dialogs

citizenmatt:
This PR fixes a modality issue with the timer to clear the highlighted yank extension's highlights. If the editor was hosted inside a modal dialog (e.g. Live Templates in Settings), or shown from a modal dialog (e.g. Resolve Conflicts dialog), then the timer callback would wait for the modal dialog to close before clearing the highlights. This fixes [VIM-2236](https://youtrack.jetbrains.com/issue/VIM-2236/Highlighted-yank-extension-doesnt-automatically-clear-highlight-when-hosted-in-settings-dialog).

These changes also fix some other minor issues with the highlighted yank extension:

* A memory leak caused by registering a disposable against a project service to clean up highlights for every single yank.
* Config variables such as `g:highlightedyank_highlight_duration` can be a number as well as a string, matching the original Vim plugin.
* Highlights are cleared when IdeaVim is disabled
* Removed unnecessary Java version of the test. It's not language specific


AlexPl292:
Looks good to me, thank you for the update!

PR #879, opened 2024-05-09
Migrate to IntelliJ Platform Gradle Plugin 2.0

citizenmatt:
This PR migrates to the new IntelliJ Platform Gradle Plugin 2.0 (beta2). This is necessary to target 242, and also allows us to use "split mode" to reproduce and debug remote development issues. It's also easier to configure, and makes it easy to run the plugin in a different IDE, even locally built or installed IDEs.

It is still a work in progress, and not yet ready to merge. I will be actively updating and force-pushing changes into this branch.

The following build features work as expected:
- [x] Build, run and debug the plugin, including loading the `AceJump` (optional) dependency
- [x] Build, run and debug `:test`, `vim-engine:test`, and `java-tests:test`
- [x] Separately run `long-running-tests` and `property-tests` (there are failures, but these match `master`)
- [x] Plugin verifier runs with a default IDE based on the configured settings (using the since-build value, so it's testing 241). The task name has changed from `runPluginVerifier` to `verifyPlugin`.
- [x] Updated "IdeaVim full verification" run configuration to use new `verifyPlugin` task name
- [x] Updated TC config files to use new `verifyPlugin` task name
- [x] Updated GitHub actions to use new `testIdeUi` task instead of previous `runIdeForUiTests` task name
- [x] The source jar file is built and correctly copied to the `lib/src` folder. 
    However, the task has been renamed from `createOpenApiSourcesJar` to `sourcesJar`, and the filename is now `IdeaVIM-{version}-sources.jar` instead of `IdeaVIM-{version}-src.jar`. This reflects the behaviour of the task set up with the Gradle `java.withSourcesJar()` helper in the `vim-engine` module. The file is still created in the `lib/src` folder, and as [the documentation states](https://plugins.jetbrains.com/docs/intellij/bundling-plugin-openapi-sources.html), there are no strict naming rules for the files in this directory. This task does not appear to be called externally, and the `buildPlugin` task's dependency has been updated to reflect the new name.
- [x] A new `runIdeSplitMode` task and equivalent "Start IJ with IdeaVim (Split Mode)" run configuration have been added, to run the IDE in split mode, with the plugin loaded in the frontend, as per remote dev/CWM.
    Split mode requires 242+ and the project is currently targeting 241. This task runs the current EAP, specified by version in `gradle.properties`. This version number will be updated for each EAP until 2024.2 is released. Once IdeaVim targets 2024.2, we can remove this version and let the `runIdeSplitMode` run with the version it is compiled against.
- [x] Remove the `since-build` value from `plugin.xml`. It is now handled by the Gradle plugin, and is now in the format `{branch}.{build}`. We would previously specify the third minor component, but this is not necessary (it might be different between IDEs).
- [x] Remove the `since-build` value from `build.gradle.kts`. The `patchPluginXml` uses the default value from the target IDE dependency.
- [x] The `until-build` value is not written to `plugin.xml`, as before.
- [x] UI tests aren't currently running. The download robot task is commented out, as it should be moved to a normal dependency. The UI test task is not yet supported by the gradle plugin.

Other changes:
- [x] `settings.gradle` has been migrated to `settings.gradle.kts`
- [x] The Gradle plugin writes various files to `.intellijPlatform/` (Ivy XML files to reference bundled plugins, etc.) This folder is added to `.gitignore`
- [x] The CodeQL autobuild GitHub action needs a hint to choose the appropriate JVM version; it can't resolve the `by project` property IdeaVim uses to specify the `sourceCompatibility` version in `gradle.properties`. 
    By default it will choose the lowest version supported by the used version of Gradle, which means version 11. We add a comment in `build.gradle.kts` as a hint to help it choose the right version. This is not a new problem, but only causes an issue now because the new plugin has dependencies that require version 17, and won't resolve otherwise. We should perhaps look at whether we should just hard code the value in `build.gradle.kts` instead of specifying it in `gradle.properties` and then repeating it in the hint comment, or we replace the autobuild action with an action that explicitly specifies the version.
- [x] The `build.gradle.kts` file has commented out `runIdeCustom` and `runIdeLocal` that can be edited and used to run the plugin in a different IDE, or to run the plugin in a locally installed IDE.

Misc (not blockers):
* Remove temporary build feature to disable "binary releases" because the DevKit plugin (on the IdeaVim developer's machine) cannot currently download sources for binary releases. Will be removed once DevKit has been updated in 241 + 242


build.gradle.kts
```diff
@@ -181,21 +197,29 @@ tasks {
     }
   }
 
-  downloadRobotServerPlugin {
-    version.set(remoteRobotVersion)
-  }
+  // TODO: Robot server task not yet available
+//  downloadRobotServerPlugin {
+//    version.set(remoteRobotVersion)
+//  }
```
citizenmatt, line 203: What's supposed to happen with this task?

build.gradle.kts
```diff
@@ -181,21 +197,29 @@ tasks {
...
-  }
+  // TODO: Use testIdeUi task
+//  runIdeForUiTests {
+//    systemProperty("robot-server.port", "8082")
+//    systemProperty("ide.mac.message.dialogs.as.sheets", "false")
+//    systemProperty("jb.privacy.policy.text", "<!--999.999-->")
+//    systemProperty("jb.consents.confirmation.enabled", "false")
+//    systemProperty("ide.show.tips.on.startup.default.value", "false")
+//    systemProperty("octopus.handler", System.getProperty("octopus.handler") ?: true)
+//  }
```
citizenmatt, line 213: Also this task? I don't know how the current UI tests work...

build.gradle.kts
```diff
@@ -221,36 +245,44 @@ gradle.projectsEvaluated {
...
-  intellijRepository.set("https://www.jetbrains.com/intellij-repository")
-  plugins.set(listOf("AceJump:3.8.11"))
-}
+    ideaVersion {
+      // Set the since-build value, but leave until-build open ended (default is MAJOR.*)
+      // Don't forget to update plugin.xml
+      // TODO: Do we need this to be here *and* in plugin.xml?
+      sinceBuild.set("233.11799.67")
+      untilBuild.set(provider { null })
+    }
```
citizenmatt, line 261: If we set `since-build` here, will the Plugin DevKit use it for analysis? We have it duplicated in `plugin.xml` right now, it would be nice to only set in one place

build.gradle.kts
```diff
@@ -274,16 +306,17 @@ tasks {
     from(sourceSets.main.get().java) {
       include("**/com/maddyhome/idea/vim/**/*.java")
     }
-    from(project(":vim-engine").sourceSets.main.get().java) {
-      include("**/com/maddyhome/idea/vim/**/*.java")
-    }
+    // TODO: This causes errors with sourceSets not be available for this project?
+//    from(project(":vim-engine").sourceSets.main.get().java) {
+//      include("**/com/maddyhome/idea/vim/**/*.java")
+//    }
```
citizenmatt, line 312: HELP! I have no idea why the build breaks when this is uncommented. The `vim-engine` module hasn't been modified at all, and this used to work with the old gradle plugin. I don't know what could be wrong.

build.gradle.kts
```diff
@@ -109,6 +108,23 @@ dependencies {
...
+    // E.g. local("/Users/{user}/Applications/IntelliJ IDEA Ultimate.app")
+    // Or something like: intellijIdeaUltimate(ideaVersion)
+    create(ideaType, ideaVersion)
+
+    pluginVerifier()
+    zipSigner()
+    instrumentationTools()
+
+    testFramework(TestFrameworkType.Platform.JUnit4)
+    testFramework(TestFrameworkType.Platform.JUnit5)
```
citizenmatt, line 122: It seems unexpected to require two frameworks, but it fails with only one, and runs with two, so...

build.gradle.kts
hsz: WIP, tasks will be available soon.

build.gradle.kts
hsz: You can set it in Gradle build file only. Those values are used by the `patchPluginXml` file to update the `plugin.xml` file. All further checks performed by the Gradle plugin respect those values.

build.gradle.kts
hsz: This `project(":vim-engine").sourceSets.main.get().java` is basically wrong approach. I'll take a look at this soon.

build.gradle.kts
hsz: Fails how?

build.gradle.kts
citizenmatt: What about the DevKit though? Does it look at the patched or unpatched plugin.xml?

build.gradle.kts
citizenmatt: Either unresolved items (the `@RunInEdt` annotation) or runtime failures, also about the wrong threading models. Is it expected to require two frameworks?

hsz: It's unexpected. I'll take a look.

build.gradle.kts
citizenmatt: After some tests, it seems that DevKit only uses the value in the unpatched `plugin.xml`, so it's still worth keeping both values, even though it would make things easier to specify only in the build scripts

AlexPl292:
As a not, we HAVE to use 2.0 to target 242: https://plugins.jetbrains.com/docs/intellij/tools-gradle-intellij-plugin.html#usage

build.gradle.kts
citizenmatt: Properly implemented. Thank Jakub!

build.gradle.kts
citizenmatt: Current state: IdeaVim uses JUnit 5 features, so requires the `JUnit5` test framework type. However, some standard test functionality, such as `EditorTestUtil` and `LightProjectDescriptor` are only defined in `JUnit4`, so it looks like both test frameworks are required, and any change would need to come from the platform.

citizenmatt: As above: Gradle plugin has renamed `TestFrameworkType.Platform.JUnit4` to `TestFrameworkType.Platform`, and we add `TestFrameworkType.JUnit5` as an additional dependency.

build.gradle.kts
citizenmatt: Resolved: Let the Gradle plugin handle patching the `since-build` value. This value now comes solely from the target IDE dependency. It is no longer specified in `plugin.xml` or `build.gradle.kts`. The format is slightly different - `{branch}.{build}` instead of `{branch}.{build}.{minor}`. This is the default for the Gradle plugin; the minor version can be different between IDEs released from the same branch.

AlexPl292:
That's a big refactoring! Thank you!

PR #883, opened 2024-05-12
Various search related fixes

citizenmatt:
This PR fixes various search related issues.

**Search highlights:**
- [x] Fix `:set hlsearch` not re-enabling highlights after `:nohlsearch` ([VIM-3257](https://youtrack.jetbrains.com/issue/VIM-3257))
- [x] Fix search highlights not visible in diff view. Highlights are now shown in all visible editors, including diff and tool windows, and an editor's highlights are updated when the editor gains focus  ([VIM-2174](https://youtrack.jetbrains.com/issue/VIM-2174))
- [x] Fix regression in new regex engine not highlighting current search match when confirming each substitution with the `:substitute/{pattern}/{string}/c` command
- [x] Fix `'incsearch'` and `'nohlsearch'` adding highlights to other windows. It should only highlight the current match in the current editor to match Vim's behaviour
- [x] Fix search highlight colour not updating when editor scheme changes
- [x] Fix incorrectly forcing `'ignorecase'` for search highlights, ignoring passed in case sensitivity atom ([VIM-3391](https://youtrack.jetbrains.com/issue/VIM-3391))
- [x] Fix search highlight that crosses newline not being removed
- [x] Fix infinite loop in old regex engine when highlighting the identifier atom (it would treat the end of line character as an identifier character!) ([VIM-2510](https://youtrack.jetbrains.com/issue/VIM-2510))

**Incremental search:**
- [x] Fix search command line input breaking due to incsearch trying to highlight outside the range of other visible windows. The range is now coerced to fit the editor being highlighted. Also added try/catch to incsearch to prevent other exceptions breaking ex entry ([VIM-3325](https://youtrack.jetbrains.com/issue/VIM-3325))
- [x] Fix incremental search navigating to the wrong offset for the current match ([VIM-2779](https://youtrack.jetbrains.com/issue/VIM-2779))

**Search behaviour:**
- [x] Add support for `[count]/` in search ([VIM-2836](https://youtrack.jetbrains.com/issue/VIM-2836))
- [x] Fix motion type when using search as a motion and providing offset or `e` flag. Default motion is exclusive. Providing the `e` flag changes it to inclusive, and an offset makes it linewise ([VIM-1940](https://youtrack.jetbrains.com/issue/VIM-1940))
- [x] Fix search for last search pattern with new offset - typing `//e` now works  ([VIM-2356](https://youtrack.jetbrains.com/issue/VIM-2356))
- [x] Fix incorrectly resetting last substitution string when it's used - e.g. `:%s/foo/~` would use the last substitution string, but also reset it to an empty string. It's now only reset when it's a value other than `~` ([VIM-3354](https://youtrack.jetbrains.com/issue/VIM-3354))
- [x] Fix regression with new regex engine not saving and reusing previously saved patterns ([VIM-3348](https://youtrack.jetbrains.com/issue/VIM-3348))
- [x] Fix regression with new regex to support `/e` flag to navigate to end of search result ([VIM-3344](https://youtrack.jetbrains.com/issue/VIM-3344))
- [x] Fix regression with new regex when substitute command string ends with backslash ([VIM-3428](https://youtrack.jetbrains.com/issue/VIM-3428))
- [x] Fix bug in old regex engine that would skip forward too far with multiline patterns and miss matches. Also fixes bug in new regex to match newline at end of file ([VIM-698](https://youtrack.jetbrains.com/issue/VIM-698))
- [x] Fix bug in old regex engine to handle end of file atom (`\%$`) ([VIM-2888](https://youtrack.jetbrains.com/issue/VIM-2888))
- [x] Fix both regex engines for multiline regex substitution ([VIM-2141](https://youtrack.jetbrains.com/issue/VIM-2141))
- [x] Fix selection being incorrectly reset when searching in Visual mode. It should be updated based on search results
- [x] Fix highlighting of current match in substitute command not scrolling a result on the last line of the file to be visible above the ex entry field ([VIM-1560](https://youtrack.jetbrains.com/issue/VIM-2141))
- [x] Fix highlighting of current match on last line being obscured by "transparent" scrollbars on Windows + Linux ([VIM-3028](https://youtrack.jetbrains.com/issue/VIM-3028))

**Misc:**
- [x] Add status bar message when search wraps. Updated status bar message handling to clear on "redraw", like Vim - clear on scroll, add/remove lines, showing ex entry, etc. ([VIM-1043](https://youtrack.jetbrains.com/issue/VIM-1043))
- [x] Fix `/` or `:` commands not registering if ex entry is cancelled by losing focus ([VIM-3293](https://youtrack.jetbrains.com/issue/VIM-3293))


AlexPl292:
Thank you for all the great updates and fixes!

I have only one question about the first commit. Are you sure that using `overrideOption` is the correct approach for fixing this issue? Also, returning always `true` from the `setGlobalValue` seems to break the contract that says that "Returns `true` if the applied new value is different to the current stored value", isn't it?

AlexPl292:
Wow, I didn't know about `autocmd OptionSet` either.
Well, the `autocmd` ticket is 4th from the top, so we're not far away from it ;)

PR #884, opened 2024-05-15
Fix indent size calculation in "vin indent object" extention

Aisper:
Previously it ignored the first indent character, which caused errors with tabs. You wouldn't get any errors in behavior with spaces as you'd typically have 4 spaces at each indent level, so 1 in calculation doesn't make a difference.

Can be tested on following text. Easier with vii
```
first level selects all
{
	second level with tab. will still select all before fix
	{
		third level with tab. Will select also the second level before fix
	}
}

first level selects all
{
    second level with spaces. Will work before fix
    {
        third level with spaces. will only select that line
    }
}
```


AlexPl292:
Looks good to me, thank you!

PR #893, opened 2024-05-24
Moving away from Swing TextAction to KeyHandler

lippfi:


vim-engine/src/main/kotlin/com/maddyhome/idea/vim/handler/MotionActionHandler.kt
```diff
@@ -83,6 +83,14 @@ public sealed class MotionActionHandler : EditorActionHandlerBase(false) {
     ): Motion
   }
 
+  /**
+   * Support for commands that can be executed either once or for each caret depending on some circumstances
+   * TODO this class should not exist at all, changes to command execution are required
+   */
+  public abstract class AmbiguousExecution : MotionActionHandler() {
```
AlexPl292, line 90: Well, I don't like the name, but I don' have a better suggestion.

If `AmbiguousExecution` is not supposed to override functions from `MotionActionHandler`, don't you want to `final override` them?

Also, I faced such issue before and made a bit different approach to this. See `ConditionalSingleExecution`.

src/main/java/com/maddyhome/idea/vim/ui/ex/ExEntryPanel.java
```diff
@@ -48,8 +50,8 @@
 /**
  * This is used to enter ex commands such as searches and "colon" commands
  */
-public class ExEntryPanel extends JPanel {
-  private static ExEntryPanel instance;
+public class ExEntryPanel extends JPanel implements VimCommandLine {
```
AlexPl292, line 53: Where are `@Override` on `deactivate`? And vals like `caret`, `label`?

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/action/ex/LeaveCommandLineAction.kt
```diff
@@ -1,35 +1,30 @@
...
- *
- * The mapping for this action means that the ex command is executed as a write action
- */
-@CommandOrMotion(keys = ["<CR>", "<C-M>", "<C-J>"], modes = [Mode.CMD_LINE])
-public class ProcessExEntryAction : VimActionHandler.SingleExecution() {
-  override val type: Command.Type = Command.Type.OTHER_SELF_SYNCHRONIZED
-
-  override val flags: EnumSet<CommandFlags> = EnumSet.of(CommandFlags.FLAG_COMPLETE_EX)
+@CommandOrMotion(keys = ["<Esc>", "<C-[>", "<C-C>"], modes = [Mode.CMD_LINE])
+public class LeaveCommandLineAction : VimActionHandler.SingleExecution() {
```
AlexPl292, line 23: How does this command work if it just returns `true`?

AlexPl292 (changes requested):
I like the changes and I'm ready to merge them.

However, there is a big issue I faced: When I pressed Ctrl-R in command line, the `Run` function of IJ has started. In Vim settings the handler was `Vim`, so it's a bug.

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/handler/MotionActionHandler.kt
lippfi: I don't like both `AmbiguousExecution` and `ConditionalSingleExecution` TBH. We should gather up and figure out how to avoid using such dirty hacks

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/action/ex/LeaveCommandLineAction.kt
lippfi: This command has `CommandFlags.FLAG_END_EX` which is handled in command consumer (it closes command line, resets commandBuilder etc) 
I don't like idea of having this flag or this logic in `CommandConsumer`, however it's not the only command with `CommandFlags.FLAG_END_EX` and some of such commands are `ForEachCaret`. This "end ex" code should be executed only once and can't be moved to commands body (at least before we add some `preExecute` method for commands that will be executed once`)

PR #894, opened 2024-05-28
Incsearch support for :global

citizenmatt:
Well this PR is a great example of yak shaving. I wanted to modify one line in the ex text field change listener to add `instanceof GlobalCommand` so it would update highlights for the `:global` command, and, well, this PR happened instead. The big problem is that `:global` has a default range of the whole file, where (nearly) all other commands have a default range of the current line, and that wasn't easily added.

So, this PR refactors range handling to allow the `:global` command to have a different default range to the other commands. It improves and simplifies the API and uses the same terminology as Vim - a "range" is a collection of "addresses". An address evaluates to a line number, and can be an explicit number, such as `1` or `2`, or a string like `.` for the current line, `$` for the last line, `'a` for a mark, etc., or even a search pattern. Addresses can also have +/- offsets. The range `%` is a collection of the first line to the last line in a file.

It also fixes a number of off-by-one errors; addresses are one-based, while IdeaVim internally is zero-based. This is particularly important because the address `0` is special. For the `:[range]move {address}` or `:[range]copy {address}` command, a value of `1` for `{address}` means move/copy to the line _after_ the first line, while `0` means move/copy to the line _before_ the first line. This `{address}` should not be confused with `{count}` in `:[range]yank {count}`, which is also 1-based, but should be treated as 1-based by IdeaVim.

As part of the refactoring, it fixes the following issues:
* Fixes off-by-one error in `:[count]goto` to go to the `[count]`th character in a file
* Fixes `:[range]copy 0` to correctly copy to the line _before_ the first line
* Fixes `:>{count}` and `:<{count}`
* Fixes correct end location of caret after various ex operations
* Fixes maintaining correct mode (i.e. Visual) when cancelling search entry
* Fixes clearing the output panel after `:global` command ([VIM-2570](https://youtrack.jetbrains.com/issue/VIM-2570))
* Fixes range containing `0` not being handled correctly, e.g. `:0,$s/...` ([VIM-1137](https://youtrack.jetbrains.com/issue/VIM-1137))
* Fixes range with missing last address, e.g. `:1,d` ([VIM-992](https://youtrack.jetbrains.com/issue/VIM-992))
* Fixes incorrect value returned from `line()` script function when in Normal mode
* Fixes `:[range]m.` move to current line ([VIM-2936](https://youtrack.jetbrains.com/issue/VIM-2936))
* Reports correct Vim errors for invalid range, mark not set, etc.
* Add more tests and minor fixes for range, count, address and register in commands such as `:copy`, `:move`, `:goto` (goto character), `:[range]` (goto line), `:delete`, `:join`, `:yank`

Finally, of course, it adds `'incsearch'` highlighting for the `:global` command ([VIM-2891](https://youtrack.jetbrains.com/issue/VIM-2891)).


citizenmatt:
Rebased and force pushed. There were some conflicts with recent changes to handling the ex text field, so [the last 7 commits should be reviewed again](https://github.com/JetBrains/ideavim/compare/4e77d8f4b058958ee09e24c077217891ee08e107...fa18d235fa33f2c5c4724d8ae0cc423e13e67231). The conflicts highlighted a bug and a regression, which are also fixed in these last commits.

PR #897, opened 2024-06-02
Fleet

lippfi:
One more set of changes for Fleet


src/main/java/com/maddyhome/idea/vim/ui/ex/ExEntryPanelService.kt
```diff
@@ -95,6 +95,18 @@ public class ExEntryPanelService : VimCommandLineService {
     return panel
   }
 
+  public override fun createWithoutShortcuts(
```
AlexPl292: `createSomething` usually used when we actually create a new instance of the object.
Here we reuse initialized instance of the ExEntryPanel, so I'd say that this method should be named also `getInstance....`

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimCommandLine.kt
```diff
@@ -11,6 +11,8 @@ package com.maddyhome.idea.vim.api
 import javax.swing.KeyStroke
 
 public interface VimCommandLine {
+  public val caret: VimCommandLineCaret
```
AlexPl292, line 15: I don't understand why this change is included in this commit.

src/main/java/com/maddyhome/idea/vim/ui/ex/ExTextField.java
```diff
@@ -209,11 +207,11 @@ private void setFontToJField(String stringToDisplay) {
 
   void setEditor(@NotNull Editor editor, DataContext context) {
     this.context = context;
-    EditorHolderService.getInstance().setEditor(editor);
+    ExEntryPanel.getInstance().setEditor(editor);
   }
 
   public Editor getEditor() {
```
AlexPl292, line 213: This requires `@Nullable` on return type now

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimSearchGroupBase.kt
```diff
@@ -48,23 +48,35 @@ public abstract class VimSearchGroupBase : VimSearchGroup {
...
+      @JvmStatic
+      get
+      @JvmStatic
+      set
 
     /**
      * The type of the last used pattern.
      */
-    private var lastPatternType: PatternType? = null
+    @JvmStatic
```
AlexPl292, line 66: Why do we need all new `@JvmStatic`? I see that their usage was moved from java class to kotlin class.

AlexPl292 (changes requested):

src/main/java/com/maddyhome/idea/vim/ui/ex/ExEntryPanelService.kt
lippfi: In IdeaVim, we reuse the initialized instance, but this is unlikely to be the case for Fleet. The implementation details of IdeaVim should not influence naming of the new API
Previously, we used `getInstanceWithoutShortcuts()` and then called `activate()` on it. Now, however, we have adopted a more intuitive approach by directly calling `createWithoutShortcuts()`

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimCommandLine.kt
lippfi: I don't have any objective reason for this; I just felt that this field would look better at the top of the class and didn't want to create a separate commit for such a minor change

PR #900, opened 2024-06-10
Fix various options and ex text field issues

citizenmatt:
This PR fixes various issues.

### Options:

- [x] The `'fileencoding'` option should not be copied during initialisation, but not all scenarios were handled. Setting the option will convert the file immediately (unlike Vim which waits until the file is saved. IntelliJ's saves are less deterministic). If the option is set in `~/.ideavimrc`, the value can be applied to new files being opened, and can result in a disruptive dialog prompting to convert the file. [VIM-3467](https://youtrack.jetbrains.com/issue/VIM-3467).
- [x] Fix soft wrap inconsistency between the main editor and the console output in the Run tool window. This is because IntelliJ has multiple soft wrap settings, for main editors, consoles, previews and others. IdeaVim will no longer copy options between different kinds of editor. IdeaVim will also look at the correct soft wrap type to check if the local value differs to the global value. [VIM-3450](https://youtrack.jetbrains.com/issue/VIM-3450).
- [x] Fixes an infinite recursion initialising editors that create a dummy inner editor while downloading the real file. When initialising options after an editor is created, IdeaVim checks for the currently selected editor, which could cause the dummy editor to be created, which would need options to be initialised… This recursion doesn't appear to happen any more in 241 (coroutines!), but the protection is still sensible. [VIM-3066](https://youtrack.jetbrains.com/issue/VIM-3066).
- [x] Fix a regression in the new regex engine causing `:global` to not honour `'ignorecase'` and `'smartcase'` options

### Ex text field

- [x] Fix text field not handling IDE zoom, or editor zoom.  [VIM-3417](https://youtrack.jetbrains.com/issue/VIM-3417).
- [x] Fix regression in ex output panel clearing text instead of closing when clicking outside of panel
- [x] Fix regression with missing newlines in `:print` output. The ex text field would strip trailing newlines, but not add them when concatenating. The test helper would not strip the newlines, so tests would pass. The `ExOutputModel.text` property will now strip trailing new lines, so tests match the UI behaviour. This means all asserts no longer require a trailing newline, so lots of changes for that, too

Also:
- [x] Fix incorrectly automatically switching to Insert mode for diff windows. IdeaVim will switch to Insert mode for some editors, e.g. consoles, git commit editor, and other editors that are not backed by a file. This change will keep IdeaVim in Normal mode for diff windows that do not have a disk based file. [VIM-3442](https://youtrack.jetbrains.com/issue/VIM-3442).

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/ex/ExOutputModel.kt
+++ b/src/main/java/com/maddyhome/idea/vim/ex/ExOutputModel.kt
@@ -28,14 +28,18 @@ public class ExOutputModel private constructor(private val myEditor: Editor) : V
     get() = if (!ApplicationManager.getApplication().isUnitTestMode) {
       ExOutputPanel.getInstance(myEditor).text
     } else {
-      field
+      // ExOutputPanel always returns a non-null string
+      field ?: ""
     }
     set(value) {
+      // ExOutputPanel will strip a trailing newline. We'll do it now so that tests have the same behaviour. We also
+      // never pass null to ExOutputPanel, but we do store it for tests, so we know if we're active or not
+      val newValue = value?.removeSuffix("\n")
       if (!ApplicationManager.getApplication().isUnitTestMode) {
-        ExOutputPanel.getInstance(myEditor).setText(value ?: "")
+        ExOutputPanel.getInstance(myEditor).setText(newValue ?: "")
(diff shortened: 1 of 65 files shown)


AlexPl292:
Awesome, thank you!

PR #904, opened 2024-06-14
Fix jumplist no longer updated from IDE actions

citizenmatt:
This PR fixes an issue where the jumplist was no longer updated when a new item to the recent places menu (e.g. via `:action GotoImplementation` or moving the caret with the mouse). It also implements `:clearjumps` and fixes an issue where a jumplist would be shared between two open projects with the same name.


lippfi:
Looks good for me, thank you for this PR!
And one more thank you for excluding generated files in `.gitignore`

PR #906, opened 2024-06-17
Add ability to change all handlers to use either Vim or IDE shortcuts.

LazyScaper:
Adds the ability to change all handlers to use either Vim or IDE shortcuts.

This is particularly useful in a new instance where you do not use a .ideavimrc file.


src/main/java/com/maddyhome/idea/vim/ui/VimEmulationConfigurable.kt
```diff
@@ -328,6 +329,38 @@ internal class VimEmulationConfigurable : Configurable {
...
+  ) : DumbAwareAction(
+    "Toggle all Handlers",
+    "Toggle all action handlers to use either IDE or Vim shortcuts",
+    AllIcons.Actions.ChangeView,
+  ) {
+    override fun update(e: AnActionEvent) {
+      e.presentation.isEnabled = true
+    }
+
+    override fun getActionUpdateThread(): ActionUpdateThread = ActionUpdateThread.EDT
```
AlexPl292, line 344: Let's switch to ActionUpdateThread.BGT

AlexPl292 (changes requested):
Hi there, that's a cool feature, I like it.

I think it is worth updating it a bit:

- Let's not only switch between "Vim" and "IDE" handlers, but rotate between "Undefined" -> "Vim" -> "IDE" -> "Undefined". I think this will be helpful in case the user clicks the button out of interest and gets scared that they "broke" everything.

- Also, we should probably update not all handlers, but only the ones that are set to "Undefined". 

What do you think about that?

src/main/java/com/maddyhome/idea/vim/ui/VimEmulationConfigurable.kt
LazyScaper: Thanks for the feedback, will do another passover tonight with the above in mind.

I like the idea of not updating all handlers, seems "safer" for the user.

LazyScaper: I have added the above funtionality.

AlexPl292:
Awesome, thank you for the feature!

PR #907, opened 2024-06-18
Fix glitch with 'scrolloff' mapping at bottom of file

citizenmatt:
This PR fixes a glitch with scrolling when `'scrolloff'` is greater than 0. Previously, the mapping of the `'scrolloff'` Vim option to IntelliJ's "vertical scroll offset" setting would update the (local, per-editor) IntelliJ setting to the same value. IntelliJ would use this when performing its own scrolling. However, IntelliJ's scrolling calculation includes the virtual space at the bottom of the file. So typing `G` would place the last line at the bottom of the screen, but then any caret movement would trigger IntelliJ's scrolling and move the last line up to include scroll offset/virtual space.

This was more noticeable with a large scroll offset, such as `999`, intended to keep the caret in the centre of the screen. After typing `G` to put the last line at the bottom of the screen, hitting backspace in Insert mode would trigger IntelliJ to update scrolling, and it would try to put the current line at the centre of the screen. IdeaVim would also update scrolling and reposition the line at the bottom of the screen, causing the editor to "twitch" as you type. Normally setting the scroll position multiple times would result in "last write wins" and there should be no glitch, but IntelliJ's backspace handling updates the scroll position with animation disabled, so the editor immediately (and visibly) scrolls, and is then immediately scrolled back. ([VIM-3474](https://youtrack.jetbrains.com/issue/VIM-3474))

This PR updates the `'scrolloff'` mapping so that the IntelliJ setting is always set to `0`. IntelliJ's scrolling no longer includes scroll offset, and IdeaVim's scrolling can then position it correctly. It also handles a potential issue with `'scrolljump'` being a negative value (Vim allows `-1` to `-100` as percentages), which IntelliJ does not validate.

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/group/OptionGroup.kt
+++ b/src/main/java/com/maddyhome/idea/vim/group/OptionGroup.kt
@@ -975,26 +975,35 @@ private class SideScrollOptionMapper(option: NumberOption, internalOptionValueAc
  * setting value, and there is no UI to modify the local IntelliJ settings. Once the value has been set in IdeaVim, it
  * takes precedence over the global, persistent setting until the option is reset with either `:set scrolloff&` or
  * `:setlocal scrolloff<`.
+ *
+ * Note that when the IdeaVim value is set, we set the IntelliJ local value to 0 rather than sharing the value. This is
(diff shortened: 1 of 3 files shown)


AlexPl292:
Great, thank you for the fix!

PR #908, opened 2024-06-18
Fix incorrect double registration when opening editors

citizenmatt:
Ensures that the editor listeners are only registered once when the editor is created. IdeaVim uses two different events to track editor creation, to try to understand how the editor is being created (new window, split, preview, etc.) and this can lead to calling the `EditorListeners.add` method twice. This would create a second `Disposable` and cause handlers to leak. This is most visible when creating the first window and then disabling IdeaVim. This window would still handle drag events like IdeaVim, even changing the caret back to block after the drag finishes.


AlexPl292:
Thank you for the fix. Is there some related issue in YouTrack?

citizenmatt:
Not that I'm aware of. It was just a bug that I hit while doing other stuff.

PR #913, opened 2024-06-20
add NERDTree action to delete files and directories

kun-codes:
I have used "d" to delete files and directories

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/extension/nerdtree/NerdTree.kt
+++ b/src/main/java/com/maddyhome/idea/vim/extension/nerdtree/NerdTree.kt
@@ -478,6 +478,7 @@ internal class NerdTree : VimExtension {
       NerdAction.ToIj("SynchronizeCurrentFile"),
     )
     registerCommand("NERDTreeMapToggleHidden", "I", NerdAction.ToIj("ProjectView.ShowExcludedFiles"))
+    registerCommand("NERDTreeMapDelete", "d", NerdAction.ToIj("\$Delete"))
     registerCommand("NERDTreeMapRefreshRoot", "R", NerdAction.ToIj("Synchronize"))
     registerCommand("NERDTreeMapMenu", "m", NerdAction.ToIj("ShowPopupMenu"))
     registerCommand("NERDTreeMapQuit", "q", NerdAction.ToIj("HideActiveWindow"))


AlexPl292:
Hi, I can't find this shortcut or command (`NERDTreeMapDelete`) in NERDTree plugin. Can you please share the reference to its documentation?

kun-codes:
Hello,

The closest thing which I could find in the documentation was [this](https://github.com/preservim/nerdtree/blob/6acfc48d80f83b7c23c4e6f9faa93c3defdb150b/doc/NERDTree.txt#L89). But a lot of neovim filetree plugins have a dedicated keybind to delete files and directories. For example:
- [lir.nvim](https://github.com/tamago324/lir.nvim/blob/5b1a927cfee45845d7d4450a1e8f2f63cf0baaac/README.md?plain=1#L46)
- [chadtree](https://github.com/ms-jpq/chadtree/blob/chad/docs/KEYBIND.md#chadtree_settingskeymapdelete)
- [neo-tree.nvim](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/206241e451c12f78969ff5ae53af45616ffc9b72/README.md?plain=1#L270)
- [drex.nvim](https://github.com/TheBlob42/drex.nvim/blob/acccc1225e61a3977d86a590420b868e708cc64a/README.md?plain=1#L114)
- [sfm.nvim](https://github.com/dinhhuy258/sfm.nvim/blob/0d020c68c617ceed386b10c9d4f548a9edc66cd8/README.md?plain=1#L160)

Even [Spacevim](https://spacevim.org/documentation/#file-tree-navigation) which is a vim configuration has dedicated keybinds for creating new files and folders .

This PR would make managing files via NERDTree much faster as the user won't have to use the right click menu to create new files or folders.

AlexPl292:
Okay, let's merge it :)
I've created a ticket for the reference: [VIM-3488](https://youtrack.jetbrains.com/issue/VIM-3488) New commands in NERDTree

kun-codes:
hello,

what is the ticket used for? Also I see [here](https://plugins.jetbrains.com/plugin/164-ideavim/versions/dev) that there are new dev versions of plugins every 2 days. Does that mean that this new feature would be available in the next dev release?

AlexPl292:
The ticket is just a reference. 

Yes, it'll be available on the next dev release.

PR #914, opened 2024-06-20
Update docs for multiple-cursors plugin

alexfu:
When I added the `terryma/vim-multiple-cursors` plugin, I struggled at first to get it working since `Ctrl+n` is mapped to something else by default; `Ctrl+n` is the default key bind for `terryma/vim-multiple-cursors`

This adds documentation on re-mapping to trigger multiple cursors and which command to use.

### DIFF
--- a/doc/IdeaVim Plugins.md
+++ b/doc/IdeaVim Plugins.md
@@ -129,8 +129,16 @@ Original plugin: [vim-multiple-cursors](https://github.com/terryma/vim-multiple-
       </details>
    
 ### Instructions
-   
+
 https://github.com/terryma/vim-multiple-cursors/blob/master/doc/multiple_cursors.txt
+
+For default behavior, add the following to your `.ideavimrc` file...
+
+```
+map <C-n> <Plug>NextOccurrence
+```
+
+This will map `Ctrl+n` to trigger multiple cursors.
 
 </details>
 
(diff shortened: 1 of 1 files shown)


citizenmatt:
The default keymaps for the IdeaVim extension don't map the Vim plugin's, for some reason. There is an issue for this: [VIM-2178](https://youtrack.jetbrains.com/issue/VIM-2178). If we're going to add the correct mapping to the doc, we should probably include all of the appropriate keymaps, as mentioned in [this comment](https://youtrack.jetbrains.com/issue/VIM-2178/multiple-cursors-defines-incorrect-keys#focus=Comments-27-4558015.0-0).

alexfu:
> The default keymaps for the IdeaVim extension don't map the Vim plugin's, for some reason. There is an issue for this: [VIM-2178](https://youtrack.jetbrains.com/issue/VIM-2178). If we're going to add the correct mapping to the doc, we should probably include all of the appropriate keymaps, as mentioned in [this comment](https://youtrack.jetbrains.com/issue/VIM-2178/multiple-cursors-defines-incorrect-keys#focus=Comments-27-4558015.0-0).

Oh, I didn't realize there was an issue for this. I can update my changes to reflect that.

alexfu:
@citizenmatt updated

AlexPl292:
I hope to deprecate existing mappings one day and switch to https://github.com/mg979/vim-visual-multi approach.

Anyway, let's have this information included.

PR #912, opened 2024-06-20
add NERDTree action to create new files and directories

kun-codes:
I have used "n" to create a new file and "N" to create a new directory

PR #918, opened 2024-06-24
add NERDTree action to copy, paste and rename files

kun-codes:
Similar functionality has been added in #912 and #913. [This](https://github.com/JetBrains/ideavim/pull/913#issuecomment-2182525822) was the reason given for them

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/extension/nerdtree/NerdTree.kt
+++ b/src/main/java/com/maddyhome/idea/vim/extension/nerdtree/NerdTree.kt
@@ -482,6 +482,7 @@ internal class NerdTree : VimExtension {
     registerCommand("NERDTreeMapNewDir", "N", NerdAction.ToIj("NewDir"))
     registerCommand("NERDTreeMapDelete", "d", NerdAction.ToIj("\$Delete"))
     registerCommand("NERDTreeMapCopy", "y", NerdAction.ToIj("\$Copy"))
+    registerCommand("NERDTreeMapPaste", "v", NerdAction.ToIj("\$Paste"))
     registerCommand("NERDTreeMapRefreshRoot", "R", NerdAction.ToIj("Synchronize"))
     registerCommand("NERDTreeMapMenu", "m", NerdAction.ToIj("ShowPopupMenu"))
     registerCommand("NERDTreeMapQuit", "q", NerdAction.ToIj("HideActiveWindow"))
--- a/src/main/java/com/maddyhome/idea/vim/extension/nerdtree/NerdTree.kt
+++ b/src/main/java/com/maddyhome/idea/vim/extension/nerdtree/NerdTree.kt
@@ -483,6 +483,7 @@ internal class NerdTree : VimExtension {
     registerCommand("NERDTreeMapDelete", "d", NerdAction.ToIj("\$Delete"))
     registerCommand("NERDTreeMapCopy", "y", NerdAction.ToIj("\$Copy"))
     registerCommand("NERDTreeMapPaste", "v", NerdAction.ToIj("\$Paste"))
+    registerCommand("NERDTreeMapRename", "<C-r>", NerdAction.ToIj("RenameElement"))
     registerCommand("NERDTreeMapRefreshRoot", "R", NerdAction.ToIj("Synchronize"))
     registerCommand("NERDTreeMapMenu", "m", NerdAction.ToIj("ShowPopupMenu"))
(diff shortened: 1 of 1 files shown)


AlexPl292:
Hi, I don't like the new commands because they're controversial in other implementations. There is no standard for them.
At the same time, I found an `m` command in NERDTree, which also implements some of the mentioned functionality. I think it would be better to hide these operations under `m` command.

kun-codes:
Hello,

for copying files, these filetree neovim plugins have a keybind

- [lir.nvim](https://github.com/tamago324/lir.nvim/blob/5b1a927cfee45845d7d4450a1e8f2f63cf0baaac/README.md?plain=1#L52)
- [chadtree](https://github.com/ms-jpq/chadtree/blob/chad/docs/KEYBIND.md#chadtree_settingskeymapcopy)
- [neo-tree.nvim](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/206241e451c12f78969ff5ae53af45616ffc9b72/README.md?plain=1#L275)
- [sfm.nvim](https://github.com/dinhhuy258/sfm.nvim/blob/0d020c68c617ceed386b10c9d4f548a9edc66cd8/README.md?plain=1#L158)

With [spacevim](https://spacevim.org/documentation/#file-tree-navigation) which is a distribution of vim config

for pasting files, the neovim plugins also have a dedicated keybind

- [lir.nvim](https://github.com/tamago324/lir.nvim/blob/5b1a927cfee45845d7d4450a1e8f2f63cf0baaac/README.md?plain=1#L54)
- [neo-tree.nvim](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/206241e451c12f78969ff5ae53af45616ffc9b72/README.md?plain=1#L274)
- [drex.nvim](https://github.com/TheBlob42/drex.nvim/blob/acccc1225e61a3977d86a590420b868e708cc64a/README.md?plain=1#L210)

Spacevim too implements a dedicated keybind for pasting files

and for renaming files
- [neo-tree.nvim](https://github.com/nvim-neo-tree/neo-tree.nvim/blob/206241e451c12f78969ff5ae53af45616ffc9b72/README.md?plain=1#L271)
- [lir.nvim](https://github.com/tamago324/lir.nvim/blob/5b1a927cfee45845d7d4450a1e8f2f63cf0baaac/README.md?plain=1#L42)
- [chadtree](https://github.com/ms-jpq/chadtree/blob/chad/docs/KEYBIND.md#chadtree_settingskeymaprename)
- [drex.nvim](https://github.com/TheBlob42/drex.nvim/blob/acccc1225e61a3977d86a590420b868e708cc64a/README.md?plain=1#L119)

Spacevim too has a a dedicated keybind for renaming files


Although I do agree that the keybinds which I suggested in my PR are non-conventional as the most appropriate keybinds in my opinion were already used.

kun-codes:
@AlexPl292 Sorry for mentioning you, but I wanted to ask if you are okay with the given keybindings or do you want to put these operations under a specified key. In your previous comment you said that the `m` key is okay for it, but the `m` is used for triggering the right click menu.

AlexPl292:
Okay, let's merge these changes, but I'd avoid further updates with the keybindings that are not used in the NERDTree plugin.
The problem is that NERDTree may introduce new commands with the same shortcuts and we'll have to decide if we'd like keep compatibility or support the NERDTree new feature.
If any further similar updates are required, we'll need to find a way that does not potentially conflict with NERDTree.

AlexPl292:
Reference: https://youtrack.jetbrains.com/issue/VIM-3539/More-NERDTree-commands

kun-codes:
Hello,

I understand what you are saying. If we keep on adding extra features to this plugin, we would soon have to decide if to add extra functions or support any new keybinds which nerdtree introduces.

I have an idea to deal with this. How about instead of adding these extra features and binding them to just one key like `a`, `A` or `z`, we select another secondary key which will act as an entry point to all these extra features. If for example we decide that the key is to be `;` then our keybinds for extra features would be `;a`, `;A` or `;z`

AlexPl292:
This actually sounds good to me. We'll need to check if the `;` is not already used, however :D

kun-codes:
You can select any key other than `;` for that. But it should be near the home row for easy usage

citizenmatt:
Perhaps we could provide the APIs that some of the NERDTree plugins are using so that these can become mini-plugins themselves?

citizenmatt:
That's pretty much what I mean. Instead of adding a feature that doesn't have a standard definition outside of core NERDTree functionality, we could allow IdeaVim's implementation to be extensible, ideally in much the same way that the original plugin is (whatever that is).

PR #923, opened 2024-06-27
Fix incsearch with result before caret and wrapscan

citizenmatt:
If all results are before the caret, make sure it's still possible to find the closest match.

Fixes [VIM-3505](https://youtrack.jetbrains.com/issue/VIM-3505)


citizenmatt:
Hang on. This is a bad fix and doesn't work with `'wrapscan'` disabled. Don't merge just yet!

citizenmatt:
Done. Force pushed with a better implementation, and based on current `master` rather than a surprisingly old commit

AlexPl292:
Awesome, thank you!

PR #925, opened 2024-07-03
Fix null reference initialising output panel fonts

citizenmatt:
Fixes an exception that would prevent debugging in Rider. The Debug tool window would create an `Editor` instance during initialisation and set the font size. If the font size was changed (e.g. the editor was zoomed), then IdeaVim's font property change listener would try to hide the ex command line and the output panel. Unfortunately, this forced creation of the output panel, which would try to position itself, but because the `Editor` instance hadn't been added to the UI, the editor component's root pane would be null, and lead to a `NullPointerException`.

This PR adds a null check to avoid the exception, and also avoids creating the panel just to make sure it's hidden.

Fixes [VIM-3515](https://youtrack.jetbrains.com/issue/VIM-3515)


AlexPl292:
Thank you for the fix!

PR #943, opened 2024-07-12
Show possible action IDs when tracking, based on current shortcut

citizenmatt:
Some actions are registered with a local component, and not with `ActionManager`, for example `NextTab` has a different implementation for editor tabs and tool window tabs. This means we can't use the action instance to look up an action ID from `ActionManager`, and the "track action IDs" feature can't show the executed action ID.

However, the action is a copy of the global "template" action, including shortcuts. We can use the shortcuts to look up and display all possible action IDs as candidates. The notification might therefore include multiple actions, but it is likely to include the required action.

Fixes [VIM-3499](https://youtrack.jetbrains.com/issue/VIM-3499).

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,13 +222,13 @@ Ex commands or via `:map` command mappings:
     * Execute an action by `{action_id}`. Works from Ex command line.
     * Please don't use `:action` in mappings. Use `<Action>` instead.
 
-### Finding action ids:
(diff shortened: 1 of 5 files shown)


AlexPl292:
This is great, thank you!

SerkanSipahi:
@citizenmatt thanks.

PR #951, opened 2024-07-23
Small refactorings around CommandBuilder

citizenmatt:
I've been looking at how extensions interact with `CommandBuilder` (spoiler: badly) and before fixing some of these extension related things, I've done some small refactorings in the following:

* Removing some unused functions + properties from `CommandBuilder`
* Removing some special case code in `CommandConsumer` that were based on action names
* Refactor `DigraphResult` implementation to use sealed classes
* Fixing some minor correctness issues with `CommandBuilder` not using the proper command part when trying to get the in-progress count or register

Pinging @lippfi as I know you're working with `CommandBuilder`. These changes should be more about implementation details than significant changes, so hopefully shouldn't cause conflicts.

### DIFF
--- a/src/test/java/org/jetbrains/plugins/ideavim/action/MotionActionTest.kt
+++ b/src/test/java/org/jetbrains/plugins/ideavim/action/MotionActionTest.kt
@@ -767,6 +767,11 @@ class MotionActionTest : VimTestCase() {
   }
 
   @Test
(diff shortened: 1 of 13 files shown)


lippfi:
Looks good to me
And the code became so much better without special cases handling
Thank you!

PR #954, opened 2024-07-25
Fleet

lippfi:


src/main/java/com/maddyhome/idea/vim/listener/VimInsertListener.kt
```diff
@@ -1,15 +0,0 @@
...
- * Use of this source code is governed by an MIT-style
- * license that can be found in the LICENSE.txt file or at
- * https://opensource.org/licenses/MIT.
- */
-
-package com.maddyhome.idea.vim.listener
-
-import com.intellij.openapi.editor.Editor
-
-interface VimInsertListener {
```
AlexPl292, line 13: Bad news, it's used in the IdeaVimExtension :)
Can we keep it and deprecate it first?

PR #959, opened 2024-08-05
Fix 'incsearch' highlight in multiple Rider instances

citizenmatt:
Unlike other IDEs, Rider has multiple client sessions. The IDE itself is the "local" session, just like other IDEs, and each external ReSharper process is treated as a "frontend" process - one for each opened solution. The code to get local editors was erroneously getting `ALL` sessions, rather than just `LOCAL` sessions, and assuming that the first was the local session. In Rider, opening three instances would add three additional "frontend" clients, and this would change the order of the returned sessions, so the first instance was an arbitrary "frontend" session, rather than the expected "local" session. This meant IdeaVim was trying to retrieve the wrong list of editors to work with. The most obvious impact of this is that search highlights were not shown during incremental search, but could also cause problems for other features, such as file marks, jump lists, updating option values across open files, etc.

I changed this code a while ago (in #883) as part of a change to make search highlights work in additional places, such as diff windows. I don't know why I changed `LOCAL` to `ALL`. There doesn't seem to be a good or obvious reason for doing so. Changing it back to `LOCAL` works fine in Rider and IDEA, and also when testing in split mode (installed on front end). I can only assume it was a mistake during that last change. If it is required for some reason, we should make sure to document why.

Fixes [VIM-3549](https://youtrack.jetbrains.com/issue/VIM-3549)

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/group/EditorGroup.java
+++ b/src/main/java/com/maddyhome/idea/vim/group/EditorGroup.java
@@ -8,9 +8,7 @@
 
 package com.maddyhome.idea.vim.group;
 
-import com.intellij.execution.impl.ConsoleViewImpl;
 import com.intellij.find.EditorSearchSession;
-import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.client.ClientAppSession;
 import com.intellij.openapi.client.ClientKind;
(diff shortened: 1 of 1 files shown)


AlexPl292:
I see, thank you!

PR #970, opened 2024-08-16
VIM-566: Add za motion support for toggling folds

The1xDeveloper:
Original request for `za` I could find [here](https://youtrack.jetbrains.com/issue/VIM-566/Support-more-Vim-folding-commands)

Personally, I wanted this so adding it.

`ExpandCollapseToggleAction` is not yet exposed in `IdeActions` I have a PR [here](https://github.com/JetBrains/intellij-community/pull/2825) to expose it.

It's up to the maintainers here if you'd prefer I wait till that's merged so I dont have a hard-coded string.

I was able to build intelliJ with this change and validated the toggling works as I expect it, feedback welcome!

Note: I dont really understand the `@TestWithoutNeovim(SkipNeovimReason.FOLDING)` so please direct me if I should remove it


AlexPl292:
Yep, this looks good to me, thank you!

> @TestWithoutNeovim(SkipNeovimReason.FOLDING)

We have tests that use neovim. However, this feature is IJ-dependent, so this test will not execute the commands on neovim. Everything is current in your code.

PR #975, opened 2024-08-25
Fix(VIM-3615): Escape closes dialog while waiting for more keys

chylex:
https://youtrack.jetbrains.com/issue/VIM-3615/Escape-closes-dialog-while-waiting-for-more-keys

Please see the issue comments; I noticed strange inconsistent behavior with IdeaVIM and dialogs.


AlexPl292:
Thank you!

PR #978, opened 2024-08-28
Fix typo

felixwiemuth:


doc/set-commands.md
```diff
@@ -8,7 +8,7 @@ Every effort is made to make these options compatible with Vim behaviour.
 However, some differences are inevitable.
 
 ```
-'clipboard'     'cb'    Defines clipboard behavioue
+'clipboard'     'cb'    Defines clipboard behaviour
```
bartelink, line 11: arguably the file is in US english :D (it has z spellings) ;)
```suggestion
'clipboard'     'cb'    Defines clipboard behavior
```

AlexPl292: Well, the original doc has a different wording, but generally [Vim docs](https://vimhelp.org/options.txt.html#%27clipboard%27) uses "behavior" form :)

AlexPl292:
Thank you for your fix!

PR #979, opened 2024-08-28
Refactor CommandBuilder, arguments, etc.

citizenmatt:
This PR is a fairly thorough refactoring of `CommandBuilder`, `Argument` and `OperatorArguments`. It might be easier to review the commits separately, as they're fairly small and self-contained, while the final diff of `CommandBuilder` might be too messy to see what's changed.

The idea of this PR is to tighten up the API around building a command, mostly to try to avoid confusion on concepts such as count. Previously, there were at least three ways to get the count for command being executed. Each `Command` had a count, and if the command was an operator+motion (such as `dw`) then the motion would also be a `Command` and have its own count. Only one of these was valid, but it was unclear which, and why. The count is also available in `OperatorArguments.count`, but it's not clear if this is the count for the whole command, or just for the operator or for the argument to the operator, which is a motion.

Main changes in this PR:

* The `Command` class no longer represents an operator+motion (`dw`) as a `Command` with another `Command` as an argument. Instead, the `Command` has an action for the operator, and an argument that can be a pair of motion action plus optional argument (e.g. for `dfx`). This means only a single count.
* `Argument` has been refactored to be a sealed hierarchy of classes.
* `CommandBuilder` has been extensively refactored to have a simpler implementation. 
    * Instead of a list of `Command` parts, there is now just the action, plus an argument (which might be a motion action with its own argument). 
    * The API is simpler to use - mostly `addAction` and `addArgument`, with no state management.
    * More state has been encapsulated too, with fewer mutable public properties.
* The `OperatorArguments` class has been substantially deprecated, although not as much as I'd like.
    * This class is confusing. It's not clear from the name what it's supposed to do. It's used while executing all kinds of actions, not just operators.
    * It contains a `mode` property that is different to `editor.mode` - it's the mode _before_ the command is completed and is only used to capture a runtime flag for the motion action to behave differently when in Operator-pending mode. This has been deprecated, and is no longer used in the codebase - separate action implementations have been registered for Op-pending mode.
    * There is also an `isOperatorPending` property that also doesn't reflect the current `editor.mode`, and is unncessary - it could be replaced with `mode is Mode.OP_PENDING`. This has also been deprecated and is no longer used.
    * The `count` properties are unclear on what they represent. Given `2d3w`, is this just the count for the operator? Furthermore, there was existing code that would multiply the count with `Command.count`, which is already multiplied when the command is built. The count is now guaranteed to be the same as `Command.count`.
    * Some usages of `OperatorArguments` have been refactored out.
    * The `count` properties are now the only useful properties in the class. It would be good to remove this class completely and pass the count directly. However, I'm not sure how disruptive this would be on third party extensions. We can look at this in the future.
* The extension handling for AceJump/EasyMotion (where an external plugin moves the caret and IdeaVim tracks the ranges for operators and selections) now wraps the captured offsets into an editor action handler, similar to `TextObjectActionHandler`, which can provide a range for a given caret. This special case handling of offsets now uses the same logic as other motions, especially text objects.

There are other minor refactorings along the way. There are no significant bugs fixed as part of this PR, but it does fix some pathological edge cases.


vim-engine/src/main/kotlin/com/maddyhome/idea/vim/action/motion/leftright/MotionShiftArrowLeftAction.kt
```diff
@@ -18,23 +18,19 @@ import com.maddyhome.idea.vim.api.moveToMotion
 import com.maddyhome.idea.vim.command.Command
 import com.maddyhome.idea.vim.handler.ShiftedArrowKeyHandler
 
-/**
- * @author Alex Plate
```
AlexPl292, line 22: 😭

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/action/motion/leftright/MotionBackspaceAction.kt
```diff
@@ -20,8 +20,8 @@ import com.maddyhome.idea.vim.command.OperatorArguments
 import com.maddyhome.idea.vim.handler.Motion
 import com.maddyhome.idea.vim.handler.MotionActionHandler
 
-@CommandOrMotion(keys = ["<BS>", "<C-H>"], modes = [Mode.NORMAL, Mode.VISUAL, Mode.OP_PENDING])
-class MotionBackspaceAction : MotionActionHandler.ForEachCaret() {
+@CommandOrMotion(keys = ["<BS>", "<C-H>"], modes = [Mode.NORMAL, Mode.VISUAL])
+open class MotionBackspaceAction(private val allowPastEnd: Boolean = false) : MotionActionHandler.ForEachCaret() {
   override fun getOffset(
```
AlexPl292, line 25: Did you check that the class is initialized properly? It has a default value, but who knows how kotlin handles that.

AlexPl292 (approved):

AlexPl292:
Thank you for your refactoring!

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/action/motion/leftright/MotionShiftArrowLeftAction.kt
citizenmatt: Oops. Sorry. Not sure why I deleted that 😳

AlexPl292: No worries, I actually think these `@author` annotations are useless and not in trend anymore. I remember I used them just I saw the similar ones by Andrei Vlasovskikh :)

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/action/motion/leftright/MotionBackspaceAction.kt
citizenmatt: Yep. Checked, and it's initialised fine and works as expected

PR #984, opened 2024-09-04
Add custom digraph support

citizenmatt:
I found an old, unfinished branch and brought it up to date 😄 

This PR will add support for adding custom digraph using the `:digraphs` command interactively or in `~/.ideavimrc`, such as `:digraphs (0 9450` for the `⓪` character ([VIM-3623](https://youtrack.jetbrains.com/issue/VIM-3623)). It also updates the output of the `:digraphs` command to match Vim, and adds headers when called with `:digraphs!`. And it updates the list of digraphs to match Vim's current set, adding a number of missing digraphs such as `oo` for "bullet", `=e` for "euro" and so on.

Finally, the output panel now includes the width of the editor gutter when calculating the approximate number of columns it's got to display text.


AlexPl292:
Yay, this is a sweet set of changes!

By the way, have you used any automation to find the missing digraphs? I remember I noticed such a problem already and thought about making a test that compares our digraph list with the one from Vim. However, for the moment, I think this might be a bit complicated to implement and maintain. We can probably just make a yearly review of the digraph set. Is there some simple way to manually compare our digraph set with the one from Vim?

citizenmatt:
Semi-automated. I copied the digraphs from the docs, ran an awk script on them, pasted it into the code and checked the git diffs. This doesn't quite work, because there are additional digraphs for Vim 5.x compatibility. So the git diff has to be handled manually. (The awk command is in the comments for the digraph list)

Fortunately, it's incredibly rare for Vim to add new digraphs - I think they've added 10 in the last 20 years!

PR #989, opened 2024-09-14
Update ApplicationConfigurationMigrator.kt

GeorgCantor:


AlexPl292:
Hi! Thank you for your interest in IdeaVim's contribution! Unfortunatelly, I don't think this change gives a lot of improvement to the code. Also, in your case, the `extractVersion` function is called twice on the found version detector.

PR #1001, opened 2024-10-07
Fix links in doc/Home.md

SanderHestvik:
Currently these links give a 404


AlexPl292:
Thank you for the fix!

PR #1007, opened 2024-10-10
VIM-1824: surround - remove whitespace with closing bracket

trimailov:
**Context**:

In vim surround extension closing brackets (`}, ], )`) should remove whitespace when using `cs` movement.

Example:
- Before: `{ example }`
- Movement: `cs{}`
- After: `{example}`

This doesn't work currently. The text is left unchanged.

**Solution**:

The bug was because brackets were replaced with a string from `SURROUND_PAIRS` map, which does not have any context about removing characters.

Inspired from VSCode's VIM plugin[^1], I have introduced new class `SurroundPair` that will carry this context about the need to trim characters.

**Disclaimer**:

I have never written in `Kotlin` so solution may be not use best practices, though at least this PR seems to fix the problem and tests are passing.

**Ticket**:
- https://youtrack.jetbrains.com/issue/VIM-1824/Vim-Surround-Does-not-remove-whitespace-with-the-closing-bracket

[^1]: https://github.com/VSCodeVim/Vim/blob/04fe42aa815f638af14e4cf4c92e88109242c3e6/src/actions/plugins/surround.ts#L455

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/extension/surround/VimSurroundExtension.kt
+++ b/src/main/java/com/maddyhome/idea/vim/extension/surround/VimSurroundExtension.kt
@@ -165,7 +165,7 @@ internal class VimSurroundExtension : VimExtension {
           .map {
             val oldValue: List<KeyStroke>? = getRegisterForCaret(REGISTER, it)
             setRegisterForCaret(REGISTER, it, null)
-            SurroundingInfo(it, null, oldValue)
+            SurroundingInfo(it, null, oldValue, false)
           }
 
         // Delete surrounding's content
@@ -181,22 +181,23 @@ internal class VimSurroundExtension : VimExtension {
           }
 
           val registerValue = getRegisterForCaret(REGISTER, it.caret)
-          val innerValue = if (registerValue.isNullOrEmpty()) null else registerValue
+          val innerValue = if (registerValue.isNullOrEmpty()) emptyList() else registerValue
           it.innerText = innerValue
-        }
 
-        surroundings.forEach {
-          if (it.innerText == null && getRegisterForCaret(REGISTER, it.caret)?.isNotEmpty() == true) {
-            it.innerText = emptyList()
+          // Valid surroundings are only those that:
+          // - are validly wrapping with surround characters (i.e. parenthesis, brackets, tags, quotes, etc.);
+          // - or have non-empty inner text (e.g. when we are surrounding words: `csw"`)
+          if (currentSurrounding != null || innerValue.isNotEmpty()) {
+            it.isValidSurrounding = true
           }
         }
 
         surroundings
-          .filter { it.innerText != null } // we do nothing with carets that are not inside the surrounding
+          .filter { it.isValidSurrounding } // we do nothing with carets that are not inside the surrounding
           .map { surrounding ->
             val innerValue = injector.parser.toPrintableString(surrounding.innerText!!)
             val text = newSurround?.let {
-              val trimmedValue = if (newSurround.shouldTrim && innerValue.take(1) == " " && innerValue.takeLast(1) == " ") innerValue.substring(1, innerValue.length - 1) else innerValue
+              val trimmedValue = if (newSurround.shouldTrim) innerValue.trim() else innerValue
               it.first + trimmedValue + it.second
             } ?: innerValue
             val textData = PutData.TextData(null, injector.clipboardManager.dumbCopiedText(text), SelectionType.CHARACTER_WISE)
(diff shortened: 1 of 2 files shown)


AlexPl292:
Hi! Thank you for your contribution. The change looks good to me, however, I noticed that the new test fails on the following pairs:

```
arrayOf("[${c} ]", "${c}[ ]", "[]"),
arrayOf("[${c}]", "${c}[]", "[]"),
```

Can you please take a look?

src/main/java/com/maddyhome/idea/vim/extension/surround/VimSurroundExtension.kt
```diff
@@ -181,21 +181,25 @@ internal class VimSurroundExtension : VimExtension {
...
 
           val registerValue = getRegisterForCaret(REGISTER, it.caret)
-          val innerValue = if (registerValue.isNullOrEmpty()) null else registerValue
+          val innerValue = if (registerValue.isNullOrEmpty()) emptyList() else registerValue
           it.innerText = innerValue
-        }
 
-        surroundings.forEach {
-          if (it.innerText == null && getRegisterForCaret(REGISTER, it.caret)?.isNotEmpty() == true) {
-            it.innerText = emptyList()
```
trimailov, line 190: This condition won't be ever `true`, because of the lines above (L183-L185). 

Those lines set `innerText` to `null` if register is null or empty. But here we loop through surroundings again and try to find `innerText == null` for cases when register is not empty.

src/main/java/com/maddyhome/idea/vim/extension/surround/VimSurroundExtension.kt
```diff
@@ -181,21 +181,25 @@ internal class VimSurroundExtension : VimExtension {
           }
 
           val registerValue = getRegisterForCaret(REGISTER, it.caret)
-          val innerValue = if (registerValue.isNullOrEmpty()) null else registerValue
+          val innerValue = if (registerValue.isNullOrEmpty()) emptyList() else registerValue
```
trimailov, line 184: Inspired from removed code [0] - set the `innerValue` as `emptyList()` - this will play nicely in line 198 when we'll convert it to string[1], for those cases when we deal with empty parenthesis.

[0] https://github.com/JetBrains/ideavim/pull/1007/files#r1810373587

[1] https://github.com/JetBrains/ideavim/pull/1007/files#diff-beac7e0e821f40c90568283d4d7ac4ab37de7362fab924aa616f94220fa85173R198

trimailov:
> Hi! Thank you for your contribution. The change looks good to me, however, I noticed that the new test fails on the following pairs:
> 
> ```
> arrayOf("[${c} ]", "${c}[ ]", "[]"),
> arrayOf("[${c}]", "${c}[]", "[]"),
> ```
> 
> Can you please take a look?

Added a fix: https://github.com/JetBrains/ideavim/pull/1007/commits/f6681a1aa25efcbdf28d590afb3d6d9f0c6fc610

Couple things though:

1. For first case (`arrayOf("[${c} ]", "${c}[ ]", "[]")`) - space should be trimmed to nothing (I have tried this on `neovim` with `vim-surround` plugin and on `VSCode` with `Vim` plugin). All spaces should be trimmed in general from my experiments in `neovim`. Thus this particular case I've edited to this: `arrayOf("[${c} ]", "${c}[]", "[]")`, though you see more cases in the tests of the PR;

2. Second case (`arrayOf("[${c}]", "${c}[]", "[]")`) seemed like somewhat orthogonal bug in general. IdeaVim just deleted the parenthesis if they are empty for any `cs` movement. It is because `CSurroundHandler` works (to my understanding) in a way where it deconstructs the text we are working with in to inner texts, surrounding parentheses, saves old text in registry, etc. After it manipulates necessary parts - the text is rebuilt [1]. Though because inner text for empty parenthesis is `null` - all this text manipulation and rebuilding part was skipped[2] and empty registry was returned instead[3]. Similarly to what I've done with introducing `SurroundingPair` to track if we should trim spaces, I've added `isValidSurrounding` attribute to `SurroundingInfo` to track if we should skip processing of the surrounding or not. 

[1] https://github.com/JetBrains/ideavim/blob/master/src/main/java/com/maddyhome/idea/vim/extension/surround/VimSurroundExtension.kt#L197-L202

[2] https://github.com/JetBrains/ideavim/blob/master/src/main/java/com/maddyhome/idea/vim/extension/surround/VimSurroundExtension.kt#L195

[3] https://github.com/JetBrains/ideavim/blob/master/src/main/java/com/maddyhome/idea/vim/extension/surround/VimSurroundExtension.kt#L208

AlexPl292:
I fixed the conflicts and rebased the branch on master. Thank you for your contribution!

PR #1010, opened 2024-10-18
Minor fixes for incsearch highlighting and caret/scroll position for undo

citizenmatt:
This PR fixes a couple of small bugs:

1. Highlighting for `'incsearch'` now only applies to editors in the current project, rather than all open projects - [VIM-3682](https://youtrack.jetbrains.com/issue/VIM-3682)
2. Fixes an issues where undo/redo does not move the caret when used in an editor with a preview, such as Markdown files, or all source files in Android Studio (which includes a preview based on the code). IdeaVim will also ensure that `'scrolloff'` is applied after undo/redo - [VIM-3671](https://youtrack.jetbrains.com/issue/VIM-3671)


AlexPl292:
Merged, thank you!

PR #1012, opened 2024-10-23
Small bug fix to use exclusive Visual caret shape from 'guicursor'

citizenmatt:
This PR will use the caret shape specified by `ve` in the `'guicursor'` option when Visual mode is active and the `'selection'` option equals `exclusive`. If the `ve` element isn't specified in `'guicursor'`, the `v` element is used instead.

This PR will also now use the `ve` element for Select mode. Previously, we were using the `i` Insert mode cursor (Vim doesn't have a separate Select cursor). This felt like a reasonable change at the time for several reasons. Firstly, Select mode is more similar to a traditional editor selection where typing inserts the text, replacing the selection, so using the Insert mode caret was an intuitive indicator of this. Secondly, when used in refactorings, the caret would be positioned after the selection, so a bar caret looks better.

However, the caret being positioned after the selection is because IdeaVim incorrectly treats Select mode as exclusive at all times, rather than based on the `'selection'` option, like Visual. It is therefore more appropriate to use the `ve` cursor than the `i` cursor. Fortunately, the default for this is a bar caret, so there is little visual difference with this change. It is, however, more correct.

At some point, we need to fix Select's implementation and make it inclusive by default ([VIM-3687](https://youtrack.jetbrains.com/issue/VIM-3687)). At this point, we will need to review the Select mode caret shape, as the bar will no longer be appropriate, but it will also be less intuitive for those coming from traditional editors.

### DIFF
--- a/src/test/java/org/jetbrains/plugins/ideavim/helper/CaretVisualAttributesHelperTest.kt
+++ b/src/test/java/org/jetbrains/plugins/ideavim/helper/CaretVisualAttributesHelperTest.kt
@@ -80,6 +80,15 @@ class CaretVisualAttributesHelperTest : VimTestCase() {
 
   @TestWithoutNeovim(SkipNeovimReason.NOT_VIM_TESTING)
   @Test
+  fun `test default exclusive visual mode caret is bar`() {
+    configureByText("Lorem ipsum dolor sit amet,")
+    enterCommand("set selection=exclusive")
+    typeText("ve")
+    assertCaretVisualAttributes(CaretVisualAttributes.Shape.BAR, 0.35F)
+  }
+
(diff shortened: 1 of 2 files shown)


AlexPl292:
Do I understand it right that there'll be no changes in the behaviour and caret shape in the case of the default options?

citizenmatt:
Mostly, yes, that's correct. There are no changes to behaviour. Select mode will still use the bar caret, but instead of it being hardcoded, it uses the `ve` segment from `'guicursor'`. The default value for this is a bar. 

The only change in caret shape is when in Visual mode, if `'selection'` is set to `exclusive`, then we also use the `ve` segment here, which matches Vim behaviour. If the `ve` segment isn't defined (it is in defaults), then we fall back to `v`.

AlexPl292:
Great, thank you!

PR #1029, opened 2024-11-07
Refactor map implementation and restore which-key compatibility

citizenmatt:
This PR restores compatibility with idea-which-key, which was broken due to changes to `CommandPartNode` in #979.

The `CommandPartNode<T>` classes have been deprecated and reimplemented in terms of a new `KeyStrokeTrie<T>` class, that efficiently maps sequences of `KeyStroke` to data `T`. The trie structure is used to store and look up the builtin Vim commands (represented by a `T` of `LazyVimCommand`). The `CommandPartNode` API is restored to how it was before #979, and should work with the last released version of which-key. 

It would be appropriate for which-key to migrate to the new API, which can provide more efficient lookup and iteration.

The new trie structure is now also used for user maps, simplifying and optimising storage and lookup of both finished key sequences and prefixes. While refactoring handling of the `map` commands, several fixes have been made:

* Improve the output of maps registered for multiple modes, e.g. `vmap` for Visual + Select, `map` for Normal, Visual, Select and Op-pending. The maps will now have the proper character indicator(s), especially when a map is made in multiple modes and then removed in one, e.g. `map foo bar` followed by `sunmap foo` will now output `nox foo bar`
* Fix a bug that would only show one map with the same left-hand side set of keys. If a map was registered with the same keys in different modes, only one map would be output.
* Filter map output by prefix (so `map foo` will show all maps beginning with `foo` instead of creating a map from `foo` to `<Nop>`) - [VIM-2981](https://youtrack.jetbrains.com/issue/VIM-2981)
* Support `map!`, `unmap!` and `mapclear!` to add a map in Insert+Command-line modes
* Implement missing `smap` and `snoremap`. `sunmap` was already supported, and mapping to Select mode was already possible with `vmap` - [VIM-2260](https://youtrack.jetbrains.com/issue/VIM-2260)
* Fix a bug removing a mapping which had keys that were also a prefix to another mapping

The support for `map!` required adding the "bang modifier" to the parser, rather than relying on the first character of the trailing string argument - this is required to tell the difference between `map! foo bar` and `map ! foo bar`.

The trie is also used in the NERDTree extension


vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/model/commands/mapping/MapCommand.kt
```diff
@@ -31,7 +31,7 @@ import javax.swing.KeyStroke
 /**
  * @author vlan
  */
-@ExCommand(command = "map,nm[ap],vm[ap],xm[ap],smap,om[ap],im[ap],lm[ap],cm[ap],no[map],nn[oremap],vn[oremap],xn[oremap],snor[emap],ono[remap],no[remap],ino[remap],ln[oremap],cno[remap]")
+@ExCommand(command = "map,nm[ap],vm[ap],xm[ap],smap,om[ap],im[ap],lm[ap],cm[ap],nn[oremap],vn[oremap],xn[oremap],snor[emap],ono[remap],no[remap],ino[remap],ln[oremap],cno[remap]")
```
AlexPl292, line 36: What happened to nomap?

AlexPl292 (approved):

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/model/commands/mapping/MapCommand.kt
citizenmatt: It's not a Vim command! There's `nmap` and `noremap` but no `nomap`

AlexPl292: Huh! I can't track the change again, but this is funny :D

PR #1031, opened 2024-11-08
Enable lookahead for argument text object

jphalip:
This allows motions like `caa` or `dia` to look ahead for potential arguments, say when the caret is positioned before the list of arguments. This would work similar to https://github.com/vim-scripts/argtextobj.vim and to standard motions like `ci"`.

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/extension/argtextobj/VimArgTextObjExtension.java
+++ b/src/main/java/com/maddyhome/idea/vim/extension/argtextobj/VimArgTextObjExtension.java
@@ -298,33 +298,52 @@ public class VimArgTextObjExtension implements VimExtension {
      * @param position starting position.
      */
     boolean findBoundsAt(int position) throws IllegalStateException {
-      if (text.isEmpty()) {
+      if (text.length() == 0) {
         error = "empty document";
         return false;
       }
+
+      // First try original position
       leftBound = Math.min(position, leftBound);
       rightBound = Math.max(position, rightBound);
+
+      // Get out of quoted text and handle any current position
       getOutOfQuotedText();
+      char ch = getCharAt(rightBound);
       if (rightBound == leftBound) {
-        if (brackets.isCloseBracket(getCharAt(rightBound))) {
+        if (ch == 0) {
+          error = "invalid position";
+          return false;
+        }
+        if (brackets.isCloseBracket(ch)) {
           --leftBound;
+        } else if (brackets.isOpenBracket(ch)) {
+          ++rightBound;
         } else {
           ++rightBound;
         }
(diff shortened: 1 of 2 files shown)


src/main/java/com/maddyhome/idea/vim/extension/argtextobj/VimArgTextObjExtension.java
```diff
@@ -298,33 +298,52 @@ private static class ArgBoundsFinder {
      * @param position starting position.
      */
     boolean findBoundsAt(int position) throws IllegalStateException {
-      if (text.isEmpty()) {
+      if (text.length() == 0) {
```
lippfi, line 301: Isn't `isEmpty()` the same, but more readable option?

src/main/java/com/maddyhome/idea/vim/extension/argtextobj/VimArgTextObjExtension.java
```diff
@@ -298,33 +298,52 @@ private static class ArgBoundsFinder {
...
       if (rightBound == leftBound) {
-        if (brackets.isCloseBracket(getCharAt(rightBound))) {
+        if (ch == 0) {
+          error = "invalid position";
+          return false;
+        }
+        if (brackets.isCloseBracket(ch)) {
           --leftBound;
+        } else if (brackets.isOpenBracket(ch)) {
+          ++rightBound;
```
lippfi, line 321: Th else branch will do the same, do we need this?

lippfi (changes requested):
Could you please have a look at the comments?

src/main/java/com/maddyhome/idea/vim/extension/argtextobj/VimArgTextObjExtension.java
jphalip: Good point. Fixed.

src/main/java/com/maddyhome/idea/vim/extension/argtextobj/VimArgTextObjExtension.java
jphalip: Good point. Fixed.

jphalip:
@lippfi Good catch with that filing example. I've made some fixes, which should address that.

lippfi:
This is what happens when I place caret inside "Hi there :(" and type `via`
image

AlexPl292:
Hi, I’m closing this PR due to inactivity. Feel free to reopen it or create a new one.

jphalip:
No worries, thanks. I've been busy but still intend to continue working on this. I'll reopen when I find the time.

PR #1034, opened 2024-11-08
Add some missing maps

citizenmatt:
* Adds a map for `<C-PageUp/Down>` and `i_<C-PageUp/PageDown>` to switch tabs - [VIM-2044](https://youtrack.jetbrains.com/issue/VIM-2044)
* Adds a map for `<S-Enter>` to scroll page forward - [VIM-2752](https://youtrack.jetbrains.com/issue/VIM-2752)
* Removes some old extended handling of special keys that was required when keys for actions were registered via a comma separated string in an XML attribute. Removes recognising `«` and `»` and also `<Comma>`

### DIFF
--- a/src/test/java/org/jetbrains/plugins/ideavim/action/scroll/ScrollPageDownActionTest.kt
+++ b/src/test/java/org/jetbrains/plugins/ideavim/action/scroll/ScrollPageDownActionTest.kt
@@ -58,6 +58,16 @@ class ScrollPageDownActionTest : VimTestCase() {
 
   @TestWithoutNeovim(SkipNeovimReason.SCROLL)
   @Test
+  fun `test scroll single page down with S-Enter`() {
+    configureByPages(5)
+    setPositionAndScroll(0, 0)
+    typeText("<S-Enter>")
+    assertPosition(33, 0)
+    assertVisibleArea(33, 67)
+  }
+
+  @TestWithoutNeovim(SkipNeovimReason.SCROLL)
+  @Test
   fun `test scroll page down in insert mode with S-Down`() {
     configureByPages(5)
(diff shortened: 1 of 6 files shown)


AlexPl292:
Thank you!

rickgladwin:
Is there a way to add changes like this as a note that will appear in the Overview tab or What's New tab in the Plugins listing in JetBrains IDEs?

This represents a breaking change for any of us that use the default behaviour for the IDE when using IdeaVim (Shift+Enter starts a new line), but the version number wasn't increased to a major version. I checked the changelog under the "What's New" tab, and this change isn't mentioned directly there either, since the change item was a feature request and not a fix.

rickgladwin:
For anyone who winds up here while trying to figure out why the behaviour changed, this remapping in `.ideavimrc` will restore the default behaviour:
```
" map Shift+Enter to 'Start New Line' in normal mode
nmap <S-Enter> o<ESC>
" map Shift+Enter to 'Start New Line' in insert mode
imap <S-Enter> <ESC>o
```

(That said, I appreciate the hard work that goes into an open source project like this. It's a great resource and contributions are important and valuable)

citizenmatt:
Release notes aren't currently tracked automatically, but you can see what YouTrack tickets have been fixed in a release with a query such as: https://youtrack.jetbrains.com/issues?q=project:VIM%20Fix%20versions:%202.17.0. There are also sometimes changes that don't have YouTrack tickets, and those are in PRs like this one.

Apologies for breaking your muscle memory with this, but it should always have been a Vim shortcut, rather than an IDE one, and thanks for posting a workaround to restore the previous behaviour. You could probably also work around this issue by using `sethandler <S-Enter> a:IDE`.

PR #1037, opened 2024-11-12
Fix memory leak with non-disposed listeners

citizenmatt:
When an editor is initialised, IdeaVim registers various listeners and uses a new per-editor disposable to remove the registrations. When the IdeaVim plugin is disabled/unloaded by the platform, the parent disposable is disposed, and all listeners are unregistered. All per-editor disposables are also disposed when IdeaVim disables Vim functionality without disbaling the whole plugin. However, the per-editor disposable is not disposed at any other time, e.g., when the editor is closed. So IdeaVim is leaking disposables.

Since each registration's disposable needs to call an editor method to remove the listener, each disposable captures a reference to the editor. Because the disposables are not removed when an editor is closed, that means IdeaVim is leaking editor instances.

(Some listener registrations simply add a reference _from_ the editor to a singleton listener handler. If there was no disposable to remove the registration, there would be no reference to the editor to keep it alive and GC could collect it. However, the disposable does maintain a reference, and all listener registrations keep the editor from being collected)

This PR will explicitly dispose the per-editor disposable when the editor is released, and adds tests to confirm this.

Fixes [VIM-3319](https://youtrack.jetbrains.com/issue/VIM-3319)


AlexPl292:
Thank you! So much PITA with disposables in this place.

PR #1038, opened 2024-11-14
Add support for escape() vim function

jphalip:


src/test/java/org/jetbrains/plugins/ideavim/ex/implementation/functions/BuiltInFunctionTest.kt
```diff
@@ -177,4 +177,28 @@ class BuiltInFunctionTest : VimTestCase() {
...
+    assertExOutput("no escaping needed")
+
+    // Test empty strings
+    typeText(commandToKeys("""echo escape('', '')"""))
+    assertExOutput("")
+
+    // Test consecutive characters that need escaping
+    typeText(commandToKeys("""echo escape('$$$$', '$')"""))
+    assertExOutput("""\$\$\$\$""")
+  }
```
citizenmatt, line 203: It would be good to extract this to a separate test class, perhaps under a `builtin` folder, so we can have separate files for each function, and allow splitting the test into more logical tests/asserts.

citizenmatt: Probably worth adding a couple of tests for invalid method calls - too many parameters, not enough parameters, what happens with invalid data types, etc.

AlexPl292:
Thank you for your PR! Can you please take a look and write a test on how this would work with multicharacter symbols like `😀`?

AlexPl292 (changes requested):
Please check the comments

jphalip:
I've improved the implementation to support unicodes, emojis, etc. and added multiple other tests.

However, maybe this is overachieving? For example, standard vim appears to ignore escaping of emojis:

Standard vim:
```vim
:echo escape("😀🎉✨", "😀✨")
😀🎉✨
```

This implementation:
```kotlin
typeText(commandToKeys("""echo escape("😀🎉✨", "😀✨")"""))
assertExOutput("\\😀🎉\\✨")
```

Let me know if you're ok with this or if you'd like to scale back to the standard behavior.

src/test/java/org/jetbrains/plugins/ideavim/ex/implementation/functions/BuiltInFunctionTest.kt
jphalip: @citizenmatt I've added a couple of tests with errors.

I also agree that it'd be nice to separate these buitlin functions out to separate class, but maybe this should be done in bulk in a separate PR?

AlexPl292:
Oh, sorry, I didn't mean THAT accurate :D
I mentioned `😀` because I thought that iterating over code points (via `String.codePoints()`) would generally solve the issue. Will working just with codepoints make the escaping more accurate? If not, let's use the original behavior because it looks much simpler and more stable :)

AlexPl292:
Note: I closed & reopened the PR by accident

jphalip:
@AlexPl292 I've reverted to my original, simpler implementation. I also added some tests for unicode, and verified that the results are the same in standard Vim.

AlexPl292:
Great, thank you!

PR #1039, opened 2024-11-14
Add partial support for getcmdtype() function

jphalip:
The `getcmdtype()` could be called from inside a prompt (search, expression, command), or from a remapping. Apparently IdeaVim currently doesn't support nested prompts (e.g. by typing `:<C-r>=` to show an expression prompt inside a command prompt), so it wouldn't be functional in that case.

It should work in a remapping though, e.g.: `:cnoremap <F2> <C-r>=getcmdtype()<CR>`. Could you advise on how I could unit test this?

For context, this is the remap that I'm using this for:

```
cnoremap <expr> <Esc> getcmdtype() =~ '[/?]' ? '<CR><Esc>' : '<Esc>'
```

This is so that if you press `<Escape>` during a search before you press enter, it keeps the cursor at the first search result instead of jumping back to where the cursor was before the search started. This enforces a behavior that you find in most text search tools, including IntelliJ's built-in search.

### DIFF
--- a/src/test/java/org/jetbrains/plugins/ideavim/ex/implementation/functions/GetCmdTypeFunctionTest.kt
+++ b/src/test/java/org/jetbrains/plugins/ideavim/ex/implementation/functions/GetCmdTypeFunctionTest.kt
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2003-2024 The IdeaVim authors
+ *
+ * Use of this source code is governed by an MIT-style
+ * license that can be found in the LICENSE.txt file or at
+ * https://opensource.org/licenses/MIT.
+ */
+
+package org.jetbrains.plugins.ideavim.ex.implementation.functions
+
+import com.maddyhome.idea.vim.api.injector
+import com.maddyhome.idea.vim.ui.ex.ExEntryPanel
+import org.jetbrains.plugins.ideavim.VimTestCase
+import org.junit.jupiter.api.Test
+import kotlin.test.assertEquals
+
+class GetCmdTypeFunctionTest : VimTestCase() {
+
+  @Test
+  fun `test getcmdtype() for a regular command`() {
+    configureByText("\n")
+    enterCommand("cmap <expr> z getcmdtype()")
+    typeText(":fooz")
+    assertEquals("foo:", (injector.commandLine.getActiveCommandLine() as ExEntryPanel).visibleText)
+  }
+
+  @Test
+  fun `test getcmdtype() for a forward search`() {
+    configureByText("\n")
+    enterCommand("cmap <expr> z getcmdtype()")
+    typeText("/fooz")
+    assertEquals("foo/", (injector.commandLine.getActiveCommandLine() as ExEntryPanel).visibleText)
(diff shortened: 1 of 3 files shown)


jphalip:
By the way, do you know what the "input command" (labelled `@`) corresponds to? For reference, here's the manual entry:

```
getcmdtype()

		Return the current command-line type. Possible return values
		are:
		    :	normal Ex command
		    >	debug mode command [debug-mode](https://neovim.io/doc/user/repeat.html#debug-mode)
		    /	forward search command
		    ?	backward search command
		    @	[input()](https://neovim.io/doc/user/builtin.html#input()) command
		    -  	[:insert](https://neovim.io/doc/user/insert.html#%3Ainsert) or [:append](https://neovim.io/doc/user/insert.html#%3Aappend) command
		    =	[i_CTRL-R_=](https://neovim.io/doc/user/insert.html#i_CTRL-R_%3D)  
		Only works when editing the command line, thus requires use of
		[c_CTRL-\_e](https://neovim.io/doc/user/cmdline.html#c_CTRL-%5C_e) or [c_CTRL-R_=](https://neovim.io/doc/user/cmdline.html#c_CTRL-R_%3D) or an expression mapping.
		Returns an empty string otherwise.
		Also see [getcmdpos()](https://neovim.io/doc/user/builtin.html#getcmdpos()), [setcmdpos()](https://neovim.io/doc/user/builtin.html#setcmdpos()) and [getcmdline()](https://neovim.io/doc/user/builtin.html#getcmdline()).
```

citizenmatt:
The input command is the prompt shown when someone calls the `input()` function. So you could do something like this in Vim:

```
cmap <expr> z getcmdtype()
```

Now every time you type `z` in any kind of prompt, it will add the appropriate symbol to the prompt text. For a command, you'll get a `:` symbol and search will give `/` or `?`. If you then type:

```
:call input("Prompt> ")
```

You'll get an input prompt for some kind of text (e.g. Vim script to capture something) and typing `z` will add `@` to the prompt.

As for testing, a simple map like that would do it. Something like:

```
enterCommand("cmap <expr> z getcmdtype()")
typeText(":z")
assertEquals(":", injector.commandLine.visibleText)
```

or 


```
enterCommand("cmap <expr> z getcmdtype()")
typeText("/fooz")
assertEquals("foo/", injector.commandLine.visibleText)
```

etc. Testing the expression register would be very similar, but with a different map:

```
enterCommand("cnoremap <F2> <C-r>=getcmdtype()<CR>")
typeText("?foo<F2>")
assertEquals("foo?", injector.commandLine.visibleText)

jphalip:
@citizenmatt Thanks a lot for the details. I just added some tests for `:/?=`. However, I'm not sure input, insert, and append command would be supported since I think they'd require nested prompts, which doesn't work in IdeaVim. Also, I'm not sure about the debug mode (`>`) -- Is that reproducible and testable with IdeaVim?

citizenmatt:
Looks good to me. I wouldn't worry about debug-mode, it's not supported by IdeaVim.

PR #1040, opened 2024-11-14
Add support for hlsearch variable

jphalip:


src/main/java/com/maddyhome/idea/vim/newapi/IjVimSearchGroup.kt
```diff
@@ -125,6 +125,19 @@ open class IjVimSearchGroup : VimSearchGroupBase(), PersistentStateComponent<Ele
...
+    }
+    for (vimEditor in vimEditors) {
+      val editor = vimEditor.ij
+      if (editor.vimLastHighlighters != null) {
+        return true
+      }
+    }
+    return false
+  }
+
```
citizenmatt, line 140: This is fine, but personally, I'd prefer an implementation like I outlined in [this discussion comment](https://github.com/JetBrains/ideavim/discussions/1036#discussioncomment-11253104) - implement `v:hlsearch` as a variable with state in `VimVariableServiceBase` and move state from `VimSearchGroup` to the variable service. 

The `v:hlsearch` variable would then be the source of truth of whether or not highlights were active, and there wouldn't be a need to check with all editors or introduce a new API.

src/test/java/org/jetbrains/plugins/ideavim/ex/implementation/services/VimVariableServiceTest.kt
```diff
@@ -113,4 +113,41 @@ class VimVariableServiceTest : VimTestCase() {
...
+    enterSearch("one")
+    enterCommand("nohlsearch")
+    enterCommand("echo v:hlsearch")
+    assertExOutput("0")
+
+    // After n command reusing previous search, should be 1 again
+    typeText("n")
+    enterCommand("echo v:hlsearch")
+    assertExOutput("1")
+  }
```
citizenmatt, line 152: As mentioned in the other PRs, extracting to a new file and splitting into separate functions would be good.

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/services/VimVariableServiceBase.kt
```diff
@@ -179,6 +179,7 @@ abstract class VimVariableServiceBase : VariableService {
       "count" -> VimInt(KeyHandler.getInstance().keyHandlerState.commandBuilder.calculateCount0Snapshot())
       "count1" -> VimInt(KeyHandler.getInstance().keyHandlerState.commandBuilder.calculateCount0Snapshot().coerceAtLeast(1))
       "searchforward" -> VimInt(if (injector.searchGroup.getLastSearchDirection() == Direction.FORWARDS) 1 else 0)
+      "hlsearch" -> VimInt(if (injector.searchGroup.isSomeTextHighlighted()) 1 else 0)
```
lippfi, line 182: There is a `Boolean.asVimInt(): VimInt` method, but it's also ok

AlexPl292 (commented):
Thank you for your PR!. What do you think about the @ citizenmatt suggestion?

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/services/VimVariableServiceBase.kt
jphalip: Good point, I made the change.

src/test/java/org/jetbrains/plugins/ideavim/ex/implementation/services/VimVariableServiceTest.kt
jphalip: I agree. I suggest we wait until the other PR for the `v:register` (#1046) variable is available so we can access the new `Variable` interface.

citizenmatt:
Sure. It's not a value that we update when a highlight is shown or hidden, it's a value we update in order to show or hide highlights. It's the flag that controls whether the highlights are shown. It defaults to `true` and we set it to `false` when we call `:nohlsearch`.

So, I would replace this read/write property in `IjVimSearchGroup`:

https://github.com/JetBrains/ideavim/blob/82d18cfbb95487da85c4cd131836746bc78c2e49/src/main/java/com/maddyhome/idea/vim/newapi/IjVimSearchGroup.kt#L67

with a read-only property, something like:

```
private val showSearchHighlight: Boolean
  get() = injector.globalOptions().hlsearch && injector.variableService.getNonNullVariableValue(Variable("v", "hlsearch")).asBoolean()
```

And then whenever we currently set `showSearchHighlight`, we set the `v:hlsearch`, either to `true` or `false` so this:

https://github.com/JetBrains/ideavim/blob/82d18cfbb95487da85c4cd131836746bc78c2e49/src/main/java/com/maddyhome/idea/vim/newapi/IjVimSearchGroup.kt#L132-L135

would be something like:

```
override fun clearSearchHighlight() {
  injector.variableService.storeVariable(Variable("v", "hlsearch"), false.asVimInt(), ???, ???, ???)
}
```

And `storeVariable` would update the flag and then call `injector.searchGroup.updateSearchHighlights`. This would be a new method added to the `VimSearchGroup` interface. It's already defined in `VimSearchGroupBase`, but not the interface.

The only tricky bit is what else to pass to `storeVariable`. It requires an editor, an execution context and a vim script context, and we don't have those or even need them for Vim variables. I'd be inclined to make them nullable.

jphalip:
@citizenmatt Thanks for the details. I think I get what you're saying and I started to look into it, but it seems that some function calls further down the stack might need some of the editor/context values, for example this one in `VimVariableServiceBase`:

```kotlin
  protected fun getDefaultVariableScope(executable: VimLContext): Scope {
    return when (executable.getExecutableContext(executable)) {
      ExecutableContext.SCRIPT, ExecutableContext.COMMAND_LINE -> Scope.GLOBAL_VARIABLE
      ExecutableContext.FUNCTION -> Scope.LOCAL_VARIABLE
    }
  }
```

jphalip:
Hi all - I've refactored the code to make use of the newly added `Variable` interface, and squashed the commits. @citizenmatt I agree that all of this can be refactored later if needed. @AlexPl292 Let me know if this looks good to you. Thanks!

AlexPl292:
Okay, the current solution seems good to me. Thank you!

PR #1043, opened 2024-11-15
Fix options initialisation in 2024.2+

citizenmatt:
This PR will fix options initialisation in 2024.2 and above.

Options are initialised based on the currently selected editor (Vim "window"). In 2024.1, we could use `FileEditorManager.selectedTextEditor` to get the last selected text editor while creating a new editor. This value can no longer be relied upon in 2024.2, as the value is reset to `null` when a new editor (that will become the next selected editor) is being created.

This meant we were always initialising window options based on the "fallback" window. Vim always has at least one window open, so there is always a window to initialise _from_. IntelliJ can close all windows, so we need to keep track of the options used by the window that was last to be closed - so if there are no open windows, we can initialise from the "fallback" window.

However, a further change in behaviour meant that we were updating this fallback window too frequently -  whenever _any_ window closed. This mostly mitigated the first problem, except for options that are mapped to IntelliJ settings, e.g. `'wrap'`. This lead to some tests for `'wrap'` being disabled for 242.

This PR will:
* Update option tests to make them easier to understand and to ensure that the fallback window is updated correctly. This should help diagnose similar problems if behaviour changes in the future
* Change the method of updating the fallback window so it's only updated when the last window in a project is closed
* Start manually tracking the last selected editor/window so that unexpected `null` values don't cause trouble
* Reinstate disabled `'wrap'` tests

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/group/OptionGroup.kt
+++ b/src/main/java/com/maddyhome/idea/vim/group/OptionGroup.kt
@@ -12,6 +12,7 @@ import com.intellij.application.options.CodeStyle
 import com.intellij.codeStyle.AbstractConvertLineSeparatorsAction
 import com.intellij.openapi.Disposable
 import com.intellij.openapi.application.ApplicationManager
+import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.editor.EditorKind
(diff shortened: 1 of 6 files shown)


AlexPl292:
Awesome, thank you for your work!

PR #1046, opened 2024-11-18
Add support for the v:register variable

jphalip:


src/test/java/org/jetbrains/plugins/ideavim/ex/implementation/services/VimVariableServiceTest.kt
```diff
@@ -113,4 +115,35 @@ class VimVariableServiceTest : VimTestCase() {
...
+    // itself, i.e. it hasn't been committed yet as the current register
+    enterCommand(
+      """
+      vnoremap <expr> y '"' . v:register . 'y'
+      """.trimIndent(),
+    )
+    typeText("vl\"zy")
+    val register = injector.registerGroup.getRegisters().filter { reg -> reg.name == 'z' }.first()
+    assertEquals("ab", register.text)
+  }
```
citizenmatt, line 148: I think it would be a good idea to extract this into a new class (perhaps in a new `variables` folder), and split down into separate functions. Hopefully, we'll implement more variables, and it'll save this test class getting overcrowded. It's also good to have tests that test a single thing.

citizenmatt:
It definitely makes sense to extract variables into their own classes as we add more, but I don't think there's much urgency, based on how many we've got right now. So I'd be happy to leave `count` etc. for another PR. 

And it's possibly too early to decide on registration. If it's only a handful of variables, it might be simpler just to statically register them in a string map. Otherwise, I'd probably follow the same pattern as the `@VimscriptFunction` and `@CommandOrMotion` annotations so we could mark up classes as Vim variables, run a processor at compile time to generate a JSON file, and then deserialise this file at startup and lazily instantiate as needed.

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/model/variables/Variable.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import com.maddyhome.idea.vim.api.ExecutionContext
+import com.maddyhome.idea.vim.api.VimEditor
+import com.maddyhome.idea.vim.vimscript.model.VimLContext
+import com.maddyhome.idea.vim.vimscript.model.datatypes.VimDataType
+
+/**
+ * Base class for Vim variables
+ */
+abstract class Variable {
```
AlexPl292, line 19: Let's change this to interface, we don't have any implementations

AlexPl292 (approved):
Looks good to me. Let's make the `abstract class` -> `interface` update and I'll merge it.

jphalip:
@AlexPl292 Thanks for the review. I just made that change.

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/model/variables/Variable.kt
jphalip: Done!

PR #1053, opened 2024-11-28
Fix repeating change action with count

citizenmatt:
Repeating the `c` command should not repeat inserting the text. This is mostly visible when the command is initially made with a count. E.g. `2c3l` will change 6 characters. Hitting `.` should also change 6 characters and shouldn't repeat inserting the text 6 times.

This PR adds `CommandFlags.FLAG_NO_REPEAT_INSERT`, as used by `s`, `S` and `C`.

Fixes [VIM-3729](https://youtrack.jetbrains.com/issue/VIM-3729)


AlexPl292:
Thank you!

PR #1054, opened 2024-12-02
Add search widget

jphalip:
This widget is activated while doing a search (forward/backward, word search, etc.). It shows the current occurrence that you're on, as well as the total number of occurrences for the search terms. Works similarly to IntelliJ's built-in search tool.

This relies on search highlights being shown. But perhaps this is acceptable since the users who would care about this widget would also likely have highlights turned on?

If tests are expected for this, could you point me to some similar examples? I couldn't find tests, for example, for the macro widget.

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/ui/widgets/search/SearchWidgetFactory.kt
+++ b/src/main/java/com/maddyhome/idea/vim/ui/widgets/search/SearchWidgetFactory.kt
@@ -1,0 +1,118 @@
+/*
+ * Copyright 2003-2024 The IdeaVim authors
+ *
+ * Use of this source code is governed by an MIT-style
+ * license that can be found in the LICENSE.txt file or at
+ * https://opensource.org/licenses/MIT.
+ */
+
+package com.maddyhome.idea.vim.ui.widgets.search
+
+import com.intellij.openapi.components.service
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.project.ProjectManager
+import com.intellij.openapi.wm.StatusBarWidget
+import com.intellij.openapi.wm.StatusBarWidgetFactory
+import com.intellij.openapi.wm.WindowManager
+import com.intellij.openapi.wm.impl.status.widget.StatusBarWidgetsManager
+import com.maddyhome.idea.vim.VimPlugin
+import com.maddyhome.idea.vim.api.VimEditor
+import com.maddyhome.idea.vim.api.globalOptions
+import com.maddyhome.idea.vim.api.injector
+import com.maddyhome.idea.vim.common.SearchListener
+import com.maddyhome.idea.vim.helper.vimLastHighlighters
+import com.maddyhome.idea.vim.newapi.ij
+import com.maddyhome.idea.vim.ui.widgets.VimWidgetListener
+import com.maddyhome.idea.vim.ui.widgets.mode.VimStatusBarWidget
+import java.awt.Component
+
+private const val ID = "IdeaVimSearch"
+
+internal class SearchWidgetFactory : StatusBarWidgetFactory {
+
+  override fun getId(): String {
+    return ID
+  }
+
+  override fun getDisplayName(): String {
+    return "IdeaVim Search"
+  }
+
+  override fun createWidget(project: Project): StatusBarWidget {
+    return VimSearchWidget()
+  }
+
+  override fun isAvailable(project: Project): Boolean {
+    return VimPlugin.isEnabled() && injector.globalOptions().showmode
+  }
+}
+
+fun updateSearchWidget() {
+  val factory = StatusBarWidgetFactory.EP_NAME.findExtension(SearchWidgetFactory::class.java) ?: return
+  for (project in ProjectManager.getInstance().openProjects) {
+    val statusBarWidgetsManager = project.service<StatusBarWidgetsManager>()
+    statusBarWidgetsManager.updateWidget(factory)
+  }
+}
+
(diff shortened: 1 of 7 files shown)


clover256s:
人类历史上最伟大的插件ideavim

AlexPl292:
Related ticket: https://youtrack.jetbrains.com/issue/VIM-2544

src/main/java/com/maddyhome/idea/vim/ui/widgets/search/SearchWidgetFactory.kt
```diff
@@ -0,0 +1,118 @@
...
+fun updateSearchWidget() {
+  val factory = StatusBarWidgetFactory.EP_NAME.findExtension(SearchWidgetFactory::class.java) ?: return
+  for (project in ProjectManager.getInstance().openProjects) {
+    val statusBarWidgetsManager = project.service<StatusBarWidgetsManager>()
+    statusBarWidgetsManager.updateWidget(factory)
+  }
+}
+
+class VimSearchWidget : StatusBarWidget, VimStatusBarWidget {
+  var content: String = ""
```
AlexPl292, line 59: Directly accessing the `content` here leaks incapsulation. Let's have special functions like `setSearch(current, total)` and `clearSearch()`.
(function names are subject to change).

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimSearchGroupBase.kt
```diff
@@ -1340,13 +1340,15 @@ abstract class VimSearchGroupBase : VimSearchGroup {
       } else if (lastPatternTrailing!![ppos + 1] == '?') {
         Direction.BACKWARDS
       } else {
+        injector.listenersNotifier.notifySearchUpdated(editor, res)
```
AlexPl292, line 1343: This is not a good place for running listeners. The `finItOffset` should be a clean function that returns the result of the search. If it's not clean, let's not make it worse at least.

AlexPl292:
Hi, thank you for your interest, this is a nice feature. I've added a related ticket.
I think this PR needs the following updates, but let's discuss if you have a different opinion.
- Let's disable this widget if the `shortmess` option has a `S` flag. This will match the Vim behaviour: https://stackoverflow.com/a/60840571. However, this looks fine to have this widget **enabled** by default. This will also match the neovim default behavior.
- Relying on highlighting is not okay. We should properly count the number of occurrences and show them in the widget, as the widget and highlighting are two orthogonal features of the search.
  - In the same way, removing the widget inside the `clearSearchHighlight` is incorrect.
- Also, in Vim the widget is disabled on the new search (when `/` is pressed), or when the other mode is entered. This doesn't happen here.

AlexPl292 (changes requested):

johngrib:
@jphalip Hi, I'm interested in the progress of this PR. I'm hoping to be able to use this feature. I'd like to know if you are currently working on any follow-up work.

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimSearchGroupBase.kt
clover256s: yes

AlexPl292:
Hi, I’m closing this PR due to inactivity. Feel free to reopen it or create a new one.

he44:
Hi @jphalip 👋 , thank you so much for implementing this feature! I recently switched to IntelliJ due to a work change, and this search widget is something I’ve been **desperately** wanting. Without the occurrence count, I’ve been awkwardly falling back to cmd + f 😢. I just pulled your branch locally and gave it a try—it’s great! 🙇

I noticed it's been a month since the last comment here, so I just wanted to gently check in and see if you had any plans for next steps. I'm very eager to see this feature shipped (especially since it’s tricky for me to use this branch directly on my work setup due to some very picky in-house plugin requirements). Please feel free to let me (or anyone really) know if we can be of any help if you have a tight schedule.

I also read through the discussion and agree that highlighting and occurrence count should be orthogonal. I came from the VSCode-Vim plugin world, where `set hls` does very little because VSCode handles its own highlighting, and the occurrence count is always visible. I did notice that currently, if I manually set nohls and restart IntelliJ, it might lead to a potential null exception.

<details>
<Summary> VSCode - Vim </Summary>

image

</details>

<details>
<Summary> Stacktrace when `set nohls` </Summary>

```
java.lang.NullPointerException
	at com.maddyhome.idea.vim.ui.widgets.search.SearchWidgetListener.searchUpdated(SearchWidgetFactory.kt:90)
	at com.maddyhome.idea.vim.common.VimListenersNotifier.notifySearchUpdated(VimListenersNotifier.kt:55)
	at com.maddyhome.idea.vim.api.VimSearchGroupBase.findItOffset(VimSearchGroupBase.kt:1351)
	at com.maddyhome.idea.vim.api.VimSearchGroupBase.processSearchCommand(VimSearchGroupBase.kt:419)
	at com.maddyhome.idea.vim.action.ex.ProcessSearchEntryAction.getOffset(ProcessExEntryActions.kt:68)
	at com.maddyhome.idea.vim.handler.MotionActionHandler.doExecuteForEach(MotionActionHandler.kt:204)
	at com.maddyhome.idea.vim.handler.MotionActionHandler.baseExecute(MotionActionHandler.kt:164)
```

</details>

(Another nit: I slightly prefer the VSCode style of "match xx of xx" over "occurrence xx of xx", just because it's shorter).
 
Thanks again for the awesome work!

PR #1078, opened 2025-01-01
Refactor Mode API and implement v_CTRL-O

citizenmatt:
This PR implements `v_CTRL-O`. That is, pressing `<C-O>` in Select mode will switch to Visual for the duration of a single command. This can be a motion, in which case we switch back to Select, or an operator, which usually removes the selection and so switches to Normal.

The implementation led to refactoring `Mode`, specifically how to represent the mode to return to. Adding a `ReturnTo.SELECT` singleton object would require updates to various `when` statements that need to convert `ReturnTo` to a `Mode`, and there are some transitions to Select that don't make sense, and would either require guessing the selection type, or throwing a runtime error. It is simpler and cleaner to just use `Mode` as the "return to" type.

This PR also updates the text in the "show mode" widget to show the pending modes, e.g. "(insert) NORMAL", "(replace) VISUAL" and so on.


AlexPl292:
Yay, this looks cool! Thank you!

PR #1079, opened 2025-01-02
Fix Del and BS not working in Select mode

citizenmatt:
This PR fixes an issue when hitting `<Del>` or `<BS>` to delete selected text in Select mode would fail to delete anything. The implementation would try to invoke native actions to delete but had two problems. Firstly, it would fail to filter out the IdeaVim action bound to the keystroke and would recurse (but fail to invoke anything). Secondly, it would also try to invoke instances of `EmptyAction`, which are placeholder actions that are used to reserve a shortcut. Code will use look for these actions by ID to look up the shortcut to use when dynamically registering the real action locally in a `Component`. We gather these locally registered actions, so it's safe to ignore the empty action.

It also correctly transitions from Select mode to Normal or Insert/Replace, depending on context. If Select mode was started from Normal (via `gh` or `v<C-G>`), then it will return to Normal. If started in Insert/Replace mode via shifted keys, then it will return to Insert/Replace. As part of this, it correctly positions the caret, especially when deleting the last text character in a line.

Fixes [VIM-3618](https://youtrack.jetbrains.com/issue/VIM-3618/Del-Backspace-in-select-Mode-not-working) and [VIM-3042](https://youtrack.jetbrains.com/issue/VIM-3042/Select-and-delete-from-insert-mode-moves-cursor-position).

~~This PR depends on #1078, specifically to handle the correct mode transitions. As such, it lists a load of unnecessary commits from that branch. Once #1078 is merged, I'll rebase, force push and mark this PR as no longer draft. Here is the list of commits in this PR: [272bea0...5275057](https://github.com/JetBrains/ideavim/compare/272bea02f2d40a9d259298f955c11fa46c477856...52750570d0459f24ecf5a9a7b3d58fcea3e08132)~~

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/group/KeyGroup.java
+++ b/src/main/java/com/maddyhome/idea/vim/group/KeyGroup.java
@@ -314,7 +314,7 @@ public class KeyGroup extends VimKeyGroupBase implements PersistentStateComponen
       if (c instanceof JComponent) {
         final List<AnAction> actions = ActionUtil.getActions((JComponent)c);
         for (AnAction action : actions) {
-          if (action instanceof VimShortcutKeyAction) {
+          if (action instanceof VimShortcutKeyAction || action == VimShortcutKeyAction.getInstance()) {
             continue;
           }
           final Shortcut[] shortcuts = action.getShortcutSet().getShortcuts();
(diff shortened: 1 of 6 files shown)


shaunpatterson:
Pulled this and tested.  
Fixes [VIM-3618](https://youtrack.jetbrains.com/issue/VIM-3618/Del-Backspace-in-select-Mode-not-working) and [VIM-3042](https://youtrack.jetbrains.com/issue/VIM-3042/Select-and-delete-from-insert-mode-moves-cursor-position).  Confirmed  (I wrote both tickets)

citizenmatt:
Rebased, and ready for review

shaunpatterson:
Tested again.  Thank you for fixing this.

AlexPl292:
You probably don't expect this information, but if you're interested, I've spent 1137.29 joules in IJ for reviewing this PR :)
I just found out about [joularjx](https://github.com/joular/joularjx) :D

citizenmatt:
Ha ha! When you mentioned joules I initially thought you meant something like calories, and was expecting you to have reviewed it on your treadmill or something! 😁

Agreed on the details about actions. It's not clear that the current implementation is correct, but it's not clear it's wrong either. Let's see what happens.

Now time to plan a PR that users even more energy!

PR #1080, opened 2025-01-04
Fix several minor search bugs relating to incsearch highlighting and Visual mode

citizenmatt:
This PR fixes a number of bugs related to `'incsearch'` highlighting while searching in Visual mode, e.g. `v/foo`

- [x] Fixes an issue where `'incsearch'` would not correctly update an empty selection, when `'selection'` was set to "exclusive". So something like `ve/foo` would work but `v/foo` would not update the selection. The selection would be correctly set when completing the search with `<CR>`
- [x] This also fixed an issue with correctly updating the selection when the search pattern was changed, such as editing it from `/foo` to `/faa`
- [x] Don't show `'incsearch'` highlighting when `'nohlsearch'` is set and Visual mode is active. Normally, `'incsearch'` highlighting is shown for the current match, even when `'hlsearch'` is disabled, but it's not shown if Visual mode is active - the selection is enough
- [x] Exit Visual mode when `'incsearch'` needs to move the caret for a Command-line command, rather than a search. E.g. something like `ve:<C-U>normal 3j` should not maintain the Visual selection, but `ve/foo` should

It also removes some obsolete tests using the no longer used `'usenewregex'` option, and splits up the search related tests into `SearchGroupTest.kt`, `IncsearchTests.kt`, `SearchHighlightsTest.kt`, etc.

~~This PR depends on the `Mode` refactoring in #1078 and includes all of those commits. To make it easier for me to manage, it's also based on #1079. Once these PRs are merged, I will rebase and force push and mark this PR as no longer draft. Here are the lists of commits for this PR: [faeaafd2...f229c5d0](https://github.com/JetBrains/ideavim/compare/faeaafd25fec9a330fdd1155cc06933f0c724e5d...f229c5d0662ec1b89ed64f78de738fb2eb8d1383)~~


citizenmatt:
Rebased and ready for review

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/state/mode/editorExtensions.kt
```diff
@@ -26,6 +26,9 @@ val VimEditor.inSelectMode: Boolean
 val VimEditor.inCommandLineMode: Boolean
   get() = this.mode is Mode.CMD_LINE
 
+val VimEditor.inCommandLineModeWithVisual: Boolean
```
AlexPl292, line 29: Can you add a bit of docs explaining what this value should return? I mean, it's clear from the implementation, but let's explicitly state the intention of this value.
For example, if this function checks that "Mode is command" & "return is visual" – everything is ok. But if it means that return is visual for current command line mode, this function requires "inCommandLineMode" check before calling it, and an exception in case the mode is not CMD.

AlexPl292 (approved):
Looks good for me. Can you add a small doc for my comment?

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/state/mode/editorExtensions.kt
citizenmatt: Done!

AlexPl292:
Thank you!

PR #1081, opened 2025-01-08
Update MotionQuoteAction.kt to support iq and aq motions

oca159:
Hi everyone,

I’d like to propose a new motion for editing text inside or around any type of quotes: ', ", or `.

The idea is to introduce motions like ciq, caq, diq, and daq to streamline these actions and make them more intuitive.

What do you think? Feedback is welcome! 😊

Notes:
I don't have experience with kotlin, so let me know if there is something wrong with my code change.

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/listener/IJEditorFocusListener.kt
+++ b/src/main/java/com/maddyhome/idea/vim/listener/IJEditorFocusListener.kt
@@ -80,7 +80,9 @@ class IJEditorFocusListener : EditorListener {
       if (ijEditor.isDisposed) return@invokeLater
       val consoleView: ConsoleViewImpl? = ijEditor.getUserData(ConsoleViewImpl.CONSOLE_VIEW_IN_EDITOR_VIEW)
       if (consoleView != null && consoleView.isRunning && !ijEditor.inInsertMode) {
+        // Switch to Insert mode, but make sure we reset the editor to actually make it apply
         switchToInsertMode.run()
+        KeyHandler.getInstance().reset(editor)
       }
     }
     KeyHandler.getInstance().reset(editor)


AlexPl292 (changes requested):
Comments are above.

oca159:
Hi @AlexPl292! 

Yes, I was exploring the functionality of this plugin: [mini.ai](https://github.com/echasnovski/mini.ai). It allows me to use motions like ciq, caq, diq, and daq to work with any type of quotes, which I find incredibly useful.

It also supports motions like cib, cab, dib, and dab for working with parentheses, curly braces, and square brackets. I understand this might be a bit unconventional since the standard is b for parentheses and B for curly braces, but I find it super comfortable and intuitive to use.

I'll check out the JetBrains plugin documentation since I currently have no experience implementing something similar as a plugin. Thanks for pointing me in the right direction!

Iamlooker:
Just a small pet peeve, Use of `charArrayOf()` would be just slightly better than list for cache locality (not very important for such small list). Mainly there is an object allocation with using `Pair<Range, Quote>` while we are not using Quote anywhere. Sorry if the PR is not complete 😅

oca159:
@Iamlooker, thanks for the feedback! 😊 I might consider creating a plugin to implement this feature in the future. I just need to learn Java or Kotlin first since I'm not experienced with them yet, haha

Iamlooker:
Not at all, what you did is really awesome keep up

citizenmatt:
@oca159 you don't need to create a separate plugin. You could add this to the Extensions folder, unless @AlexPl292 would prefer to see it external?

AlexPl292:
Yes, we can put it in IdeaVim repository, but as an extension, not as a core functionality.
The `extensions` folder is this one: https://github.com/JetBrains/ideavim/tree/master/src/main/java/com/maddyhome/idea/vim/extension

oca159:
new pr: https://github.com/JetBrains/ideavim/pull/1087

PR #1084, opened 2025-01-14
Fix race condition resetting editor on focus gain

citizenmatt:
When an editable console editor gets focus, IdeaVim automatically enters Insert mode. However, we can't always tell a console editor window is editable when it gains focus, so we do an additional check via `application.invokeLater`. If this switches to Insert mode, it doesn't correctly reset the (newly focused console) editor, so even though the mode is "Insert", it was last reset with "Normal". This PR will reset the editor state (including key handler) so that the new mode is "applied".

Fixes [VIM-3773](https://youtrack.jetbrains.com/issue/VIM-3773)


AlexPl292:
Thank you!

PR #1090, opened 2025-01-29
fix: distanceTo to works with nested quotes and brackets (mini.ai) plugin

oca159:
### Why?
The mini.ai extension had an issue with nested quotes or brackets selection.

### The Problem
When dealing with nested quotes, the extension should select the innermost quotes containing the cursor. However, it was incorrectly selecting outer quotes due to the distance calculation logic.

### Example:

current state: This is a "simple '|example'" that works

After pressing ciq it should produce the following result:

expected: This is a "simple '|'" that works

However it is behaving like this:

actual: This is a "|" that works

### The Solution
The distanceTo function has been refactored to prioritize:

- Ranges that contain the cursor
- Smaller ranges over larger ones
- Heavily penalize ranges where the cursor is outside

I noticed that the distanceTo function was wrong, this pr fixes it.

Please take a look at the new tests to understand the problem.

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/extension/miniai/MiniAI.kt
+++ b/src/main/java/com/maddyhome/idea/vim/extension/miniai/MiniAI.kt
@@ -220,11 +220,16 @@ private fun findClosestQuoteRange(
 }
 
 private fun TextRange.distanceTo(caretOffset: Int): Int {
-  return if (caretOffset < startOffset) {
-    startOffset - caretOffset
-  } else if (caretOffset > endOffset) {
-    caretOffset - endOffset
-  } else {
-    0 // Caret is inside the range
+  val rangeLength = endOffset - startOffset
+
+  // If caret is inside the range
+  if (caretOffset in startOffset..endOffset) {
+    // Return the length of the range - smaller ranges get priority
+    return rangeLength
   }
+
+  // If caret is outside, make the distance much larger
+  val startDistance = kotlin.math.abs(caretOffset - startOffset)
+  val endDistance = kotlin.math.abs(caretOffset - endOffset)
+  return (startDistance + endDistance) * 1000 + rangeLength
 }
--- a/src/main/java/com/maddyhome/idea/vim/extension/miniai/MiniAI.kt
+++ b/src/main/java/com/maddyhome/idea/vim/extension/miniai/MiniAI.kt
@@ -194,42 +194,60 @@ private fun addAction(action: TextObjectActionHandler) {
   keyHandlerState.commandBuilder.addAction(action)
 }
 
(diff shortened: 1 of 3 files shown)


oca159:
@AlexPl292, please take a look at it, and I apologize for the inconvenience. I’ve added more tests to validate the new logic, including cases with nested quotes and brackets. It now works perfectly. I’ll make sure to include more comprehensive tests in the future to prevent this situation from happening again.

AlexPl292:
Hi there! No worries, and thanks for the detailed explanation of the issue!

I have a few questions about the fix. To be honest, I find the logic of “increasing” the distance for the outer range a bit unintuitive and potentially error-prone. What do you think about rewriting this function to explicitly calculate the ranges that cover the caret and those that are outside of it, then selecting the appropriate range based on some clear logic?

Also, I believe the following test case isn’t working as expected due to this logic. After all the calculations, the `"` range ends up as 30008, while the `'` range becomes 30006 and gets priority. However, my understanding is that we should be replacing the `"` range in this case.

What do you think?

By the way, since you’ve already modified the code, could you also move the tests into the main IdeaVim module, alongside other tests like `VimIndentObjectTest`? From what I see, your tests don’t rely on any Java-specific functionality, so they don’t need to be in the `java-tests` module.

```
  @Test
  fun testFromAlex() {
    doTest(
      "ciq",
      "<caret>This is a \"'simple'\" test",
      "This is a \"<caret>\" test",
      Mode.NORMAL(),
      JavaFileType.INSTANCE,
    )
    assertSelection(null)
  }
```

oca159:
Hi @AlexPl292,

I’ve added your test and made the necessary updates. Just a quick note: after running ciq, the mode should be Insert instead of Normal, I updated your test example in my code to accordingly. I’ve also refined the logic based on your feedback, and it seems to be working well now.

Additionally, I moved the test to its correct location. Could you please review it when you have a moment and let me know your thoughts? Everything should be functioning as expected now.

Thanks!

oca159:
Hi @AlexPl292,

I’ve added your test and made the necessary updates. Just a quick note: after running ciq, the mode should be Insert instead of Normal. I’ve also refined the logic based on your feedback, and it seems to be working well now.

Additionally, I moved the test to its correct location. Could you please review it when you have a moment and let me know your thoughts? Everything should be functioning as expected now.

Thanks!

oca159:
I’ve also added some additional tests to ensure that the order of quotes doesn’t affect the behavior and that all quotes are treated with equal preference when running motions like ciq. The key factor should be the distance to the next quote, as that’s what determines the outcome.

AlexPl292:
Looks cool, thank you!

PR #1094, opened 2025-01-30
Add extension: vim-searchindex

johngrib:
I have implemented [google/vim-searchindex](https://github.com/google/vim-searchindex) plugin.
vim-searchindex is a small plugin that shows the total number of search matches and the current match index when searching with `/`, `?`, `*`, `#`, `g*`, `g#` or `n`, `N`.

You can enable the plugin with the following command:

```
:set searchindex
```

| vim | screenshot  |
| --- | --- |
| searchindex on vim | image | 
| searchindex on ideavim | screenshot 2025-01-30 22 33 06 |

Please let me know if any of my code violates the conventions of IdeaVim or IDEA.

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/extension/searchindex/VimSearchIndexExtension.kt
+++ b/src/main/java/com/maddyhome/idea/vim/extension/searchindex/VimSearchIndexExtension.kt
@@ -1,0 +1,158 @@
+/*
+ * Copyright 2003-2025 The IdeaVim authors
+ *
+ * Use of this source code is governed by an MIT-style
+ * license that can be found in the LICENSE.txt file or at
+ * https://opensource.org/licenses/MIT.
+ */
+package com.maddyhome.idea.vim.extension.searchindex
+
+import com.intellij.openapi.application.ApplicationManager
+import com.intellij.openapi.diagnostic.logger
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.project.ProjectManager
+import com.intellij.openapi.wm.StatusBar
+import com.intellij.openapi.wm.StatusBarWidget
+import com.intellij.openapi.wm.StatusBarWidgetFactory
+import com.intellij.openapi.wm.WindowManager
+import com.intellij.openapi.wm.impl.status.EditorBasedWidget
+import com.intellij.util.Consumer
+import com.maddyhome.idea.vim.VimPlugin
+import com.maddyhome.idea.vim.api.VimEditor
+import com.maddyhome.idea.vim.api.injector
+import com.maddyhome.idea.vim.common.ModeChangeListener
+import com.maddyhome.idea.vim.extension.VimExtension
+import com.maddyhome.idea.vim.newapi.ij
+import com.maddyhome.idea.vim.state.mode.Mode
+import com.maddyhome.idea.vim.ui.ex.ExEntryPanel
+import java.awt.Component
+import java.awt.event.MouseEvent
+import java.util.regex.Pattern
+
+internal object SearchIndex {
+  internal const val ID = "IdeaVimSearchIndex"
+  internal const val DISPLAY_NAME = "IdeaVim searchindex"
+  private var currentText = ""
+
+  fun update(text: String) {
+    currentText = text
+    val windowManager = WindowManager.getInstance()
+    ProjectManager.getInstance().openProjects.forEach {
+      val statusBar = windowManager.getStatusBar(it)
+      statusBar.updateWidget(ID)
+    }
+  }
+
+  fun getCurrentText(): String = currentText
+}
+
+internal class SearchIndexWidgetFactory : StatusBarWidgetFactory {
+  override fun getId() = SearchIndex.ID
+  override fun getDisplayName() = SearchIndex.DISPLAY_NAME
+  override fun createWidget(project: Project) = SearchIndexWidget(project)
+  override fun disposeWidget(widget: StatusBarWidget) {}
+  override fun canBeEnabledOn(statusBar: StatusBar) = true
+  override fun isAvailable(project: Project) = true
+}
+
+internal class SearchIndexWidget(project: Project) :
+  EditorBasedWidget(project),
+  StatusBarWidget.TextPresentation {
+
+  override fun ID() = SearchIndex.ID
+  override fun getPresentation() = this
+  override fun getClickConsumer(): Consumer<MouseEvent>? = null
+  override fun getText() = SearchIndex.getCurrentText()
+  override fun getTooltipText() = null
+  override fun getAlignment() = Component.RIGHT_ALIGNMENT
+}
+
+internal class VimSearchIndexExtension : VimExtension, ModeChangeListener {
+  init {
+    LOG.info("Looking at SearchGroup implementation: ${VimPlugin.getSearch().javaClass}")
+  }
+
+  companion object {
+    private val LOG = logger<VimSearchIndexExtension>()
+  }
+
+  override fun getName(): String = "searchindex"
+
+  override fun init() {
+    LOG.info("Initializing SearchIndex extension")
+    injector.listenersNotifier.modeChangeListeners.add(this)
+  }
+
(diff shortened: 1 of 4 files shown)


src/main/resources/META-INF/plugin.xml
```diff
@@ -65,6 +65,7 @@
     <statusBarWidgetFactory id="IdeaVim-Icon" implementation="com.maddyhome.idea.vim.ui.StatusBarIconFactory" order="last, before IdeaVimMode"/>
     <statusBarWidgetFactory id="IdeaVimShowCmd" implementation="com.maddyhome.idea.vim.ui.ShowCmdStatusBarWidgetFactory" order="first"/>
     <statusBarWidgetFactory id="IdeaVimMacro" implementation="com.maddyhome.idea.vim.ui.widgets.macro.MacroWidgetFactory" order="first, after IdeaVimShowCmd"/>
+    <statusBarWidgetFactory id="IdeaVimSearchIndex" implementation="com.maddyhome.idea.vim.extension.searchindex.SearchIndexWidgetFactory" order="after IdeaVimShowCmd" />
```
johngrib, line 68: Could you advise if this is an appropriate position for the widget? I'm not entirely sure about the best placement in the status bar hierarchy.

src/main/java/com/maddyhome/idea/vim/extension/searchindex/VimSearchIndexExtension.kt
```diff
@@ -0,0 +1,183 @@
...
+
+  override fun dispose() {
+    LOG.info("Disposing SearchIndex extension")
+    injector.listenersNotifier.modeChangeListeners.remove(this)
+  }
+
+  override fun modeChanged(editor: VimEditor, oldMode: Mode) {
+    LOG.info("Mode changed from $oldMode to ${editor.mode}")
+
+    if (oldMode is Mode.CMD_LINE && editor.mode !is Mode.CMD_LINE) {
```
johngrib, line 172: I ended up using modeChanged since I couldn't find a dedicated search event listener. If there's a proper search event that I missed, please let me know.

src/main/java/com/maddyhome/idea/vim/extension/searchindex/VimSearchIndexExtension.kt
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    private fun getCurrentIndex(
+      editor: Editor,
+      results: List<TextRange>,
+    ): Int {
+      val caretOffset = editor.caretModel.offset
+      var current = 0
+      results.forEachIndexed { index, range ->
+        if (range.startOffset <= caretOffset && caretOffset <= range.endOffset) {
```
AlexPl292, line 138: We can exit the loop when the correct index is found.

src/main/java/com/maddyhome/idea/vim/extension/searchindex/VimSearchIndexExtension.kt
```diff
@@ -0,0 +1,183 @@
...
+  fun getCurrentText(): String = currentText
+}
+
+internal class SearchIndexWidgetFactory : StatusBarWidgetFactory {
+  override fun getId() = SearchIndex.ID
+  override fun getDisplayName() = SearchIndex.DISPLAY_NAME
+  override fun createWidget(project: Project) = SearchIndexWidget(project)
+  override fun disposeWidget(widget: StatusBarWidget) {}
+  override fun canBeEnabledOn(statusBar: StatusBar) = true
+  override fun isAvailable(project: Project) = true
```
AlexPl292, line 65: Since this is an extension, I'd say that this widget should be available only in case the extension is enabled.

src/main/java/com/maddyhome/idea/vim/extension/searchindex/VimSearchIndexExtension.kt
```diff
@@ -0,0 +1,183 @@
...
+import com.maddyhome.idea.vim.state.mode.Mode
+import com.maddyhome.idea.vim.ui.ex.ExEntryPanel
+import java.awt.Component
+import java.awt.event.MouseEvent
+import javax.swing.KeyStroke
+
+internal object SearchIndex {
+  internal const val ID = "IdeaVimSearchIndex"
+  internal const val DISPLAY_NAME = "IdeaVim searchindex"
+  private var currentText = ""
```
AlexPl292, line 45: Since the search happens in a specific project, we should not use a single "currentText" property that is common for all projects.
Let's create a project-level service that stores the value of the "currentText" per-project.

src/main/java/com/maddyhome/idea/vim/extension/searchindex/VimSearchIndexExtension.kt
```diff
@@ -0,0 +1,183 @@
...
+    }
+  }
+
+  override fun getName(): String = "searchindex"
+
+  override fun init() {
+    LOG.info("Initializing SearchIndex extension")
+    injector.listenersNotifier.modeChangeListeners.add(this)
+
+    listOf("n", "N", "*", "#", "g*", "g#").forEach {
```
AlexPl292, line 152: This approach looks hacky and doesn't support mappings. I see, that there is also an issue that `modeChanged` also looks more like a hack.
If IdeaVim doesn't have a proper listener for search, I think it's okay to add one. It's better to have a custom listener that hacky implementation.

src/main/java/com/maddyhome/idea/vim/extension/searchindex/VimSearchIndexExtension.kt
```diff
@@ -0,0 +1,183 @@
...
+        }
+      }
+      return current
+    }
+  }
+
+  override fun getName(): String = "searchindex"
+
+  override fun init() {
+    LOG.info("Initializing SearchIndex extension")
```
AlexPl292, line 149: The INFO level of logging is too verbose and will clutter the log file. Let’s change it to “DEBUG.”

But I appreciate you considering it!

src/main/java/com/maddyhome/idea/vim/extension/searchindex/VimSearchIndexExtension.kt
```diff
@@ -0,0 +1,183 @@
...
+        return
+      }
+
+      LOG.info("search index - Processing pattern: $pattern")
+
+      // FIXME: global search option -> previous search option
+      val ignoreCase = injector.options(editor = editor.vim).ignorecase
+      LOG.info("Ignore case: $ignoreCase")
+
+      val results = injector.searchHelper.findAll(
```
AlexPl292, line 110: Searching twice on each `n` key looks like a performance issue. I left a comment about the new search listener. I think, it may also send the information about the number of found results.

src/main/resources/META-INF/plugin.xml
AlexPl292: We don't have any rules about it now, so I think it's okay to put it there.

AlexPl292 (changes requested):
Hey there, nice to see you again!

First of all, we have a PR with a similar functionality on reiview. However, it has a different way of implementing it. Also, note that if the mentioned PR will be merged, must probably we'll turn that option ON by default.

Regarding your PR, I left a few comments. Also, I noticed that the original Vim plugin hides the search information after entering VISUAL/INSERT mode. In this implementation, the information remains.

AlexPl292:
Oh, I didn’t realize that vim-searchindex _suggests_ using the native functionality.

In that case, it would be better to have this as a core feature rather than an extension.

@johngrib What do you think about collaborating on https://github.com/JetBrains/ideavim/pull/1054 to create a proper implementation?

johngrib:
I hastily submitted this PR simply because I wanted to make vim-searchindex available in ideavim. I wasn't aware of `set shortmess-=S` at the time. 
After reading @citizenmatt's comment, I tested vim and neovim today with just `shortmess-=S` set and no other settings, and found they work exactly like vim-searchindex. 
So I also think it would be more appropriate as a core feature rather than an extension. I agree with @AlexPl292's suggestion to improve #1054.

PR #1095, opened 2025-02-03
feat: improve mini.ai extension behavior to be exactly as the original plugin for complex scenarios.

oca159:
### Overview
This PR modifies the mini.ai‐like text object logic to prioritize brackets or quotes on the same line as the caret before considering the rest of the buffer. It addresses scenarios like the unit test `testNextBracketInCurrentLineHasPriority`, where the user expects “next bracket on the same line” to be chosen over a larger multi‐line bracket pair.

### What changed
1. Two‐pass approach for finding text objects:
- First pass: Gather and choose from bracket/quote pairs that occur on the caret’s line (using the same mini.ai “cover‐or‐next” logic).
- Second pass: If no suitable text object is found on that line, we then gather pairs from the entire buffer and pick again with the “cover‐or‐next” approach.
2. Inner vs. Outer handling remains the same:

- If the user requests an “inner” text object (e.g., ib or iq), we still trim away the surrounding characters (brackets or quotes) from the final range.
3. Updated test:

- The test testNextBracketInCurrentLineHasPriority now passes because the engine will pick parentheses (or bracket pairs) on the same line before expanding to multi‐line brackets.

### Why this matters
Some users and tests require that if there is a nearby bracket pair on the same line, it takes precedence over a larger pair spanning multiple lines. This is a common expectation in code editing: if the cursor is inside parentheses in the same line, we don’t want a big curly brace range to override it.

By implementing the two‐stage approach (“line first, then entire file”), we ensure that any local pairs on the caret’s line get chosen first, aligning with user expectations (and ensuring existing tests pass).

### How to verify
1. Run the existing bracket/quote text object tests, particularly testNextBracketInCurrentLineHasPriority.
2. Optionally, try manual scenarios where a small bracket or quote range is on the same line, while a larger multi‐line bracket also encloses the caret. Confirm that the small (same‐line) range is chosen by default.

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/extension/miniai/MiniAI.kt
+++ b/src/main/java/com/maddyhome/idea/vim/extension/miniai/MiniAI.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2003-2023 The IdeaVim authors
+ * Copyright 2003-2025 The IdeaVim authors
  *
  * Use of this source code is governed by an MIT-style
  * license that can be found in the LICENSE.txt file or at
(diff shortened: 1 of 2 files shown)


oca159:
Hi @AlexPl292,

I’ve been working on a new version of the mini.ai extension to more closely match the behavior of the original plugin for both quotes and brackets. It took me longer than expected, but now it behaves exactly as the original mini.ai does in all test scenarios—especially the tricky ones involving “false” strings and multi-line brackets.

### New Tests and Behaviors
1. False string in the middle
Previously, using `cib` in a situation like:

`"first" false | string "second"`

would improperly merge both quoted sections into something like:

`"first"|"second"`

Now, just like the real mini.ai plugin, the “false” string in the middle is correctly ignored, resulting in:

`"first" false string "|"`

This behavior aligns with what the original mini.ai does, avoiding accidental merging of unrelated quotes.

2. Cover‐or‐Next Algorithm
The plugin now uses the “cover‐or‐next” approach for brackets and quotes, first checking if the caret is covered by a bracket or quote on the same line and only then expanding the search to surrounding lines. For example, previously:

```
{
    |print("hello")
}
```

pressing `cib` would delete everything inside the curly braces, leaving:

```
{
    |
}
```

which could be quite frustrating. Now, consistent with the original mini.ai, it only targets the parentheses around "hello":

```
{
    print(|)
}
```

This is much more intuitive when working with nested brackets on multiple lines.

### Summary
I’ve introduced new helper functions to achieve these behaviors and added test cases illustrating scenarios where the old implementation fell short. Please take a look at the new tests to see how each scenario is handled. I believe this new logic provides a much more seamless and predictable experience, matching the original mini.ai plugin’s design.

Let me know what you think or if you have any questions!

Thanks!

AlexPl292:
Thank you for the improvements and the overall quality of the code and PR description. This really helps us with the review process.

Starting in February, we will launch the [Awards for Quality Contributions](https://github.com/JetBrains/ideavim/blob/master/CONTRIBUTING.md#awards-for-quality-contributions). If you’re interested in receiving a one-year All Products Pack subscription for this contribution, please let me know in the comments or via email at alex.plate@ideavim.dev. I’ll need an email address you have access to in order to send the certificate.

PR #1097, opened 2025-02-11
The CommentaryTextObject in commentary no longer contains spaces.

CCCC-L:
CommentaryTextObject no longer contains spaces to be consistent with neovim

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/extension/commentary/CommentaryExtension.kt
+++ b/src/main/java/com/maddyhome/idea/vim/extension/commentary/CommentaryExtension.kt
@@ -224,14 +224,16 @@ internal class CommentaryExtension : VimExtension {
       val endOffset = editor.vim.getLineEndOffset(logicalLine, true)
       val startElement = file.findElementAt(startOffset) ?: return false
       var next: PsiElement? = startElement
+      var hasComment = false
       while (next != null && next.textRange.startOffset <= endOffset) {
-        if (next !is PsiWhiteSpace && !isComment(next)) {
-          return false
+        when {
+          next is PsiWhiteSpace -> {} // Skip whitespace elementl
+          isComment(next) -> hasComment = true // Mark when we find a comment
+          else -> return false // Non-comment content found, exit early
         }
         next = PsiTreeUtil.nextLeaf(next, true)
       }
-
-      return true
+      return hasComment
     }
 
     private fun isComment(element: PsiElement) =


AlexPl292:
Hi, thank you for your contribution! Can you please add some tests for your case in `CommentaryExtensionTest`?
Also, take a look at the TeamCity failed build, as it contains some failed tests from this class.

CCCC-L:
> Hi, thank you for your contribution! Can you please add some tests for your case in `CommentaryExtensionTest`? Also, take a look at the TeamCity failed build, as it contains some failed tests from this class.



> Hi, thank you for your contribution! Can you please add some tests for your case in `CommentaryExtensionTest`? Also, take a look at the TeamCity failed build, as it contains some failed tests from this class.

Since blank lines are no longer regarded as comments, the following several test methods cannot pass. Can I modify the contents of both "Expected" and "Actual"?

```
test text object deletes single line comment from leading whitespace 2
test text object deletes single line comment from leading whitespace 3
test text object deletes adjoining multi-line block comments 2
test text object deletes JavaDoc comment from leading whitespace
test text object deletes single line comments separated by whitespace
test text object deletes JavaDoc comment and adjoining comments separated by whitespace
test text object deletes single line comment from trailing whitespace
test text object deletes disjointed single line comments from whitespace
```

AlexPl292:
This depends if the new behavior matches the behavior of the vim-commentary extension. You can install the plugin on Vim and check if it works the same.
I've checked a few tests, and it looks like the current behavior in tests matches the Vim. For example, `test text object deletes single line comment from trailing whitespace` is correct. Also, vim seems not to handle block comments, but if they'll be replaced with a simple `//` comment `test text object deletes JavaDoc comment from leading whitespace` also seems to have a correct processing.

CCCC-L:
> This depends if the new behavior matches the behavior of the vim-commentary extension. You can install the plugin on Vim and check if it works the same. I've checked a few tests, and it looks like the current behavior in tests matches the Vim. For example, `test text object deletes single line comment from trailing whitespace` is correct. Also, vim seems not to handle block comments, but if they'll be replaced with a simple `//` comment `test text object deletes JavaDoc comment from leading whitespace` also seems to have a correct processing.

Yes, in the vim-commentary plugin, the empty lines above and below the comment are also considered as part of the comment. However, in neoVim, this is not the case. Neovim does not consider the empty lines above and below the comment as part of the comment.

For example, in the following text, if the cursor is on the first line, when you press `dgc`, it will not delete both comments. Instead, it will only delete `// commentary1`.
```java
// commentary1

// commentary2
```

AlexPl292:
Okay, I see what you mean.
For us it looks like the neovim approach is more expected. Let's make it like in neovim and update the tests accordingly.

AlexPl292:
Looks good to me, thank you!

PR #1107, opened 2025-02-21
Add missing shortcuts for WORD left/right in Select

citizenmatt:
This PR adds `<C-Left>` and `<C-Right>` actions in Select mode to move to the previous/next WORD. It adds new actions, similar to the existing actions for `<C-Left>` and `<C-Right>` for Visual mode. They're separate actions because the Visual mode actions are also mapped to the `B` shortcut, which isn't valid for Select.

Note that the tests don't all follow Vim behaviour. Firstly, there is a separate bug/mis-implementation that entering select with e.g. `gh` will move the caret one character to the right in IdeaVim, but not in Vim. This will be addressed in a separate PR that does a bunch of work around Select mode, inclusive/exclusive selection, off-by-one errors and so on. Secondly, IdeaVim treats Select mode as always being "exclusive", as if `:set selection=exclusive` has been called. This alters the selection rules and causes differences in behaviour. This is less likely to be fixed as it's much more intuitive behaviour.


AlexPl292:
Yep, thank you!

PR #1118, opened 2025-02-27
Make yank motion respect exclusive-linewise

nath:
I used the code in VimChangeGroupBase.getDeleteRangeAndType to do the check.

For an example of what this fixes, with the document:
```
    ${c}Hello!
```
executing "y}llp" before this change yields
```
    HelHello!
    lo!
```
but doing it with this change yields
```
    Hello!
    Hello!
````

### DIFF
--- a/src/test/java/org/jetbrains/plugins/ideavim/action/copy/YankMotionActionTest.kt
+++ b/src/test/java/org/jetbrains/plugins/ideavim/action/copy/YankMotionActionTest.kt
@@ -10,6 +10,7 @@ package org.jetbrains.plugins.ideavim.action.copy
 
 import com.maddyhome.idea.vim.api.injector
 import com.maddyhome.idea.vim.newapi.vim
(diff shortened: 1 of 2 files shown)


nath:
There has to be a new line after the text, just the line `    Hello!` where you do that will lead to `HeHello!llo!` as you mentioned, but if there's a blank line below the `}` can go to it will become linewise.

From :help exclusive-linewise "If the motion is exclusive, **the end of the motion is in column 1** and the start of the motion was at or before the first non-blank in the line," so that's what's missing without the new line.

I think this commit should check that bit correctly with the `&& !editor.anyNonWhitespace(motionRange.endOffset, 1)`

nath:
@AlexPl292 Hey, just wanted to get your thoughts on the above. This behavior is already implemented in ideavim for deletes, which is where I pulled the logic from, this just adds it to yanks.

PR #1123, opened 2025-03-10
Move status bar widgets after dotCover indicator

citizenmatt:
Rider's dotCover indicator is marked as "last" rather than relative to other dotnet icons. This position pushes out IdeaVim widgets, so make sure we're after "dotCoverIndicator".

See also [DCVR-13021](https://youtrack.jetbrains.com/issue/DCVR-13021/dotCover-status-bar-indicator-in-Rider-incorrect-position)

Screenshot 2025-03-10 at 6 03 43 pm
Screenshot 2025-03-10 at 6 00 15 pm


AlexPl292:
Sweet, thank you!

PR #1126, opened 2025-03-17
Fix caret location after :move command

citizenmatt:
Fixes the caret location after the `:move` command, especially with regard to relative addresses, such as `:+2m.`. Fixes [VIM-3837](https://youtrack.jetbrains.com/issue/VIM-3837).

Also makes some minor changes to clarify some behaviour that appeared related but wasn't. Namely, that `line('v')` returns the opposite end of the current selection (related to [VIM-3838](https://youtrack.jetbrains.com/issue/VIM-3838))

### DIFF
--- a/src/test/java/org/jetbrains/plugins/ideavim/ex/implementation/commands/MoveCommandTest.kt
+++ b/src/test/java/org/jetbrains/plugins/ideavim/ex/implementation/commands/MoveCommandTest.kt
@@ -148,7 +148,7 @@ class MoveCommandTest : VimTestCase() {
     enterCommand("m 0")
     assertState(
       """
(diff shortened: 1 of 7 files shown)


AlexPl292:
Hi. I've checked the behavior in Vim and Neovim regarding the caret location after the move, and it turned out that Neovim keeps the column of the caret after the move. This is how it works in IdeaVim now, and I'd say this behavior is more intuitive. What do you think?

citizenmatt:
This is due to the default values of the `'startofline'` option. Vim has it default to on, while Neovim defaults it to off (see [neovim/neovim#11135](https://github.com/neovim/neovim/pull/11135)). This PR will position the caret in the column correctly, based on the value of this setting.

PR #1132, opened 2025-03-20
Fix unhandled key stroke after surround operation

citizenmatt:
Fixes [VIM-3841](https://youtrack.jetbrains.com/issue/VIM-3841/Unhandled-key-press-after-surround-operation).

Given a command such as `veS"`, IdeaVim will enter Visual mode, surround the selection with `"` and exit Visual. A subsequent keystroke such as `i` has no effect.

This is because the key handler is not reset when leaving Visual mode, and continues to try and process keystrokes as Visual. This PR resets the key handler.

See discussion at #1124.


citizenmatt:
We should (at some point) update extension mappings to properly reset the command builder and key handler when the mapping is executed, like we do with regular mappings.

AlexPl292:
Thank you!

PR #1138, opened 2025-03-25
Support 32-bit Unicode codepoints in digraphs

citizenmatt:
This PR changes the way custom digraphs are stored from 16-bit `Char` instances to 32-bit integer codepoints. This means we can support digraphs such as `digraph cr 128308 ", which will generate: 🔴

It supports creating and using these digraphs, as well as listing them in output to `:digraph`, and in the output to `:ascii`.

Fixes [VIM-3842](https://youtrack.jetbrains.com/issue/VIM-3842)


AlexPl292:
Sweet, merged!

PR #1139, opened 2025-03-25
Expand `~` to home directory in XDG_CONFIG_HOME

citizenmatt:
The `XDG_CONFIG_HOME` variable should have a default value of `$HOME/.config`, but if it's set by a user, it might contain a `~`, e.g. `~/dotfiles/config`. This is probably incorrect, although this isn't stated explicitly in the docs - only that the default value should be `$HOME/.config`, which when set in a shell, will expand to a full path. 

The `~` is only recognised by shells, and expanded before being passed to a process, so an incorrect `XDG_CONFIG_PATH` variable will likely work in many cases. However, the JVM does not automatically expand `~`, so an environment variable in this form will fail when used with IdeaVim.

Since it's not obvious that the environment variable is likely incorrect, this PR will update IdeaVim to expand the `~` to the user's home directory before loading `$XDG_CONFIG_HOME/ideavim/ideavimrc`

Addresses [VIM-3844](https://youtrack.jetbrains.com/issue/VIM-3844), [VIM-2001](https://youtrack.jetbrains.com/issue/VIM-2001)


AlexPl292:
Thank you!

PR #1153, opened 2025-04-08
Refactor mapping to better handle replaying failed mappings

citizenmatt:
This PR refactors `MappingProcessor`, primarily to remove the need for the `mappingComplete` parameter, which is required for a forthcoming PR. However, a bigger impact is on the way unhandled keys are processed after a mapping fails.

- [x] Extract and centralise a single implementation to replay unhandled keys when a mapping sequence no longer matches a prefix or the `'timeout'` expires
- [x] Fix the implementation of replaying unhandled keys to find the longest mapping LHS in the abandoned sequence, instead of just looking at the sequence up to the previous character. E.g. given `map ab foo`, `map abcde bar` and an input of `abcdz`, the mapping processor will now execute the `ab` mapping (producing `foo`) and then try to handle `cdz` separately.
- [x] Remove the `mappingComplete` parameter from `KeyConsumer`. Because of the above change, it is no longer necessary to identify the last character in a sequence to stop further mapping. 
- [x] Remove the special case handling of `<Plug>` when processing unhandled mappings. Previously, we would discard all characters for a sequence that began with `<Plug>`. It's unclear why this was done, possibly because replaying the keys in Normal mode would rarely result in what the user expected. However, this is Vim behaviour, and IdeaVim has been updated to match.
- [x] Output `"<Plug>"` and `"<Action>"` as text when replaying unhandled `<Plug>` and `<Action>` maps in Insert mode. Like in Vim, the `<Plug>` key is a special keystroke that can't be typed (`<Action>` is the same). When replaying a `<Plug>` or `<Action>` mapping prefix in Insert mode, the special keystroke is converted to plain text, again following Vim behaviour.
- [x] Uncomment tests for the `:normal` command. Not sure why they were commented, but they work
- [x] Removed special case handling of modes in the `:normal` command implementation, as well as removing the `SAVE_VISUAL` command flag. This fixes an issue with an incorrect caret location when invoking `:normal` from Visual. IdeaVim (like Vim) always switches to Normal before executing a command, but the `SAVE_VISUAL` flag would maintain the selection. Because the command handler was expecting Visual mode, the selection was never removed and the caret was in the wrong place. Removing all of this means the default behaviour works as expected.
- [x] Fix an issue with the `:normal` command and multi-character mappings. Because `:normal` was setting `mappingComplete` to `true`, it wasn't processing the last character correctly.
- [x] Rename the `SAVE_VISUAL` flag to `SAVE_SELECTION` to better indicate what's actually happening. IdeaVim always leaves Visual and returns to Normal before calling a command handler. The flag keeps the selection, not the Visual mode.

With regard to the `mappingComplete` parameter, this has been removed from `KeyConsumer`, but not from `KeyHandler.handleKey`. This function is used by external plugins, so will need to remain. It hasn't been marked as deprecated because there isn't an alternative yet. That will be taken care of in a forthcoming PR.

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/group/ChangeGroup.kt
+++ b/src/main/java/com/maddyhome/idea/vim/group/ChangeGroup.kt
@@ -7,7 +7,7 @@
  */
 package com.maddyhome.idea.vim.group
 
-import com.intellij.codeInsight.actions.AsyncActionExecutionService.Companion.getInstance
+import com.intellij.codeInsight.actions.AsyncActionExecutionService
 import com.intellij.openapi.Disposable
 import com.intellij.openapi.actionSystem.IdeActions
 import com.intellij.openapi.application.ApplicationManager
@@ -15,6 +15,7 @@ import com.intellij.openapi.command.CommandProcessor
 import com.intellij.openapi.command.UndoConfirmationPolicy
 import com.intellij.openapi.diagnostic.logger
 import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.editor.actionSystem.TypedActionHandler
 import com.intellij.openapi.editor.actions.EnterAction
 import com.intellij.openapi.editor.event.EditorMouseEvent
 import com.intellij.openapi.editor.event.EditorMouseListener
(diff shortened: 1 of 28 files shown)


AlexPl292:
Sweet, thank you!

citizenmatt:
Should have added some YouTrack issues, specifically [VIM-3489](https://youtrack.jetbrains.com/issue/VIM-3489/Keys-Are-Not-Executed-After-Timeout)

PR #1166, opened 2025-04-22
Remove obsolete internal options

citizenmatt:
Removes the `'commandormotionannotation'` and `'vimscriptfunctionannotation'` internal options that were used as feature flags, but there's no longer a switch on behaviour.


AlexPl292:
Sweet, thank you.

PR #1167, opened 2025-04-23
Always print float numbers using period as a decimal separator, VIM-3894

cypok:
[VIM-3894](https://youtrack.jetbrains.com/issue/VIM-3894/Unit-tests-fail-on-host-with-German-locale)


AlexPl292:
Thank you!

PR #1168, opened 2025-04-23
Replacement string fixes

cypok:
Fixes
* [VIM-3510](https://youtrack.jetbrains.com/issue/VIM-3510) IdeaVim fails switching case within backreferences,
* [VIM-3895](https://youtrack.jetbrains.com/issue/VIM-3895) Missing backreference should be substituted by empty string.


AlexPl292:
Thank you for the fixes!

PR #1173, opened 2025-05-05
Stop switching fragment editors to Insert mode

citizenmatt:
Fixes [VIM-1217](https://youtrack.jetbrains.com/issue/VIM-1217/Enable-for-Code-Fragments-Webstorm). This ticket was to enable IdeaVim in injected code fragments. This already worked, but the fragment editor started in Insert mode because IdeaVim would see it as an editor that was not backed by a real file. With this change, IdeaVim will check if the fragment editor's virtual file is an injected code fragment file, backed with a real file.


AlexPl292:
Oh, sweet, thank you!

PR #1182, opened 2025-05-08
VIM-3798 Replace with register empty text objects

vumi19:
Link to the YouTrack issue: [VIM-3798](https://youtrack.jetbrains.com/issue/VIM-3798)

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/extension/replacewithregister/ReplaceWithRegister.kt
+++ b/src/main/java/com/maddyhome/idea/vim/extension/replacewithregister/ReplaceWithRegister.kt
@@ -15,6 +15,7 @@ import com.maddyhome.idea.vim.api.ExecutionContext
 import com.maddyhome.idea.vim.api.ImmutableVimCaret
 import com.maddyhome.idea.vim.api.VimEditor
 import com.maddyhome.idea.vim.api.getLineEndOffset
+import com.maddyhome.idea.vim.api.getText
 import com.maddyhome.idea.vim.api.globalOptions
 import com.maddyhome.idea.vim.api.injector
 import com.maddyhome.idea.vim.command.MappingMode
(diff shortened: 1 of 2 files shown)


src/main/java/com/maddyhome/idea/vim/extension/replacewithregister/ReplaceWithRegister.kt
```diff
@@ -148,6 +149,22 @@ private fun doReplace(editor: Editor, context: DataContext, caret: ImmutableVimC
 
   var usedType = savedRegister.type
   var usedText = savedRegister.text
+
+  val selection = visualSelection.caretsAndSelections.values.first()
+  val startOffset: Int = selection.vimStart
+  val endOffset: Int = selection.vimEnd + 1
+
+  val isEmptySelection: Boolean = startOffset == endOffset
```
AlexPl292, line 157: Don’t you think this logic is a little strange?
Usually, the start offset is lower than the end offset. I'd expect that incrementing the end offset by one will make the difference between the start offset and the end offset even bigger, but here we expect them to become the same value.

src/main/java/com/maddyhome/idea/vim/extension/replacewithregister/ReplaceWithRegister.kt
```diff
@@ -148,6 +149,22 @@ private fun doReplace(editor: Editor, context: DataContext, caret: ImmutableVimC
 
   var usedType = savedRegister.type
   var usedText = savedRegister.text
+
+  val selection = visualSelection.caretsAndSelections.values.first()
```
AlexPl292, line 153: This logic is incorrect for the multicaret scenarios. This function is started for different carets, but we get the `first()` value for each call.

src/main/java/com/maddyhome/idea/vim/extension/replacewithregister/ReplaceWithRegister.kt
```diff
@@ -148,6 +149,22 @@ private fun doReplace(editor: Editor, context: DataContext, caret: ImmutableVimC
...
+
+  val isEmptySelection: Boolean = startOffset == endOffset
+
+  if (isEmptySelection) {
+    val editor: VimEditor = editor.vim
+
+    val beforeChar: String = editor.getText(startOffset - 1, startOffset)
+    val afterChar: String = editor.getText(endOffset, endOffset + 1)
+
+    usedText = beforeChar + usedText + afterChar
```
AlexPl292, line 165: Well, this might give you the correct execution results, but the approach is fundamentally incorrect. Instead of inserting the string between, say, brackets, you’re actually removing the brackets and inserting the text with additional wrapping.
While it may seem to work for now, it could lead to issues during future refactorings, since the operation being performed differs from what is expected.

AlexPl292 (changes requested):

AlexPl292:
I'm closing the PR as we discussed that it would be better to take a different approach.

PR #1183, opened 2025-05-12
Add support for non-control-character mapleader such as \<C-Space>

azjf:
Add support for non-control-character mapleader such as \<C-Space>


vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimStringParserBase.kt
```diff
@@ -416,18 +430,24 @@ abstract class VimStringParserBase : VimStringParser {
...
+                  }
                 }
+              } else {
+                keyCode = specialKey.keyCode
               }
+            }
+            if (keyCode != null) {
               result.append(keyCode.toChar())
             } else {
               result.append("<").append(specialKeyBuilder).append(">")
```
citizenmatt, line 453: I think it might be better to modify this line to return `\<${specialKeyBuilder}>`. If we're unable to convert the special key to an actual string value, then it would be better to let the `\<...>` syntax flow through to the rest of the editor.

If we don't do this, then IdeaVim will consider `let mapleader="<C-Space>"` to be the same as `let mapleader="\<C-Space>"` but only this last syntax is valid for Vim.

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimStringParserBase.kt
```diff
@@ -178,7 +182,12 @@ abstract class VimStringParserBase : VimStringParser {
   private fun getMapLeader(): List<KeyStroke> {
     val mapLeader: Any? = injector.variableService.getGlobalVariableValue("mapleader")
     return if (mapLeader is VimString) {
-      stringToKeys(mapLeader.value)
+      val v: String = mapLeader.value
+      if (v[0] == '<' && v[v.length - 1] == '>') {
```
citizenmatt, line 186: We should check for `\<...>`, but only if we let the escape char flow from the parser (see comment below)

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/helper/EngineStringHelper.kt
```diff
@@ -40,7 +40,13 @@ object EngineStringHelper {
...
       c = key.keyCode.toChar()
     } else if (c == KeyEvent.CHAR_UNDEFINED && key.modifiers and InputEvent.CTRL_DOWN_MASK != 0) {
-      c = (key.keyCode - 'A'.code + 1).toChar()
+      val code = key.keyCode
+      if (code >= 'A'.code && code <= '_'.code) {
+        // Ctrl-(A..Z [\]^_) are ASCII control characters
+        c = (code - 'A'.code + 1).toChar()
+      } else {
+        return "^" + code.toChar()
+      }
```
citizenmatt, line 49: I'm not sure about these changes. Are they necessary for this change? Are there any tests?

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimStringParserBase.kt
```diff
@@ -27,26 +27,30 @@ abstract class VimStringParserBase : VimStringParser {
...
       } else {
-        (key.keyCode - 'A'.code + 1).toChar()
+        "^" + key.keyCode.toChar()
       }
     } else if (key.keyChar == KeyEvent.CHAR_UNDEFINED && key.keyCode == KeyEvent.VK_ENTER) {
-      return '\u000D'
+      return "\u000D"
     }
-    return key.keyCode.toChar()
+    return key.keyCode.toChar().toString()
```
citizenmatt, line 53: I'm not sure about these changes. Are they necessary for this change? Are there any tests?

citizenmatt (changes requested):
Nice changes! I think we need to be careful that we don't support `let mapleader="<C-Space>"` (not valid in Vim). I can see this changing if/when we fix recording key strokes as part of macros.

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimStringParserBase.kt
azjf: Thank you, and I have fixed it now.

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimStringParserBase.kt
azjf: Same reason as the above one.

AlexPl292:
Hi, I apologize for the delay: I missed the update of the PR. Thank you for the changes; I've merged them!

PR #1188, opened 2025-05-15
Add support for `:=` to print line number

citizenmatt:
Supports `:{range}= [flags]` to output the line number for the given range. If there is no range, the default is `$`, which prints the last line number. If there's a range such as `3,6`, the last line is printed.

The `[flags]` can be any value in `l#p`. In Vim, this prints out the line number and the contents of the line over two separate lines. `#` shows the line with a line number prefix, `l` shows the content like `:list` with non-printable characters made visible and `p` outputs like `:print` (so just the text, no line number prefix). The flags can be combined, although if `l` is specified, it wins and the non-printable characters are displayed.

IdeaVim only has the status bar to show output, so can only show one line of output. If a flag is specied, the output is different to Vim - the single line output includes the line number and the text (very similar to the last line of output in Vim with `=#`).

Fixes [VIM-3921](https://youtrack.jetbrains.com/issue/VIM-3921)

### DIFF
--- a/src/test/java/org/jetbrains/plugins/ideavim/ex/implementation/commands/PrintLineNumberTest.kt
+++ b/src/test/java/org/jetbrains/plugins/ideavim/ex/implementation/commands/PrintLineNumberTest.kt
@@ -1,0 +1,106 @@
+/*
+ * Copyright 2003-2025 The IdeaVim authors
+ *
+ * Use of this source code is governed by an MIT-style
+ * license that can be found in the LICENSE.txt file or at
+ * https://opensource.org/licenses/MIT.
+ */
+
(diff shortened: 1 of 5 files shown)


danielzgtg:
"Status text" is disabled by [default](https://www.jetbrains.com/help/idea/guided-tour-around-the-user-interface.html#status-bar). Can you make this feature work without it?

danielzgtg:
> <C-G>

Wow, thx, that's faster than `:f`. That's also faster than `:=` and includes the line number so I'll be using it for my original use case. Still, this `:=` support is nice-to-have for completeness.

> the same thing
> at some point [...] we can update them

I agree. Consistency.

AlexPl292:
Great, thank you!

PR #1196, opened 2025-06-01
Fix VIM-3819 Arrow keys in macros are broken

Malandril:
This PR fixes https://youtrack.jetbrains.com/issue/VIM-3819/Arrow-keys-in-macros-are-broken that was caused when the `Register` that was refactored to be immutable in f1f86b5c .

It makes the register based on `KeyStroke` instead of the copied text.


AlexPl292:
Great, this also fixes [VIM-3902](https://youtrack.jetbrains.com/issue/VIM-3902) Alt key (and other) combinations not recorded in macros
Thank you for the change!

PR #1222, opened 2025-06-27
Feat(VIM-2143) expand environment variables in source command

zuberol:
…cription

### DIFF
--- a/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/model/commands/SourceCommand.kt
+++ b/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/model/commands/SourceCommand.kt
@@ -35,7 +35,7 @@ data class SourceCommand(val range: Range, val modifier: CommandModifier, val ar
     context: ExecutionContext,
     operatorArguments: OperatorArguments,
   ): ExecutionResult {
-    val path = expandUser(argument.trim())
+    val path = argument.trim().vimExpanded()
     val file = File(path)
     injector.vimscriptExecutor.executeFile(
       file,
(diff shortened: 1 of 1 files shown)


vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/model/commands/SourceCommand.kt
```diff
@@ -48,13 +48,23 @@ data class SourceCommand(val range: Range, val modifier: CommandModifier, val ar
...
 
-  private fun expandUser(path: String): String {
-    if (path.startsWith("~")) {
+  private fun String.vimExpanded(): String {
+    var expanded = this
+    if (startsWith("~")) {
       val home = System.getProperty("user.home")
       if (home != null) {
-        return home + path.substring(1)
+        expanded = home + substring(1)
```
AlexPl292, line 56: Hi, TBH I find a floating `substrint(1)` extremely confusing. I didn't even realize what the receiver of this function.
Please write an explicit receiver like `this.substring`.

AlexPl292: Same for `startsWith("~")`

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/model/commands/SourceCommand.kt
```diff
@@ -48,13 +48,23 @@ data class SourceCommand(val range: Range, val modifier: CommandModifier, val ar
...
       }
     }
-    return path
+
+    val envRe = Regex("\\$[A-Za-z0-9_]+")
```
AlexPl292, line 60: I feel like this approach skips some cases. For example, the env variable can have brackets `${PATH}`, or the dollar can be escaped like `\$PATH`.
On windows, the variables use a different syntax `%PATH%`.
Also, I'm not sure if the env variable can start with the number.

This expansion seems to be a non-trivial task. I'd try to search some existing ways in java/kotlin to expand it. Or, maybe there is an existing utility in the IJ platform (you can ask about it in `#ij-dev` slack channel).

If nothing will work, let's try to write our own expander, but keeping in mind the cases I described. Probably, it would make sense to read the specification on the env variables (if there is any), or ask AI what cases there are.

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/model/commands/SourceCommand.kt
```diff
@@ -48,13 +48,23 @@ data class SourceCommand(val range: Range, val modifier: CommandModifier, val ar
...
+    val env = System.getenv()
```
AlexPl292, line 61: As I see from the help you referencing, this expansion may be also useful in other places. Can I ask you to move it to the injector, to one of the classes?
Also, since this functionality is not trivial, it'll require some tests.

AlexPl292 (changes requested):

AlexPl292:
Also, in the commit message please always add a reference to the commit you're working on.

AlexPl292:
Yay, copilot now supports Kotlin 🔥

AlexPl292:
Hi, I’m closing this PR due to inactivity. If you’d like to resume work on it, you’re welcome to reopen it anytime.

PR #1223, opened 2025-06-27
Feat(VIM-3791): support for "G" and "gg" motions inside the NERDtree

zuberol:


AlexPl292:
Looks good to me, except the commit message :)
Can you please mention the ticket number there?
Also, after fixing the issue, you can mark the ticket as "Ready To Release".
If the ticket is a bug, you can use the form `Fix(VIM-XXX):` and the ticket will be switched to "Ready To Release" automatically.

wxh06:
I think these operations could be done by IntelliJ action `Tree-selectFirst` and `Tree-selectLast`.
I'll include this in my next PR.

PR #1224, opened 2025-07-01
Fix(VIM-2142): support `gU` and `gu` in VISUAL mode

wxh06:
Note that the behavior of `u` and `gu` (also `U` and `gU`) in visual mode is completely identical.


AlexPl292:
Looks good to me; I'm ready to merge it.

However, what makes you think that `gu` in visual mode is the same as `u`? Did you find any documentation on that?
I didn't see any documentation on `v_gu`, but it works in Vim. I feel like this is an undocumented feature.
Since we don't have a clear specification for this command, we have two options: investigate Vim's source code or assume how it works. Investigating the source code is an overkill for this case, so let's assume it works the same as `u`. However, this knowledge should be documented somewhere — either in a commit message or (better) directly in the code.

So, please add a comment on the commands saying that `gu` and `gU` are undocumented features and that we assume they work the same as `u` and `U`.

IdeaVim is a long-term project, so we need to consider "future us" and add explanations in any unclear areas. This way, after many years, we won't have to guess what happened or why these decisions were made.

AlexPl292:
BWT, ignore the comments from the copilot, they're useless. But I'm happy that it supports Kotlin finally.

wxh06:
That's fair, I'm going to add comments to document the behavior in source code.

BTW I've taken a look at their implementation in VSCodeVim/Vim#3603 - VSCodeVim also assumes that they are equivalent in visual mode.

PR #1226, opened 2025-07-03
Fix(VIM-2413): correct the range of line-wise case change commands

wxh06:
We observed that the start of the range is the leftmost non-whitespace character OR the current position, whichever is closer to the left.

### DIFF
--- a/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimChangeGroupBase.kt
+++ b/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimChangeGroupBase.kt
@@ -1857,16 +1857,11 @@ abstract class VimChangeGroupBase : VimChangeGroup {
     caret: VimCaret,
     context: ExecutionContext?,
     type: VimChangeGroup.ChangeCaseType,
-    argument: Argument,
+    argument: Argument.Motion,
     operatorArguments: OperatorArguments,
   ): Boolean {
-    if (argument !is Argument.Motion) {
-      throw RuntimeException("changeCaseMotion requires a Motion argument, but got $argument")
-    }
-
     var range = injector.motion.getMotionRange(
-      editor, caret, context!!, argument,
-      operatorArguments
+      editor, caret, context!!, argument, operatorArguments
     )
     if (range == null) return false
 
--- a/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimChangeGroupBase.kt
+++ b/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimChangeGroupBase.kt
@@ -1860,11 +1860,24 @@ abstract class VimChangeGroupBase : VimChangeGroup {
     argument: Argument,
     operatorArguments: OperatorArguments,
   ): Boolean {
-    val range = injector.motion.getMotionRange(
+    if (argument !is Argument.Motion) {
+      throw RuntimeException("changeCaseMotion requires a Motion argument, but got $argument")
+    }
+
+    var range = injector.motion.getMotionRange(
       editor, caret, context!!, argument,
       operatorArguments
     )
-    return range != null && changeCaseRange(editor, caret, range, type)
+    if (range == null) return false
+
+    // If the motion is linewise, we need to adjust range.startOffset to match the observed Vim behavior
+    if (argument.isLinewiseMotion()) {
+      val pos = editor.offsetToBufferPosition(range.startOffset)
+      // The leftmost non-whitespace character OR the current caret position, whichever is closer to the left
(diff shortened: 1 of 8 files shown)


vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimChangeGroupBase.kt
```diff
@@ -1860,11 +1860,20 @@ abstract class VimChangeGroupBase : VimChangeGroup {
...
-    val range = injector.motion.getMotionRange(
+    var range = injector.motion.getMotionRange(
       editor, caret, context!!, argument,
       operatorArguments
     )
-    return range != null && changeCaseRange(editor, caret, range, type)
+    if (range == null) return false
+
+    // If the motion is linewise, we need to adjust range.startOffset to match the observed Vim behavior
+    if ((argument as Argument.Motion).isLinewiseMotion()) {
```
wxh06: Since an `RuntimeException` will be thrown in `getMotionRange` if `argument !is Argument.Motion`. I think we can safely cast `argument` to `Argument.Motion` here.

AlexPl292: This is true, but it's not safe in the long-term. This may happen that at some moment `getMotionRange` will be able to process non-Motion arguments and `changeCaseMotion` will accept a non-Motion argument. So, I wouldn't rely on some internal details of some functions.

Instead, I'd ask the question: if the `argument` is not a Motion, is it a deal breaker? If yes, throw an exception, if not, use safe cast.

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimMotionGroupBase.kt
```diff
@@ -396,13 +396,11 @@ abstract class VimMotionGroupBase : VimMotionGroup {
...
     if (argument.isLinewiseMotion()) {
-      if (caret.getBufferPosition().line != editor.lineCount() - 1) {
-        start = editor.getLineStartForOffset(start)
-        end = min((editor.getLineEndForOffset(end) + 1).toLong(), editor.fileSize()).toInt()
-      } else {
-        start = editor.getLineStartForOffset(start)
-        end = editor.getLineEndForOffset(end)
-      }
+      start = editor.getLineStartForOffset(start)
+      end = if (caret.getBufferPosition().line != editor.lineCount() - 1)
```
AlexPl292: Never use `if` without parentheses if it's a multi-line `if`. It's just a matter of time before you'll miss it, add one more line, and you'll waste a day debugging.

AlexPl292 (changes requested):
Thank you for separating the refactoring and business logic changes into two separate commits — that’s very helpful.

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimChangeGroupBase.kt
wxh06: I think `changeCaseMotion` should not accept argument other than `Motion` (also due to its name). But we cannot change the function signature here since it is overriding the interface method, so we have no choice but to ensure the type at runtime.

In addition, I'd like to ask whether `logger.error` is needed when throwing a `RuntimeException`. Or the exception will be automatically logged instead?
Further more, do we have to throw the exception when a non-`Motion` argument is passed? The `changeMotion` method just simply return `false`.
https://github.com/JetBrains/ideavim/blob/2377408028673edfcbf4f6ba8001b778f5264f98/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimChangeGroupBase.kt#L1173

AlexPl292: > I think changeCaseMotion should not accept argument other than Motion (also due to its name). But we cannot change the function signature here since it is overriding the interface method, so we have no choice but to ensure the type at runtime.

We can change `changeCaseMotion` in both implementation and interface. However, yes, we still have to make a cast because `execute` function accepts the `Argument`. However, we may use "fail fast" approach and do the check/cast right in the `execute` method.

> In addition, I'd like to ask whether logger.error is needed when throwing a RuntimeException. Or the exception will be automatically logged instead?
Further more, do we have to throw the exception when a non-Motion argument is passed? The changeMotion method just simply return false.

Both thrown exception and LOG.error do log the exception. However, there is an important difference: LOG.error does not interrupt the code execution. So, if you want to log the exception and continue the code execution, use the log. If you want this code to interrupt because it won't make sense to continue, use the throwing.
In this case, I'd actually prefer a safer approach, like in `changeMotion`: checking the type of the argument and LOG.error if it's not the motion. The reason for that is there might be some post-action processing and state machine update. If we'll just throw an exception, this logic may break.

PR #1227, opened 2025-07-03
Fix(VIM-2375): do not save file with `ZQ`

wxh06:
ZQ is defined to `Quit without checking for changes (same as ":q!").`

### DIFF
--- a/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/action/file/FileCloseAction.kt
+++ b/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/action/file/FileCloseAction.kt
@@ -16,9 +16,9 @@ import com.maddyhome.idea.vim.command.Command
 import com.maddyhome.idea.vim.command.OperatorArguments
 import com.maddyhome.idea.vim.handler.VimActionHandler
 
-@CommandOrMotion(keys = ["ZZ"], modes = [Mode.NORMAL])
-class FileSaveCloseAction : VimActionHandler.SingleExecution() {
-  override val type: Command.Type = Command.Type.OTHER_WRITABLE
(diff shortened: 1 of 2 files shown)


vim-engine/src/main/kotlin/com/maddyhome/idea/vim/action/file/FileCloseAction.kt
```diff
@@ -0,0 +1,32 @@
...
+import com.maddyhome.idea.vim.api.ExecutionContext
+import com.maddyhome.idea.vim.api.VimEditor
+import com.maddyhome.idea.vim.api.injector
+import com.maddyhome.idea.vim.command.Command
+import com.maddyhome.idea.vim.command.OperatorArguments
+import com.maddyhome.idea.vim.handler.VimActionHandler
+
+@CommandOrMotion(keys = ["ZQ"], modes = [Mode.NORMAL])
+class FileCloseAction : VimActionHandler.SingleExecution() {
+  override val type: Command.Type = Command.Type.OTHER_READONLY
```
AlexPl292: Those types are a bit messy and I'd like to get rid of them. However, let's try to keep it clean anyway.
I'd suggest using `OTHER_SELF_SYNCHRONIZED` here are we're not intended to run write or read command.

wxh06: Is `OTHER_SELF_SYNCHRONIZED` applicable to `FileSaveCloseAction`? Then VIM-3266 can also be solved.

AlexPl292: I think yes

PR #1228, opened 2025-07-04
Fix(VIM-3044): correct the behavior of `I` in (linewise) Visual mode

wxh06:
The bahavior seems undocumented. Based on my observation in Vim:

- For visual selections spanning multiple lines, keep caret position if it's on the first line.
- Otherwise move the caret to the start of the first selected line.

### DIFF
--- a/src/test/java/org/jetbrains/plugins/ideavim/action/change/insert/VisualInsertActionTest.kt
+++ b/src/test/java/org/jetbrains/plugins/ideavim/action/change/insert/VisualInsertActionTest.kt
@@ -17,7 +17,7 @@ import org.jetbrains.plugins.ideavim.TestWithoutNeovim
 import org.jetbrains.plugins.ideavim.VimTestCase
 import org.junit.jupiter.api.Test
 
-class VisualBlockInsertActionTest : VimTestCase() {
+class VisualInsertActionTest : VimTestCase() {
   // VIM-1379 |CTRL-V| |j| |v_b_I|
   @TestWithoutNeovim(SkipNeovimReason.VISUAL_BLOCK_MODE)
   @Test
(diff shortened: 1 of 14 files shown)


vim-engine/src/main/kotlin/com/maddyhome/idea/vim/action/change/insert/VisualInsertAction.kt
```diff
@@ -38,7 +46,17 @@ class VisualBlockInsertAction : VisualOperatorActionHandler.SingleExecution() {
     return if (vimSelection.type == SelectionType.BLOCK_WISE) {
       injector.changeGroup.initBlockInsert(editor, context, vimSelection.toVimTextRange(false), false)
     } else {
-      injector.changeGroup.insertBeforeFirstNonBlank(editor, context)
+      // For visual selections spanning multiple lines, keep caret position if it's on the first line
+      // Otherwise move the caret to the start of the first selected line
+      for ((caret, selection) in caretsAndSelections) {
+        val range = selection.toVimTextRange()
+        val posStart = editor.offsetToBufferPosition(range.startOffset)
+        val nextLineStart = editor.getLineStartOffset(editor.normalizeLine(posStart.line + 1))
```
AlexPl292, line 54: I'd leave a small node why "+1" is needed.

wxh06:
> This update will defenetlly require tests

Yeah this is exactly why it is opened as a draft PR.
I'm currently working on testing.

src/test/java/org/jetbrains/plugins/ideavim/action/change/insert/VisualInsertActionTest.kt
```diff
@@ -101,28 +101,76 @@ class VisualBlockInsertActionTest : VimTestCase() {
...
+        hard by the torrent of a mountain pass.
+    """.trimIndent()
+    doTest(listOf("vjIHello<esc>"), before, after)
+    doTest(listOf("VjIHello<esc>"), before, after)
+  }
+
+  @Test
+  fun `test insert in non-block visual spanning multiple lines up`() {
+    val before = """
+      .
```
AlexPl292: What are these dots needed for?

wxh06: To retain the 2-space indentation on the following lines. I wonder if there could be a better way?

AlexPl292: Please, use 'trimMargin` for it then.
https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/trim-margin.html

PR #1229, opened 2025-07-04
Fix(VIM-3266): Set `FileSaveCloseAction` to `OTHER_SELF_SYNCHRONIZED`

wxh06:
As discussed in <https://github.com/JetBrains/ideavim/pull/1227#discussion_r2182963443>.


AlexPl292:
When making such a change, you should either write tests or, at the very least, manually verify it. Have you tried doing any of this? If so, can you share the results?

wxh06:
I tried pressing `ZZ` and the it did close the read-only file normally without showing "Clear Read-Only Status" popup.
However, I have no ways to verify that the file is actually saved, since IntelliJ will save the file at the time when making the file read-only (even with autosave disabled) and no further modification is allowed.

PR #1230, opened 2025-07-08
Refactor command line to support control characters

citizenmatt:
This PR makes a number of refactorings to the command line infrastructure. It simplifies keystroke handling, removes a lot of unused or unnecessary Swing shortcut and action handling, removes deprecated code usage and several other refactorings. 

It also adds support for inserting and rendering control characters in command lines, with e.g. `<C-V><Esc>` or `<C-V>x80` ([VIM-239](https://youtrack.jetbrains.com/issue/VIM-239), [VIM-3423](https://youtrack.jetbrains.com/issue/VIM-3423), [VIM-3907](https://youtrack.jetbrains.com/issue/VIM-3907)). The control characters are rendered in Vim style (e.g. `^[` or `<80>`) and formatted in a different colour. The prompt character while inserting a literal, digraph or register also has custom formatted rendering.

Screenshot 2025-07-08 at 3 53 28 pm

## Full details:

- [x] Remove non-Vim key bindings `<S-Ins>` and `<A-V>` from ex command line. These are undocumented and non-standard, and can easily be added with `cmap` if needed.
- [x] Remove obsolete `ExShortcutKeyAction`. The shortcuts are already registered with `VimShortcutKeyAction`, which uses `KeyHandler` to process keys, including mapping.
- [x] Remove unnecessary Swing action registration. All actions are handled by `KeyHandler`, and nothing was actually registered here.
- [x] Simplify and document Swing keymap registration. All shortcuts are handled by the IdeaVim pipeline, not Swing. However, we want to remove existing key mappings so that existing Swing actions don't get invoked accidentally. This is subtle and confusing, so document what's going on.
- [x] Remove unused mouse listener.
- [x] Remove unused ex text field instance without shortcuts. This was required for modal input, but that's already supported using the main text field instance and a modal input key consumer in `KeyHandler`.
- [x] Simplify keystroke handling in `ExTextField`, removing `DefaultExKeyHandler`. The keystroke is now passed to the `KeyHandler` directly, possibly mapping the keystroke, and passing it back to the Swing text field via `ExEntryPanel.handleKey`. 
    The previous dataflow would pass the keystroke to the super class, where it would be converted to a Swing action, handled by the `DefaultExKeyHandler` catch-all action handler, converted back to a keystroke (an imprecise/lossy conversion with control characters) and then passed to `KeyHandler`. The key handler can map it, and then pass the keystroke to `ExEntryPanel.handleKey`, where a flag is set and it's passed to the text field's super class again, converted to an action, handled by `DefaultExKeyHandler` again, which checks the flag and finally passed to the super class to add as character 😰 
- [x] Replace deprecated command line history code, update to use proper API. This fixes warnings in `ExTextField`, but revealed bugs in the `:history` command output.
- [x] Update `:history` command output, adding tests.
- [x] Refactor `ExTextField` to move as much state as possible to `ExEntryPanel`, encapsulating `ExTextField` better.
- [x] Fix scrolling issues when editing a long text that requires scrolling. Vim actions would get the current text from the field, modify it and replace it all. This resets scroll location because the current text is deleted before being set. Added new methods for replacing or deleting a range of text in the command line, to maintain scroll location.
- [x] Insert control characters into the command line with e.g. `<C-V>x80` or `<C-V><Esc>`. Note that while control characters are accepted, they might not work as expected because the main IntelliJ editor doesn't like some characters.
- [x] Render control characters in Vim style, e.g. `<80>` or `^L`, highlighted as "SpecialKey". This is handled with a custom Swing `View` for the text field.
- [x] Render the prompt character (`^` for inserting literals, `?` for digraphs`, `"` for registers) with a custom `View` instead of manually managing `actualText`, `visibleText` and offsets. Fixes an exception due to incorrect offset.
- [x] Various other smaller refactorings, e.g. split `VimInterceptor` to fix warnings, use enums instead of constancts for digraph states (better for debugging), etc.


citizenmatt:
> ### Bug: History Command Range Calculation Bug
> The `:history` command's behavior has changed: when `myLast` is 0, it now shows only the `myFirst` entry instead of all entries from `myFirst` to the end (e.g., `:history 5` now shows only entry 5). Additionally, the calculation for negative `myLast` incorrectly references an undefined `entry` variable; it should likely be `entries.size + myLast + 1` similar to the `myFirst` calculation.
> 
> `vim-engine/src/main/kotlin/com/maddyhome/idea/vim/history/VimHistoryBase.kt#L45-L50`

This is actually correct Vim behaviour. `:history 5` displays entry 5 only. The second number is required to display a range, e.g. `:history 5,10`

citizenmatt:
> ### Bug: Abbreviation Matching Flaw in Prefix Parsing
> Incorrect abbreviation matching logic in parseTypePrefix method. The code uses `arg.commonPrefixWith(type)` which will incorrectly match invalid abbreviations. For example, "ca" would match "cmd" (returning the common prefix "c"), when it should only match if the entire arg is a prefix of type. This should be `type.startsWith(arg)` instead to properly validate abbreviations.
> 
> `vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/model/commands/HistoryCommand.kt#L114-L121`

This is also correct Vim behaviour. Typing `:history ca` will report an error with `a` as trailing characres `E488: Trailing characters: a`

src/main/java/com/maddyhome/idea/vim/ui/ex/ExDocument.java
```diff
@@ -91,14 +75,155 @@ private void addInputMethodAttributes(AttributeSet attributeSet) {
...
+        }
+        offset += s.length();
+
+        start = pos;
+      }
+    }
+
+    private String filterNewLines(Document document, String string) {
+      // A text field shouldn't have new lines
+      Object filterNewlines = document.getProperty("filterNewlines");
```
AlexPl292, line 227: This requires a note that this property is embedded and how does this property work.

src/main/java/com/maddyhome/idea/vim/ui/ex/ExDocument.java
```diff
@@ -91,14 +75,155 @@ private void addInputMethodAttributes(AttributeSet attributeSet) {
...
+        final AttributeSet a = isPrintable ? attr : getNonPrintableAttributes(attr);
+        final String s = string.substring(start, pos);
+        fb.insertString(offset, s, a);
+        offset += s.length();
+
+        start = pos;
+      }
+
+      if (fb.getDocument() instanceof ExDocument exDocument && exDocument.isOverwrite()) {
+        int rest = originalOffset + string.length();
```
AlexPl292, line 161: I see the super method of `insertString` and there is just a call for `fb.insertString`. Why do we have to process the overwrite mode here explicitly?

AlexPl292: Is it because we have a custom "replace" mode?

AlexPl292 (approved):
Thank you for the changes and the detailed description.

I feel like `ExDocumentFilter` is quite complicated because we do a lot of work manually. However, it seems like we do that because we have a custom replace mode, isn't it? So, we can't just use a call to `super.insertString` and then use `setCharacterAttributes` to set the needed attributes to the non-printable chars, right?

If this is the point, I'm ready to merge the changes.

src/main/java/com/maddyhome/idea/vim/ui/ex/ExDocument.java
citizenmatt: I'd have to dive back in to be 100% certain, but basically yes, it's all about the overwrite mode. IIRC, the default text field and document don't have any support for overwrite/replace (what Vim calls "overstrike") so we have to implement it ourselves. 

Most of the time, e.g. while typing, the filter's `replaceString` method is called. But `insertString` is called during other operations such as paste (IntelliJ's paste, which we can't easily opt out of).

src/main/java/com/maddyhome/idea/vim/ui/ex/ExDocument.java
citizenmatt: Embedded meaning it's a Swing property? I'm working on some further refactorings around command line. I can add the comments later.

citizenmatt:
Yes, we have to implement Vim's "overstrike" mode manually, so we can't just call super. Also, we split the text based on attributes while inserting. Calling super and then setting char attributes would insert the document, update views and then split the document and update views again. It's better to split once with the right attributes.

AlexPl292:
Okay, I got it. Thank you!

PR #1243, opened 2025-07-17
[VIM-3947] remove AWT KeyStrokes

zuberol:
This commit replaces AWT classes with custom ones. The API is basically the copy of the original one. I agree that it can be improved, but the current version unblocks development in Fleet, which was the main goal.


src/main/java/com/maddyhome/idea/vim/group/RegisterGroup.java
```diff
@@ -70,8 +72,8 @@ public void saveData(final @NotNull Element element) {
       else {
         logger.trace("Save register as 'keys'");
         final Element keys = new Element("keys");
-        final List<KeyStroke> list = register.getKeys();
-        for (KeyStroke stroke : list) {
+        var list = register.getKeys();
```
AlexPl292, line 75: Minor: there is no need to change `final List` to `var`. It's enough to change the type.

AlexPl292 (commented):
Hi, I've made a review of the changes, but I'd like to make one more a bit later (next week).
Here are the first notes:
- Do I get it right that changes are intended to be squashed before merging? If not, the commits require a hard refactoring to get rid of wip commits, make proper commit messages, get rid of changes that are undone in following changes. If we'll squash them, it's better to do by you to provide a proper commit message.
- The file `.name` should not be included in these changes
- I see a huge number of unused variables in VimKeyStroke. Is it intended?
- These changes do require a testing. This would allow to catch small issues, like the one with incorrect modifiers parsing.
- It was mentioned by Matt that we should also take a look at the [caching](https://youtrack.jetbrains.com/issue/VIM-3947) opportunities. Actually, when I checked the KeyStroke implementation, it turned out that there is caching inside. Can you please take a look on this?

src/main/java/com/maddyhome/idea/vim/extension/matchit/Matchit.kt
```diff
@@ -41,6 +41,7 @@ import com.maddyhome.idea.vim.handler.toMotionOrError
 import com.maddyhome.idea.vim.helper.EditorHelper
 import com.maddyhome.idea.vim.helper.PsiHelper
 import com.maddyhome.idea.vim.helper.enumSetOf
+import com.maddyhome.idea.vim.helper.keyStroke
```
AlexPl292, line 44: Unused import

src/main/java/com/maddyhome/idea/vim/extension/textobjindent/VimIndentObject.java
```diff
@@ -32,6 +34,9 @@
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import javax.swing.*;
```
AlexPl292, line 37: In this file all the changes are not related to the PR

src/test/java/org/jetbrains/plugins/ideavim/ex/parser/expressions/DoubleQuotedStringTest.kt
```diff
@@ -167,36 +169,36 @@ class DoubleQuotedStringTest {
...
-    assertEquals(listOf(getKeyStroke(87, CTRL_DOWN_MASK)), injector.parser.stringToKeys(23.toChar().toString()))
-    assertEquals(listOf(getKeyStroke(88, CTRL_DOWN_MASK)), injector.parser.stringToKeys(24.toChar().toString()))
-    assertEquals(listOf(getKeyStroke(89, CTRL_DOWN_MASK)), injector.parser.stringToKeys(25.toChar().toString()))
-    assertEquals(listOf(getKeyStroke(90, CTRL_DOWN_MASK)), injector.parser.stringToKeys(26.toChar().toString()))
-    assertEquals(listOf(getKeyStroke(91, CTRL_DOWN_MASK)), injector.parser.stringToKeys(27.toChar().toString()))
-    assertEquals(listOf(getKeyStroke(92, CTRL_DOWN_MASK)), injector.parser.stringToKeys(28.toChar().toString()))
-    assertEquals(listOf(getKeyStroke(93, CTRL_DOWN_MASK)), injector.parser.stringToKeys(29.toChar().toString()))
-    assertEquals(listOf(getKeyStroke(94, CTRL_DOWN_MASK)), injector.parser.stringToKeys(30.toChar().toString()))
-    assertEquals(listOf(getKeyStroke(95, CTRL_DOWN_MASK)), injector.parser.stringToKeys(31.toChar().toString()))
+    assertEquals(listOf(getKeyStroke(65, CTRL_DOWN_MASK).vimKeyStroke), injector.parser.stringToKeys(1.toChar().toString()))
```
AlexPl292, line 172: Can we use `VimKeyStroke` directly here?

src/test/java/org/jetbrains/plugins/ideavim/helper/StringHelperTest.kt
```diff
@@ -117,6 +118,6 @@ class StringHelperTest : VimTestCase() {
   private fun parseKeyStroke(s: String): KeyStroke {
     val actualStrokes = injector.parser.parseKeys(s)
     assertEquals<Any>(1, actualStrokes.size, injector.parser.toKeyNotation(actualStrokes))
-    return actualStrokes[0]
+    return actualStrokes[0].keyStroke
```
AlexPl292, line 121: Here we can return `VimKeyStroke` and use it further in test

AlexPl292 (changes requested):
Okay, I’ve finished the review.

I’ve also left a few minor comments.
Since there’s no urgency for Fleet on the web, I’d prefer to merge the task once it’s fully complete. This is a solid starting point, so we can continue with the remaining parts of the ticket.

Let’s update the ticket: cross out the completed items and proceed with what’s mentioned in the comments, as well as any other remaining requirements.

PR #1244, opened 2025-07-18
Fix issues with new islands theme

citizenmatt:
The [recent UI refresh](https://blog.jetbrains.com/platform/2025/06/testing-a-fresh-look-for-jetbrains-ides/) draws the IDE frame with a nice gradient fill. Any buttons drawn on the frame, like the main toolbar, or the tool window "strips" are drawn as though their background is transparent, while the opaque tool windows and editor look like "islands" on the frame.

This breaks the colour highlighting of the mode status bar widget, so it no longer shows a different colour per mode, only the new gradient fill. The implementation also affects the output panel (because it's drawn with `IdeRoot` too), so the output of e.g. `:map` is also drawn with a gradient background.

The implementation of islands is done by wrapping the `Graphics` instance used to draw the `IdeRoot`, including all of its children. This wrapped `Graphics` object will draw the fancy background whenever it's asked invoked for a fill-like command.

There are several ways to avoid this. If there's a `JBPanel` instance in the parent hierarchy, the `IdeBackgroundUtil.NO_BACKGROUND` client property will opt out a component and its children. This is because `JBPanel.getComponentGraphics` is set up to either return a wrapped `Graphics` or check for the property and return the original instance. Alternatively, the `IdeBackgroundUtil.getOriginalGraphics` will unwrap the `Graphics` instance if it's wrapped. Both approaches are used here to make the output panel and status bar widget work.

This has been tested with islands enabled in 2025.2, and with the current 2025.1.

Interestingly, `ExEntryPanel` is implemented in essentially the same way as `ExOutputPanel`, so one would expect this to also be broken. However, it works just fine. I'm not entirely sure why. I _think_ it's because it sets the panel's parent to the editor component, and the editor has already opted out of the new frame background.

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/ui/ExOutputPanel.java
+++ b/src/main/java/com/maddyhome/idea/vim/ui/ExOutputPanel.java
@@ -13,6 +13,9 @@ import com.intellij.ide.ui.LafManagerListener;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.editor.Editor;
 import com.intellij.openapi.project.Project;
+import com.intellij.openapi.wm.impl.IdeBackgroundUtil;
+import com.intellij.ui.ClientProperty;
+import com.intellij.ui.components.JBPanel;
 import com.intellij.ui.components.JBScrollPane;
 import com.intellij.util.IJSwingUtilities;
 import com.maddyhome.idea.vim.KeyHandler;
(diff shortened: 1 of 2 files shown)


AlexPl292:
Thank you for the fix. One day, we need to refresh it. The widget now looks outdated: no curved edges, no alignment with other elements :( 
Even the colorless theme uses the wrong background color.

citizenmatt:
Hmm. Good point. I didn't think about checking the look-n-feel, just making it work again. Perhaps I'll ping one of the designers, see if they can suggest something for us.

PR #1245, opened 2025-07-21
Refactor NERDTree implementation

wxh06:
To extend NERDTree support to all `Tree` components, we need to extract common actions to reuse the code.

- [x] Extract navigation-related NERDTree actions to make them reuasble
- [x] Fix [VIM-3985](https://youtrack.jetbrains.com/issue/VIM-3985) NERDTree-x behavior
- [x] Fix [VIM-3437](https://youtrack.jetbrains.com/issue/VIM-3437) NERDTree-X behavior
- [x] Avoid `KeyStrokeTrie` getting stuck in an invalid state


src/main/java/com/maddyhome/idea/vim/extension/nerdtree/Mappings.kt
```diff
@@ -0,0 +1,165 @@
...
+    // TODO support going [count] lines upward/downward or to line [count]
+    register("k", Action.ij("Tree-selectPrevious"))
+    register("j", Action.ij("Tree-selectNext"))
+    register("G", Action.ij("Tree-selectLast"))
+    register("gg", Action.ij("Tree-selectFirst"))
+
+    // FIXME lazy loaded tree nodes are not expanded
+    register("NERDTreeMapOpenRecursively", "O", Action.ij("FullyExpandTreeNode"))
+    // This action respects `ide.tree.collapse.recursively`. We may have to ask the user to disable it
+    register("NERDTreeMapCloseDir", "x", Action.ij("Tree-selectParentNoCollapse", "CollapseTreeNode"))
```
wxh06: Do we need to **temporarily** disable `ide.tree.collapse.recursively` (by `AdvancedSettings.setBoolean`) to make the action behavior identical to the one in Vim NERDTree? Or prompt the user to do so manually?

AlexPl292: Definitely not asking the user to do it manually. Asking the user to do something is a bad pattern that should generally be avoided, as the user's goal is to finish their work instead of responding to IDE questions. Of course, sometimes we have to do this, but this particular question is too small and unimportant to bother the user.

Generally, I don't believe this is important for the user, and they won't even notice the difference. In such cases, use the following framework: If the change is complicated or produces a lot of code, it's safer to keep things as they are. If there is a Vim configuration option for this in the nerdtree plugin, implement it and default to either Vim or IJ state. If there is no option, and the change is minor and unimportant, take a moment to reconsider and just choose one approach.

In this particular example, I'd like to leave this decision to you. 
However, I personally suggest not making any changes and maintaining a native IJ experience. The reasons for this are:
- Overriding this behaviour will practically break this setting.
- It's unreliable to depend on a registry option. It may be renamed or removed in the future, resulting in obsolete or broken code for us.
Generally, we should write an automated test if we plan to change IJ behavior here. (See the UI tests on GitHub if you decide to go this route).

AlexPl292: Oh oh, after I answered to your comment, I've noticed this ticket
[VIM-3437](https://youtrack.jetbrains.com/issue/VIM-3437) NERDtree: "Recursively close all children of the current node" does not work

Well, in this case, since the NERDTree documentation explicitly states that it recursively closes the nodes, then we have to obey and also do it recursively.

wxh06: [VIM-3437](https://youtrack.jetbrains.com/issue/VIM-3437) is about capital `X`. I implemented this action manually instead of relying on IntelliJ's `ide.tree.collapse.recursively` due to reasons in the comment:

https://github.com/JetBrains/ideavim/blob/cc6f8573a3dc23f931e4273de7d5e25dd9204fce/src/main/java/com/maddyhome/idea/vim/extension/nerdtree/Mappings.kt#L98-L111

https://github.com/JetBrains/ideavim/blob/cc6f8573a3dc23f931e4273de7d5e25dd9204fce/src/main/java/com/maddyhome/idea/vim/extension/nerdtree/Mappings.kt#L143-L162

Regarding lower case `x`, I'd like to leave the decision (whether to close the dir recursively) to users.
My original idea was to use a tooltip (like the one in SpeedSearch or those due to shortcut conflict in IdeaVim) to remind users of the reason for the behavior when they trigger it for the first time, and to inform them how to modify the behavior to be consistent with Vim.

src/main/java/com/maddyhome/idea/vim/extension/nerdtree/Mappings.kt
```diff
@@ -0,0 +1,162 @@
...
+      },
+    )
+    register(
+      "NERDTreeMapJumpLastChild",
+      "J",
+      Action { _, tree ->
+        val selectedPath = tree.selectionPath ?: return@Action
+        val parent = selectedPath.parentPath
+        var row = tree.getRowForPath(selectedPath)
+        while (parent.isDescendant(tree.getPathForRow(row + 1))) row++
```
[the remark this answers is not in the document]
wxh06: `TreePath.isDescendant` does accept `null` as its argument and returns `false`, which is the expected value.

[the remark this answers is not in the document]
wxh06: `boolean isDescendant(TreePath aTreePath)` returns true if `aTreePath` is a descendant of this `TreePath`.

src/main/java/com/maddyhome/idea/vim/extension/nerdtree/Mappings.kt
```diff
@@ -0,0 +1,163 @@
...
```
wxh06: Oops, `tree.getPathForRow(row + 1)` is supposed to be the _child_ of `parent`

src/main/java/com/maddyhome/idea/vim/extension/nerdtree/Mappings.kt
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @constructor Creates an empty trie for key mappings
+ * @param name The name of the KeyStrokeTrie instance (for debug purposes)
+ */
+internal class Mappings(name: String) {
+  private val trie = KeyStrokeTrie<NerdAction>(name)
+  val getAction = trie::getData
+
+  private val _keyStrokes = mutableSetOf<KeyStroke>()
+  val keyStrokes: Set<KeyStroke> get() = _keyStrokes.toSet()
```
AlexPl292: Please, take a look this `toSet()` function once again:
- If the code accessing it is single-threaded or you don't expect the `_keyStrokes` to be modified while the `keyStrokes` is accessed, then just using `_keyStrokes` is enough.
- It's this code should be thread-safe, this approach is good.

src/main/java/com/maddyhome/idea/vim/extension/nerdtree/AbstractDispatcher.kt
```diff
@@ -0,0 +1,69 @@
...
+    val keyChar = keyStroke.keyChar
+    if (keyChar != KeyEvent.CHAR_UNDEFINED) {
+      keyStroke = KeyStroke.getKeyStroke(keyChar)
+    }
+
+    keys.add(keyStroke)
+    mappings.getAction(keys)?.let { action ->
+      val component = e.getData(PlatformDataKeys.CONTEXT_COMPONENT)
+      if (component is Tree) action.action(e, component)
+      else LOG.error("Component is not a tree: $component")
```
AlexPl292: Do never use `if` without brackets unless it's a super trivial one-lines. It's a matter of time until you'll add a new line after `if` and will wounder why it doesn't work. Also, it's easier to miss the `else` without the brackets.

src/main/java/com/maddyhome/idea/vim/extension/nerdtree/AbstractDispatcher.kt
```diff
@@ -0,0 +1,69 @@
...
+        return strokeCache.second
+      }
+      return KeyStroke.getKeyStrokeForEvent(inputEvent)
+    }
+    return null
+  }
+
+  companion object {
+    val LOG = logger<AbstractDispatcher>()
+  }
```
AlexPl292, line 77: Please use `vimLogger`

src/main/java/com/maddyhome/idea/vim/extension/nerdtree/Mappings.kt
```diff
@@ -0,0 +1,66 @@
...
+        NerdTree.Util.callAction(null, id, event.dataContext.vim)
+
+      /**
+       * Creates an Action that executes a series of IntelliJ actions identified by their IDs.
+       *
+       * @param ids A variable number of strings representing the IDs of the actions to execute.
+       * @return An Action that runs the specified actions when triggered.
+       */
+      fun ij(vararg ids: String): Action = Action { event, _ -> ids.forEach { id -> executeAction(event, id) } }
+    }
```
AlexPl292, line 68: Calling actions one by one is a complicated topic. The action initially called from the EDT can easily switch to the background thread, postpone execution and do almost anything. I see, there is a single use of multiple actions, is it possible to replace it with code alternative?

src/main/java/com/maddyhome/idea/vim/extension/nerdtree/Mappings.kt
```diff
@@ -0,0 +1,168 @@
...
+        TreeUtil.scrollToVisible(tree, path, false)
+      },
+    )
+    register("NERDTreeMapJumpNextSibling", "<C-J>", Action.ij("Tree-selectNextSibling"))
+    register("NERDTreeMapJumpPrevSibling", "<C-K>", Action.ij("Tree-selectPreviousSibling"))
+  }
+}
+
+private fun JTree.nonLeafDescendants(path: TreePath): List<TreePath> {
+  if (model.isLeaf(path.lastPathComponent)) return emptyList()
```
AlexPl292: Correct me if I'm wrong, but it looks like you're traversing the whole tree in order to close the leaf.

We should not do that, this will be a huge performance issue both mathematically and in reality.
The tree can be really huge, think about some `node_modules` folder where a half of the npm can be downloaded. Traversing the whole tree will be really slow.
Also, there is a thing called an async tree where the elements are not loaded unless they're needed. This is really useful in, for example, the project tree where you have to access the file system in order to get the tree information. Calling `getChild` will cause loading the whole tree.

AlexPl292 (changes requested):
Hi, please check my comments.
Also, please review your commit messages. The split in commits is great, but some commit messages can be improved. Please, refer to this doc: https://gist.github.com/finalfantasia/bd0070673ca27e5f7473

src/main/java/com/maddyhome/idea/vim/extension/nerdtree/Mappings.kt
wxh06: You're right. I didn't find a good way to actually close every subdirectory recursively.

IntelliJ does not expose an interface to close directories recursively. It is done internally in `Tree.collapsePath` if `ide.tree.collapse.recursively` is set.

<https://github.com/JetBrains/intellij-community/blob/801affaf9a1f0da5fc6390d1ccaa2b8cb11e612c/platform/platform-api/src/com/intellij/ui/treeStructure/Tree.java#L536-L552>

Additionally, IntelliJ's approach only closes visible rows and may leave deep descendants expanded, especially when users use both `X` and `x` with `ide.tree.collapse.recursively` unset.

But we may consider it an edge case and adopt IJ's implementation?
If we decide to adopt it, the question comes to: should we temporarily set `ide.tree.collapse.recursively` as a workaround or reimplement the logic in our codebase?

src/main/java/com/maddyhome/idea/vim/extension/nerdtree/AbstractDispatcher.kt
wxh06: I see. You've mentioned this before, but I thought this piece of code could be considered "one-line" at that time. Now it seems that only simple ones like `break` will do.

src/main/java/com/maddyhome/idea/vim/extension/nerdtree/AbstractDispatcher.kt
wxh06: This also needs to be changed.

https://github.com/JetBrains/ideavim/blob/4eadfc1fbaf47797899cb66402ca3168bdbc5dcb/src/main/java/com/maddyhome/idea/vim/extension/nerdtree/NerdTree.kt#L539

src/main/java/com/maddyhome/idea/vim/extension/nerdtree/Mappings.kt
AlexPl292: Let's adopt the IJ implementation. 
Temporarily enabling `ide.tree.collapse.recursively` sounds easier for me to use. However, try to write a test that will check that this option still exists in IJ.

src/main/java/com/maddyhome/idea/vim/extension/nerdtree/Mappings.kt
```diff
@@ -0,0 +1,159 @@
...
+        // Note that we cannot use `tree.collapsePaths` here since it does not respect `ide.tree.collapse.recursively`
+        TreeUtil.listChildren(path.lastPathComponent as TreeNode).filterNot(TreeNode::isLeaf)
+          .map(path::pathByAddingChild).forEach(tree::collapsePath)
+        AdvancedSettings.setBoolean("ide.tree.collapse.recursively", recursively)
+
+        tree.scrollPathToVisible(path)
+      },
+    )
+
+    register("NERDTreeMapJumpRoot", "P", Action.ij("Tree-selectFirst"))
```
wxh06, line 130: The first row is not the root of "External Libraries" in IntelliJ.
Is it supposed to jump to the root of project files instead of libraries?

src/test/java/org/jetbrains/plugins/ideavim/extension/nerdtree/NerdTreeTest.kt
```diff
@@ -0,0 +1,20 @@
...
+package org.jetbrains.plugins.ideavim.extension.nerdtree
+
+import com.intellij.openapi.options.advanced.AdvancedSettings
+import org.jetbrains.plugins.ideavim.VimTestCase
+import org.junit.jupiter.api.Test
+
+class NerdTreeTest : VimTestCase() {
+  @Test
+  fun `test collapse recursively advanced setting id`() {
+    AdvancedSettings.getBoolean("ide.tree.collapse.recursively") // will throw if the id is invalid
```
AlexPl292: Please, use `assertDoesNotThrow`

AlexPl292 (approved):
Please, fix the tiny comment and we're good to merge

PR #1262, opened 2025-07-31
Fix initialising non-default colorcolumn option

citizenmatt:
When initialising a new window, we copy options from the current window. Because there might not be an initial window when loading options from `~/.ideavimrc`, we initialise a hidden "fallback" window. This isn't a real IntelliJ `Editor`, and so it's local settings for "visual guides" is hard coded as an empty list. If there is a global value set in the IDE, this empty list looks like we've edited it, and this value gets copied to any new window, and hides the visual guides there. This fix will return the global value for the fallback window, and since the local value matches the global value, it's treated as default and doesn't update the new window.

Fixes [VIM-3984](https://youtrack.jetbrains.com/issue/VIM-3984/The-IDE-shows-visual-guides-incorrectly-with-IdeaVim-enabled)

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/group/OptionGroup.kt
+++ b/src/main/java/com/maddyhome/idea/vim/group/OptionGroup.kt
@@ -19,6 +19,7 @@ import com.intellij.openapi.editor.ScrollPositionCalculator
 import com.intellij.openapi.editor.ex.EditorEx
 import com.intellij.openapi.editor.ex.EditorSettingsExternalizable
 import com.intellij.openapi.editor.impl.softwrap.SoftWrapAppliancePlaces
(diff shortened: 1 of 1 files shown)


AlexPl292:
Looks good to me, but is this a specific behavior of this particular setting, or should this be moved to the general logic?

citizenmatt:
As far as I can tell, it's specific behaviour for this setting. The fallback window is an instance of `TextComponentEditorImpl`, so it's a simple text component that looks like an `Editor`. When it creates its settings object, it doesn't pass in an instance of `Editor` (itself), so the settings object thinks `editor` is null. For most cases, this means falling back to default or project settings, but for the margin guide, it just uses an empty list. I don't know why it doesn't use the default value like other options.

PR #1269, opened 2025-08-07
Fix command line missing due to empty colour

citizenmatt:
Fixes [VIM-3993](https://youtrack.jetbrains.com/issue/VIM-3993)


AlexPl292:
Thank you, merged.

PR #1274, opened 2025-08-19
Extend NERDTree support to Tree components other than Project

wxh06:
Related issue: [VIM-3253](https://youtrack.jetbrains.com/issue/VIM-3253/Support-j-k-to-move-in-all-list-related-UI)

### Supported operations

- [x] Navigation
- [x] ActivateNode
- [x] SpeedSearch (via #1280)

Currently can be enabled via `:set NERDTreeEverywhere`

### Limitations

Enabling extended NERDTree will make some mappings available only in Project NERDTree (e.g. `gi`, `go`) inoperable.

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/extension/nerdtree/NerdTree.kt
+++ b/src/main/java/com/maddyhome/idea/vim/extension/nerdtree/NerdTree.kt
@@ -21,6 +21,7 @@ import com.intellij.openapi.project.ProjectManager
 import com.intellij.openapi.startup.ProjectActivity
 import com.intellij.openapi.wm.ToolWindowId
 import com.intellij.openapi.wm.ex.ToolWindowManagerEx
+import com.intellij.ui.treeStructure.Tree
 import com.maddyhome.idea.vim.api.ExecutionContext
 import com.maddyhome.idea.vim.api.VimEditor
 import com.maddyhome.idea.vim.common.CommandAliasHandler
@@ -33,6 +34,7 @@ import com.maddyhome.idea.vim.key.MappingOwner
 import com.maddyhome.idea.vim.key.RequiredShortcut
 import com.maddyhome.idea.vim.newapi.ij
 import com.maddyhome.idea.vim.newapi.vim
+import java.awt.KeyboardFocusManager
 import java.util.concurrent.locks.ReentrantReadWriteLock
 import javax.swing.SwingConstants
 import kotlin.concurrent.read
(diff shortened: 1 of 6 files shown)


src/main/java/com/maddyhome/idea/vim/extension/nerdtree/NerdTreeEverywhere.kt
```diff
@@ -0,0 +1,33 @@
...
+internal class NerdTreeEverywhere {
+  @Service
+  class Dispatcher : AbstractDispatcher(mappings) {
+    init {
+      templatePresentation.isEnabledInModalContext = true
+
+      mappings.registerNavigationMappings()
+      mappings.register("NERDTreeMapActivateNode", "o", Mappings.Action { _, tree ->
+        val listener = tree.getActionForKeyStroke(KeyStroke.getKeyStroke(KeyEvent.VK_ENTER, 0))
+        listener.actionPerformed(ActionEvent(tree, ActionEvent.ACTION_PERFORMED, null))
```
AlexPl292, line 66: Is it possible to programmatically open the node? Action execution is a fragile thing, it's better to avoid it.

src/main/java/com/maddyhome/idea/vim/extension/nerdtree/NerdTree.kt
```diff
@@ -126,6 +128,23 @@ internal class NerdTree : VimExtension {
...
+    // NERDTree Everywhere
+    KeyboardFocusManager.getCurrentKeyboardFocusManager().addPropertyChangeListener("focusOwner") {
+      val newFocusOwner = it.newValue
+      val oldFocusOwner = it.oldValue
+      val dispatcher = service<NerdTreeEverywhere.Dispatcher>()
+      if (newFocusOwner is Tree) {
+        val shortcuts =
+          NerdTreeEverywhere.mappings.keyStrokes.map { RequiredShortcut(it, MappingOwner.Plugin.get(pluginName)) }
+        // It's okay to have the shortcut registered multiple times
+        dispatcher.registerCustomShortcutSet(KeyGroup.toShortcutSet(shortcuts), newFocusOwner)
```
AlexPl292: I don't think so. Generally, it's a bad practice to have multiple registrations of the same thing. I mean, if the new registration pops out the existing one, it's ok. But if they accumulate – it's better to avoid it.
This may work fine for now, but usually it ends in things like multiple execution of memory learks. Is it possible to pre-check if we have our registration there already?

AlexPl292 (commented):
Please see the comments

src/main/java/com/maddyhome/idea/vim/extension/nerdtree/NerdTree.kt
wxh06: I had checked the source code of `registerCustomShortcutSet`. It does skip the registration when the Action has already been registered.

<https://github.com/JetBrains/intellij-community/blob/idea/2025.2/platform/editor-ui-api/src/com/intellij/openapi/actionSystem/AnAction.java#L248-L251>

src/main/java/com/maddyhome/idea/vim/extension/nerdtree/NerdTreeEverywhere.kt
wxh06: Things get complex (or even impossible) to find a programmatically generic way of "opening a node" on trees with considerably different functionalities.

Taking Bookmark tree as an example, the `ActionListener` of <kbd>ENTER</kbd> is registered directly to the tree. Also, it will fallback to the `oldListenaer` obtained by using `getActionForKeyStroke`
<https://github.com/JetBrains/intellij-community/blob/idea/2025.2/platform/bookmarks/src/com/intellij/ide/bookmark/actions/extensions.kt#L121-L135>

Some trees, such as the navigation tree in Settings dialog, do not even have a mechanism to activate leaf node.
And the Problems tree has non-leaf nodes which are not expandable.

However, I am considering whether triggering a double-click event is more reliable than <kbd>ENTER</kbd>, which may be modified in the Keymap settings.

PR #1275, opened 2025-08-20
Fix a couple of bugs calculating inlay offsets

citizenmatt:
This PR fixes:

- [x] Up/down vertical motion tries to match the current column on the new line. It matches visual column rather than buffer column so that any inline or partial fold doesn't move the cursor too drastically. Inline inlay hints are represented as variable width visual columns and shouldn't be counted as part of the caret location. When inserting an inlay hint at an offset, it is always inserted into a visual column before the offset regardless of the "related to text" flag. So if there are any inlay hints at the caret offset, they should always be before the caret. 
    However, this relies on the caret being positioned on the visual column that represents the text character at the offset. When IdeaVim moves the caret, this is the case, but if IdeaVim does not move the caret, of if an inlay hint is added after the last move, this assumption is not necessarily valid - the caret might be positioned on an inlay hint, and the calculation to remove inlay hints is incorrect and could produce a negative number.
    This scenario can be seen by opening a new file with the caret on an empty line. The IDE might add an inlay hint to prompt for LLM code generation, and at this point, the caret is positioned at the wrong visual column, and the calculation produces a negative number, causes an exception and doesn't move the caret. This PR ensures the caret is correctly positioned before doing the calculation. See [VIM-4007](https://youtrack.jetbrains.com/issue/VIM-4007).
    This description is longer than the fix 😁 
- [x] After hitting `$` and then moving up/down onto a line with inlay hints, the caret is not placed at the end of the line. The new column is correctly calculated, but then normalised against the length of the visual line. Unfortunately, this calculation was completely wrong, and it would normalise against the length of the logical line. This logical line length obviously doesn't include inlay hints, meaning the normalised location was one or more characters less than the visual line length. See [VIM-3997](https://youtrack.jetbrains.com/issue/VIM-3997).

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/group/MotionGroup.kt
+++ b/src/main/java/com/maddyhome/idea/vim/group/MotionGroup.kt
@@ -27,7 +27,6 @@ import com.maddyhome.idea.vim.api.getLeadingCharacterOffset
 import com.maddyhome.idea.vim.api.getVisualLineCount
 import com.maddyhome.idea.vim.api.injector
 import com.maddyhome.idea.vim.api.lineLength
-import com.maddyhome.idea.vim.api.normalizeVisualColumn
 import com.maddyhome.idea.vim.api.normalizeVisualLine
 import com.maddyhome.idea.vim.api.visualLineToBufferLine
 import com.maddyhome.idea.vim.command.Argument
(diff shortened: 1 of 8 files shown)


AlexPl292:
Beautiful, thank you.

PR #1276, opened 2025-08-20
Indicate error when an unrecognized key sequence is entered in NERDTree

wxh06:
Make the user aware by invoking `VimMessages.indicateError` to beep when an unrecognized key sequence is entered in NERDTree.


AlexPl292:
Thank you!

PR #1279, opened 2025-08-20
Fix count accepted as valid in text object

citizenmatt:
This PR fixes command builder incorrectly accepting a count inside a text object, such as `di3w`. This is not a valid Vim command and should result in the delete command failing, and only the `w` succeeding.

Fixes [VIM-3960](https://youtrack.jetbrains.com/issue/VIM-3960)


AlexPl292:
Yep, thank you!

PR #1280, opened 2025-08-21
Improve SpeedSearch and disposal of NERDTree

wxh06:
- Initiate SpeedSearch right after <kbd>/</kbd> is pressed in NERDTree
- Clear NERDTree's KeyStroke buffer after <kbd>&lt;ESC&gt;</kbd> is pressed
- [VIM-3981](https://youtrack.jetbrains.com/issue/VIM-3981) make `:set noNERDTree` work
- Improve the concurrency of NERDTree mapping registration/disposal lifecycle
- Correct the behavior of <kbd>P</kbd> (`NERDTreeMapJumpRoot`) to jump to the root node

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/extension/nerdtree/AbstractDispatcher.kt
+++ b/src/main/java/com/maddyhome/idea/vim/extension/nerdtree/AbstractDispatcher.kt
@@ -25,7 +25,16 @@ internal abstract class AbstractDispatcher(private val mappings: Mappings) : Dum
   private val keys = mutableListOf<KeyStroke>()
 
   override fun update(e: AnActionEvent) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/maddyhome/idea/vim/extension/nerdtree/AbstractDispatcher.kt
```diff
@@ -83,3 +92,5 @@ internal abstract class AbstractDispatcher(private val mappings: Mappings) : Dum
     val LOG = vimLogger<AbstractDispatcher>()
   }
 }
+
+private const val ESCAPE_KEY_CODE = 27
```
AlexPl292: Please use KeyEvent.VK_ESCAPE instead of providing a custom constant

PR #1295, opened 2025-08-30
Implement c_CTRL-R commands

citizenmatt:
This PR implements various `c_CTRL-R` commands, to insert text into the command line. See [VIM-2511](https://youtrack.jetbrains.com/issue/VIM-2511).

* `<C-R><C-W>` to insert the word at or following the editor's caret
* `<C-R><C-A>` to insert the WORD at or following the editor's caret
* `<C-R><C-L>` to insert the current line
* `<C-R><C-F>` to insert the Filename at or following the editor's caret, as delimited by the `'isfname'` option
* `c_CTRL-R_CTRL-R` and `c_CTRL-R_CTRL-O` prefixes to insert the register/word/WORD/Filename literally (the other actions insert the contents as keystrokes, so `hello^Hworld` would appear on the command line as `hellworld`)

The key part of the PR is supporting commands that have another command as a prefix. That is, `c_CTRL-R {register}` is a command with a `<C-R>` mapping that then waits for a character before the action handler is executed. It is also a prefix to `c_CTRL-R_CTRL-W`. To support this, the order of the key consumers has changed, so that `CommandKeyConsumer` is processed before `CharArgumentConsumer`, so `<C-W>` is no longer treated as a character argument and matched as a command. The in-progress `InsertRegisterAction` is replaced with `InsertWordUnderCaretAction`, which has no arguments and is immediately invoked.

The PR has a lot of commits, sorry, but many of them are simple mechanical refactorings - renames, moving code, etc. A lot are extracting tests from `ExEntryTest` to a separate test class for each command line action, and then adding new test classes for the new actions.

Although it adds support for `'isfname'` (related to [VIM-4018](https://youtrack.jetbrains.com/issue/VIM-4018)) and `<C-R><C-F>` to insert the Filename under the cursor, it doesn't add support for `'path'` and `c_CTRL-R_CTRL-P`, as that adds file IO complexity.


citizenmatt:
Hmm. The `TestWithoutPrimaryClipboard` annotation doesn't work so well with macOS. It's checking to see if the `DISPLAY` environment variable is set, but that doesn't mean the current build is an X11 build. I've got some failing tests that don't run locally because of this.

citizenmatt:
Forgot to comment. Tests fixed and force pushed. Didn't do anything about the implementation of the `TestWithoutPrimaryClipboard` annotation. It doesn't run on macOS, but I can't see an easy way to change it.

AlexPl292:
Thank you for the PR!

PR #1296, opened 2025-08-30
[VIM-3541]Can't escape out of insert mode in read-only files

magidc:
# Bug description
## Steps to reproduce

- Use **Help | Find Action**, search for `"http request"`, hit **Enter**  
- Provide a name, hit **Enter**  
- Go into insert mode (type `i`)  
- Select **"Requests with Tests and Scripts"** from the **"Examples"** drop-down near the top right corner of the editor window  
- (Verify you're still in insert mode)  
- Hit **Esc** to exit insert mode  
- (Verify this action had no effect)  

---

### Expected result (correct behavior)

If you navigate to a read-only file while in insert mode, you should either:  
- be put into normal mode, **OR**  
- be allowed to press **Esc** to exit insert mode.  

---

### Actual result (faulty behavior)

If you navigate to a read-only file while in insert mode, you **remain in insert mode**.  
You cannot escape into normal mode. To get around this, you must:  

1. Return to an editable editor  
2. Escape into normal mode  
3. Then return to the read-only editor in order to use normal-mode commands.  

# Solution
I have opted for limiting the command actions in read-only files to the ones that do not change, or allow to change its contents. I have noticed that this bug was actually happening in more situations that the one described. What I did is a simply one-line change that blocks the execution of any command that extends from ChangeEditorActionHandler class when the editor is not writable.

### DIFF
--- a/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/KeyHandler.kt
+++ b/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/KeyHandler.kt
@@ -18,6 +18,7 @@ import com.maddyhome.idea.vim.command.OperatorArguments
 import com.maddyhome.idea.vim.diagnostic.VimLogger
 import com.maddyhome.idea.vim.diagnostic.trace
 import com.maddyhome.idea.vim.diagnostic.vimLogger
+import com.maddyhome.idea.vim.handler.ChangeEditorActionHandler
 import com.maddyhome.idea.vim.impl.state.toMappingMode
 import com.maddyhome.idea.vim.key.KeyConsumer
 import com.maddyhome.idea.vim.key.KeyStack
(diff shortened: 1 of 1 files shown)


AlexPl292:
Hello, thank you for your interest! There is an already onging a PR with the same issue: https://github.com/JetBrains/ideavim/pull/1283

Take a look at it, it has a long discussion. Also, if https://github.com/JetBrains/ideavim/pull/1283 will be finished, it will get a priority in merging.

magidc:
Thanks! I didn´t notice there was already an ongoing PR for the same issue. I will try to contribute to the discussion

magidc:
Just one question. I have developed some plugins on top of IdeaVim, one of them provides extra text objects like any kind of quoted text, list items or function arguments. 

There is already a plugin included in IdeaVIM to provide a text object for function arguments called "argtextobj". I was never comfortable with it because it fails to select arguments in some cases, specially when the argument is parametrized like "Map<String, Concept> arg". The plugin detects arguments doing a pure text analysis. My version seems to be more reliable as it is based on the PSIElements of the internal Intellij model.

I would like to contribute to IdeaVIM and fix the "argtextobj" plugin but the change would be a complete reimplementation. Should I do it this way or should I include as a new plugin?

Also, am I allowed to add my other IdeaVIM plugins to the [document where IdeaVIM plugins are listed](https://github.com/JetBrains/ideavim/wiki/IdeaVim-Plugins)

Thanks!

PR #1297, opened 2025-09-02
[Full IDE Integration] "Follow links" MVP

wxh06:
This pull request adds suppport for Vimium-like "following the link" with hint labels.
The default shortcut is <kbd>Alt</kbd> + <kbd>F</kbd>.

At present, only a simple hint generation algorithm has been implemented—specifically, letter permutation from the main row of the QWERTY keyboard.
Hints will remain unchanged between consecutive openings wherever possible.

### Implemented features

- Highlighting the clicked component
- Displaying only hints prefixed by the currently entered letters during input

### Current limitations

- A small pop-up needs to be visible and will appear at the center of the window in order to listen to user input; we might utilize this space by adding meaningful prompts

### Planned implementations

- Customizable keyboard layouts/alphabets for hints
- Generating hints based on
  1. components' depths within the accessible tree using an algorithm similar to Huffman coding
  2. the keyboard keys corresponding to a component's on-screen position


src/main/java/com/maddyhome/idea/vim/extension/hints/ToggleHintsAction.kt
```diff
@@ -0,0 +1,131 @@
...
+import java.awt.Component
+import java.awt.Container
+import javax.swing.AbstractButton
+import javax.swing.JLabel
+import javax.swing.JPanel
+import javax.swing.JTextArea
+import javax.swing.JTextField
+import javax.swing.SwingUtilities
+
+/**
```
AlexPl292: Although I appreciate good comments, there are cases where they add no value.
A simple way to check this is to see whether the comment introduces any words not already present in the class or method name. For example, if you have a class named ToggleHintsAction and the comment repeats the exact same words, it provides nothing new.
In such cases, I’d consider the comment unnecessary and remove it.

src/main/java/com/maddyhome/idea/vim/extension/hints/ToggleHintsAction.kt
```diff
@@ -0,0 +1,131 @@
...
+ * Toggle hints action
+ */
+class ToggleHintsAction : ToggleOptionAction(object : Option {
+  /** Whether the hints are enabled */
+  private var enabled = false
+
+  /** Cover panel container */
+  private val cover: JPanel = JPanel().apply {
```
AlexPl292: The goal of comments is not to have one on each function or field, but simplify the understanding. Here, I'd add also what "Cover" is.

src/main/java/com/maddyhome/idea/vim/extension/hints/ToggleHintsAction.kt
```diff
@@ -0,0 +1,131 @@
...
+    // no layout manager (absolute positioning)
+    layout = null
```
AlexPl292: `(absolute positioning)` is a good comment. It explains why have you decided not to use any layout manager.

src/main/java/com/maddyhome/idea/vim/extension/ShortcutDispatcher.kt
```diff
@@ -0,0 +1,112 @@
...
+   * getDefaultKeyStroke is needed for NEO layout keyboard VIM-987
+   * but we should cache the value because on the second call (isEnabled -> actionPerformed)
+   * the event is already consumed
+   *
+   * @author Alex Plate
+   */
+  private var keyStrokeCache: Pair<KeyEvent?, KeyStroke?> = null to null
+
+  /**
+   * @author Alex Plate
```
AlexPl292, line 94: Am I?... If this is some existing code, why not to reuse it?

src/main/java/com/maddyhome/idea/vim/extension/nerdtree/AbstractDispatcher.kt
```diff
@@ -11,26 +11,44 @@ package com.maddyhome.idea.vim.extension.nerdtree
...
 /**
  * Handles keyboard shortcuts and delegates them to appropriate actions.
  */
-internal abstract class AbstractDispatcher(private val mappings: Mappings) : DumbAwareAction() {
-  private val keys = mutableListOf<KeyStroke>()
+internal abstract class AbstractDispatcher(name: String, mappings: Map<List<KeyStroke>, NerdTreeAction>) :
+  ShortcutDispatcher<NerdTreeAction>(
+    name,
+    mappings,
+    object : Listener<NerdTreeAction> {
```
AlexPl292: Please extract a class instead of using an anonymous object in the constructor. This is confusing.

src/main/java/com/maddyhome/idea/vim/extension/ShortcutDispatcher.kt
wxh06: I moved the code from the dispatcher of NERDTree and it caused git blame to break, so I added `@author`

src/main/java/com/maddyhome/idea/vim/extension/hints/ToggleHintsAction.kt
wxh06: Thanks for pointing out. `cover` is the mask layer container for placing all hints; I just added it in the comments.

PR #1299, opened 2025-09-03
Fix casting error in `KeyStrokeTrie.getEntries`

wxh06:
Return an empty sequence if the prefix node is not found in `KeyStrokeTrie`.
Passing a non-existent prefix previously would result in a casting error.

I had intended to add unit tests for it, but `KeyStrokeTrie.add` uses `injector`, which isn't initialized in the tests for vim-engine.

### DIFF
--- a/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/key/KeyStrokeTrie.kt
+++ b/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/key/KeyStrokeTrie.kt
@@ -144,7 +144,7 @@ class KeyStrokeTrie<T>(private val name: String) {
       }
     }
 
-    val node = prefix?.let { getTrieNode(it) as TrieNodeImpl<T> } ?: root
+    val node = prefix?.let { (getTrieNode(it) ?: return emptySequence()) as TrieNodeImpl<T> } ?: root
     return sequence { yieldTrieNode(node) }
   }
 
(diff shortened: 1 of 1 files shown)


citizenmatt (changes requested):
Looks good to me, thanks. I wonder what your repro steps where?

I can make it fail with `:map zzz` - this should list any maps for `zzz`, and if they don't exist, it should print "No mappings found", but it hits this exception. It's a sneaky one because the command is executed as Vimscript, and `VimScriptExecutorBase.execute` catches all exceptions but only logs them as warnings. Perhaps this should be an error so we see the exception?

https://github.com/JetBrains/ideavim/blob/ad584d553c88278b183ee168abbd0f37eb333b28/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimScriptExecutorBase.kt#L70-L74

While it's hard to create a unit test, I think it would be possible to create a new test in `MapCommandTest` to check that the output of `map zzz` is "No mapping found" - which would require a small change to `KeyGroup.showKeyMappings`.

wxh06:
> I wonder what your repro steps where?

To be honest, I didn't find the problem through Vim commands.
I was using `KeyStrokeTrie` to construct a generic shortcut 'dispatcher' for use in UI components such as NERDTree.

<https://github.com/JetBrains/ideavim/blob/bc0cf025823734a10774e568c3ceadeb5a25f071/src/main/java/com/maddyhome/idea/vim/extension/ShortcutDispatcher.kt#L69>

My `actionPerformed` threw immediately after I entered a unrecognized sequence. Since it was not caught, it was easy to find the cause.

citizenmatt:
Right. We've been lucky not to hit it before. Most usage avoids passing a prefix, so we've got away with it for a while.

AlexPl292:
Thank you!

PR #1300, opened 2025-09-03
Refactor VimDataType

citizenmatt:
This PR continues the refactoring of the `VimDataType` API started in #1127. It deprecates the native accessor functions (`asString`, `asBoolean`, etc.) in favour of explicitly converting to the required datype and using the strongly typed accessor property to get the correct value. This makes the intent of the caller more explicit (e.g. the caller expects an expression to be a Number, and needs to treat that as a boolean value) and allows following Vim conversion rules more easily/obviously (only Number and String are automatically converted. Float isn't).

Specifically, the PR will:
- [x] Deprecate `toString` as a way to get a string representation of a Vim datatype. It can be implicitly called unexpectedly, it's not clear what the semantics of the resulting string are, and it doesn't respect Vim datatype conversions.
- [x] Introduce `toOutputString` as a more semantic replacement for `toString`. It will get a string representation of a Vim object for output purposes, such as error messages. All datatypes implement this function.
- [x] Introduce `toInsertableString` as another replacement for `toString`. It will return a string representation of a Vim datatype that can be inserted into the editor's text. It is mostly the same as `toOutputString`, but `VimList` overrides it to output each list item on a separate line.
- [x] Deprecate `asString` as a way to get the value of a String datatype. It is replaced with `toVimString().value` which makes it clear that the caller is expecting an expression to be a String, or convertible to a String. Only `VimNumber` is allowed to convert to a String. It is not removed because it is still in use by external plugins.
- [x] Deprecate `asBoolean` as a way to treat a Number as a boolean value. It is replaced with `toVimNumber().booleanValue`, which again makes it clear that the caller is expecting an expression to be a Number, or something that can be converted to a Number (i.e. String). The original implementation would incorrectly treat a Float as a boolean value. It is also used by external plugins.
- [x] Remove `asDouble`. It will get the value of a Float, but also incorrectly convert a Number or a String to a double value. It is not used externally, so is removed, and replaced with `toVimFloat().value`, becaus the caller is expecting a Float.


vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/model/functions/handlers/listFunctions/JoinFunctionHandler.kt
```diff
@@ -30,11 +30,18 @@ internal class JoinFunctionHandler : FunctionHandler() {
...
       throw exExceptionMessage("E1211", "1") // E1211: List required for argument 1
     }
-    val argument2 = argumentValues.getOrNull(1)?.evaluate(editor, context, vimContext)?.asString() ?: " "
-    return VimString(argument1.values.joinToString(argument2) { it.toString() })
+    val separator = argumentValues.getOrNull(1)?.evaluate(editor, context, vimContext)?.asString() ?: " "
+    return VimString(list.values.joinToString(separator) {
+      // According to the docs, String is used as-is, while List and Dictionary (and presumably the other datatypes) are
+      // converted in the same way as Vim's string() function, which has the same results as toEchoString()
+      when (it) {
+        is VimString -> it.value
```
AlexPl292: I'm confused. As I see `toOutputString()` returns `value` for the `VimString`. Why not to reuse this function for `VimString`?

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/model/functions/handlers/listFunctions/JoinFunctionHandler.kt
```diff
@@ -30,11 +30,18 @@ internal class JoinFunctionHandler : FunctionHandler() {
...
+    if (list !is VimList) {
       throw exExceptionMessage("E1211", "1") // E1211: List required for argument 1
     }
-    val argument2 = argumentValues.getOrNull(1)?.evaluate(editor, context, vimContext)?.asString() ?: " "
-    return VimString(argument1.values.joinToString(argument2) { it.toString() })
+    val separator = argumentValues.getOrNull(1)?.evaluate(editor, context, vimContext)?.asString() ?: " "
+    return VimString(list.values.joinToString(separator) {
+      // According to the docs, String is used as-is, while List and Dictionary (and presumably the other datatypes) are
+      // converted in the same way as Vim's string() function, which has the same results as toEchoString()
+      when (it) {
```
AlexPl292: Is `toEchoString` a real thing? I don't see that function in the code.

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/model/datatypes/VimDataType.kt
```diff
@@ -15,37 +15,50 @@ abstract class VimDataType {
...
-    throw NotImplementedError("implement me :(")
-  }
+  /**
+   * Returns a textual representation of the object, suitable for use in output messages
+   *
+   * Will return a textual representation of an object that isn't a String, such as a List, Dictionary or Funcref.
+   *
+   * Used by `:echo` and `:throw`, `join()`, formatting option values, as well as user facing error messages.
+   */
+  abstract fun toOutputString(): String
```
AlexPl292, line 57: I believe it would make sense to document some examples of how this fun will work for non-primitives like Lists, Dicts, or funcrefs. Here or in the function implementations.

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/model/functions/handlers/listFunctions/JoinFunctionHandler.kt
citizenmatt: Oops. Think that slipped through from an original round of renaming and refactorings

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/model/functions/handlers/listFunctions/JoinFunctionHandler.kt
citizenmatt: I think I started with the assumption that `toOutputString()` would be the same as Vim's `string()` function, which would return a String wrapped in single quotes. Then the implementation of `join()` would match the docs - we use our equivalent of `string()`, except for `VimString`.

But `toOutputString()` isn't the same as `string()` - for all other uses, we want to output the string value without any kind of quotes. So it's safe to use `toOutputString` for all datatypes here.

citizenmatt:
Thanks for the feedback. Updated!

AlexPl292:
Thank you!

PR #1307, opened 2025-09-08
Show message with :map and no mapping found

citizenmatt:
This PR adds tests for a scenario found in #1299. When typing `:map foo` and there were no mappings that began with `foo`, and before the change in #1299, IdeaVim would throw an exception instead of reporting "No mapping found".

While adding the tests for this scenario, it became apparent that `:map foo` will display any mappings that are a prefix of the `foo` arg, as well as any mappings that have `foo` as a prefix. So it will show a mapping that has a LHS of `f` or `food`. This PR updates the `KeyStrokeTrie` to handle this scenario, by optionally including nodes that have data (mappings) in the prefix as well as descendants of the prefix. It modifies the fix from #1299.

Note that this modifies `KeyStrokeTrie.getEntries` to add another default parameter. This function is called from external plugins, but without any parameters. AIUI, adding another default parameter is not a breaking change, but we should check this is true before release.


AlexPl292:
Thank you for the fix! I'll verify that the plugins maintain compatibility.

PR #1311, opened 2025-09-11
Restore API for external plugin compatibility

citizenmatt:
Restores the `KeyStrokeTrie.getEntries` overload with a single default parameter. This fixes the API breakage when adding a second default parameter.


AlexPl292:
Thank you!

PR #1313, opened 2025-09-12
Split resource bundles between host and engine, so IDE features work again

citizenmatt:
This PR splits the `IdeaVimBundle.properties` file between engine resources (mostly error messages) and host resources (mostly UI strings). This means that the `VimMessages.message` accessor can now be marked up with the `@PropertyKey` annotation to provide IDE functionality for properties - completion, validation, rename/find usages, folding and preview.

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/action/change/OperatorAction.kt
+++ b/src/main/java/com/maddyhome/idea/vim/action/change/OperatorAction.kt
@@ -10,7 +10,6 @@ package com.maddyhome.idea.vim.action.change
 import com.intellij.vim.annotations.CommandOrMotion
 import com.intellij.vim.annotations.Mode
 import com.maddyhome.idea.vim.KeyHandler
-import com.maddyhome.idea.vim.VimPlugin
 import com.maddyhome.idea.vim.api.ExecutionContext
 import com.maddyhome.idea.vim.api.VimCaret
 import com.maddyhome.idea.vim.api.VimEditor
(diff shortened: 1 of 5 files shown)


vim-engine/src/main/resources/messages/IdeaVimEngineBundle.properties
```diff
@@ -55,14 +56,15 @@ E70=E70: Empty {0}%[]
...
+E121=E121: Undefined variable: {0}
 E129=E129: Function name required
 E146=E146: Regular expressions can''t be delimited by letters
 E147=E147: Cannot do :global recursive with a range
 E148=E148: Regular expression missing from global
-e174.command.already.exists.add.to.replace.it=E174: Command already exists: add ! to replace it
-e176.invalid.number.of.arguments=E176: Invalid number of arguments
-e183.user.defined.commands.must.start.with.an.uppercase.letter=E183: User defined commands must start with an uppercase letter
-e184.no.such.user.defined.command.0=E184: No such user-defined command: {0}
+E174=E174: Command already exists: add ! to replace it
```
AlexPl292, line 66: BTW, I'd actually prefer to have some details of the error to be included in the key name. This is not useful for IJ, but when revieweing the code in GitHub, this simplifies understanding what the error is about. Something like at some moment YouTrack started to include the header of the issue in the link.

AlexPl292:
Thank you for your update! This required a cleanup at some point.

vim-engine/src/main/resources/messages/IdeaVimEngineBundle.properties
citizenmatt: Yeah, I wasn't sure what naming standard to go for, and just decided to keep it simple. If I need to add an error to code, it's easier to use the actual code that Vim prints out rather than figuring out that the `e_no_command` is the same code. I guess we could do another grand rename to match the naming style just replaced - `e176.invalid.number.of.arguments`?

AlexPl292: Yeah, this would be cool. It's sad to waste time renaming it back, though 😓

citizenmatt: Sooner or later I'll have a boring rainy evening or something...

PR #1322, opened 2025-09-23
Rework `:let` command implementation

citizenmatt:
This PR reworks the way `:let` is implemented, simplifying the implementation and fixing a number of edge cases.

Previosuly, `LetCommand` was a large `when` statement that would reimplement the `:let` assignment logic for each lvalue expression type (variable, register, option, sublist expression, indexed expression, ...). It's hard to get consistent behaviour across the implementations, and there were several bugs and edge cases around data type conversions and how the operator was applied for different data and expression types.

The `LetCommand` is simplified to have a single implementation of the core `:let` assignment functionality (e.g. `:let s[0:1]+=[1,2]`). It will get the current value of the lhs and rhs expressions, apply appropriate data conversions and calculate the new value, and then assign the new value to the lvalue expression.

It pushes this assignment out to a new interface method `LValueExpression.assign`, which is implemented by all expressions that can act as an lvalue (register, option, variable, sublist expression, etc.) This means the expression type implementation clas is now responsible for reading from and writing to the expression (and locking, which is important for subsequent work).

This PR also adds a load of extra tests to ensure that the behaviour matches Vim as best as possible, and to catch and fix a number of edge case bugs. It also adds the `..=` concatenation assignment operator, and implements the `string()` Vim function to help with testing.

This PR is the first step towards fixing some issues with `lockvar`.

### DIFF
--- a/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/model/commands/LetCommand.kt
+++ b/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/model/commands/LetCommand.kt
@@ -11,24 +11,16 @@ package com.maddyhome.idea.vim.vimscript.model.commands
 import com.intellij.vim.annotations.ExCommand
 import com.maddyhome.idea.vim.api.ExecutionContext
 import com.maddyhome.idea.vim.api.VimEditor
-import com.maddyhome.idea.vim.api.injector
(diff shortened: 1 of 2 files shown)


AlexPl292:
Wow, this is a big change! Merged!

citizenmatt:
Sorry! I do _try_ to make small PRs… 😅

PR #1330, opened 2025-10-02
Support hints for status bar widgets

wxh06:
The project switching and version control menus in the top-left corner remain unavailable due to accessibility issues (failure to implement the Accessible interface, specifically) with the ToolbarComboButton component in the IntelliJ platform.


AlexPl292:
Thank you!

PR #1332, opened 2025-10-08
Improve support for recording macros with code completion

chylex:
Three years after #528, I finally got tired of having my macros broken whenever code completion introduces an import, so I re-architected my original solution to fix that.

The new solution creates a temporary range marker at the start of the code completion, so any import or other code that appears before the marker will just nudge the marker along, and the macro will be updated from the correct offset.

The new solution also fixes an issue with completing a static member with a partial type name. For example, `WiE.A` would get incorrectly recorded as `WiEdowEvent.ACTION_EVENT_MASK`.

---

Commit message:

> Fixes wrong recorded inputs when code completion introduces an import.
>
> Fixes wrong recorded inputs when completing a static member with a partial type name. Example: `WiE.A` -> `WindowEvent.ACTION_EVENT_MASK`


chylex:
I'll check what's going on with the failing tests tomorrow.

chylex:
Actually, the 4 failing tests appear to be already failing in `master`.

AlexPl292:
Yes, this is a great fix. Thank you!

PR #1341, opened 2025-10-09
FEAT[VIM-268]: Completion for :e command

magidc:
Hi, this is my implementation for the requested feature of TAB completion on :edit and :write commands [VIM-268].

My solution modifies the handler for VIM command line `com/maddyhome/idea/vim/ui/ex/ExTextField.kt`. The handler updates the content in the command line when a character is input by the user, if not valid character it delegates the keystroke action. 
What I did was to intercept when TAB is input and try to add a completion if the command is supported. If the command is not supported (only works for :e, :edit, :w, :write) the execution follows the previous flow. If it is supported, the current argument is used to match potential file paths according the following logic:

- If the argument is the path of an existing directory: It completes with the first child
- If the argument is the path of an existing file: It completes with the next sibling according alphabetical order
- If the argument is an absolute path and does not corresponds with an existing file: It completes with the next sibling that has the same prefix
- If the argument is a relative path: It converts is to absolute using the project base path and completes with the next sibling that has the same prefix

This is the same behavior implemented in VIM or Neovim.

I have test it locally by running on the IDE but I have not developed unit test because I frankly don't know how I unit test this case. It is my first contribution to IdeaVim and I am not fully used to the codebase. I will be happy to receive your guidance and feedback.

Thanks!
Ricardo


magidc:
By the way, the code as it is can not be executed because an missing class in other part of the codebase:
```
[2025-10-09 23:29:34] [autobuild] e: file:///home/runner/work/ideavim/ideavim/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/regexp/VimRegex.kt:23:35 Unresolved reference 'VimSelectionModel'.
[2025-10-09 23:29:34] [autobuild] e: file:///home/runner/work/ideavim/ideavim/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/regexp/VimRegex.kt:699:5 'getSelectionModel' overrides nothing.
[2025-10-09 23:29:34] [autobuild] e: file:///home/runner/work/ideavim/ideavim/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/regexp/VimRegex.kt:699:39 Unresolved reference 'VimSelectionModel'.
```
Locally, I had commented this errors in order to be able to launch IdeaVIM into the IDE and make sure the new feature works as I expected. I did not include this commented part in this pull request as it is responsablity of other maintainer.

AlexPl292:
Hello, thank you for your interest! I've spotted some issues when looking at this change: when there is a folder, firstly, we suggest the folder path, and then we loop inside the folder elements.

Also, can you please take a look at the conflicts?

magidc:
Hi, thanks for the feedback,  I have made some improvements in order to behave better on input directory paths:

- If no prefix is added (`:e `) the suggestions will loop through all project base directory files and subdirectories 
- If prefix is added (`:e .i`), the suggestions will loop through the project base directory files and subdirectories that start by the given prefix
- If there is some change into the suggested entry (deleting or adding chars, move cursor, press arrow key), the next suggestion will take this new input as the prefix and it will loop according to it.
- If the prefix start by "/" (`:e /home/me/`), it will treated as a base path to loop through.
- If the prefix is the path of directory absolute or relative to base project path (:`e src/`), the suggestions will loop through it.

It is significant to notice that we can find to similar situations with different behaviors:

- If the command line shows something like `:e src/` because the **user manually typed it**, the suggestions will be looping through the `src` directory files and subdirectories
- If the command line shows something like `:e src/` **as result of looping suggestions**, the next suggestions will continue looping through the base path

This is the same behavior as in Vim editor.
Let me know what do you think.

Thanks  a lot

magidc:
Hey, I have reimplemented some parts to better support autocompletion for special cases like `:e .` or  `:e .`.
The implementation is based in a iterator that holds the full list of possible completions taking into consideration the search input provided by the user. It is possible to iterate from this list by pressing `TAB`.
It supports relative paths from project root and absolute paths. I have basically mirrored the behavior of **VIM** in every possible use case.

Since this pull requests is opened for months, would you mind to take a look @AlexPl292 ?
Thanks

AlexPl292:
Hi @magidc, thanks for the work on this — and apologies for the long silence after your January ping.

While this PR was waiting, we implemented this feature internally and merged it in April. I should have caught the overlap and flagged it sooner.

The merged version ended up with a broader scope (Shift-Tab + arrow cycling, `~/` expansion, case-insensitive matching, per-command completion-type registration via an extension point, plus command-name completion in a follow-up for VIM-258), and lives in the backend module for remote dev support.

Closing this one since the feature is in master. Hope you'll consider contributing again.

magidc:
No problem, I will contribute again for sure. I am already maintaining two popular plugins for IdeaVim so I am interested on keeping IdeaVIM in a good shape.
To be honest, I decided to fix this issue also because of the reward program. Would be possible to get a free subscription anyway?

AlexPl292:
@magidc The certificates I had were valid only in 2025, but let me see what I can do. Could you please ping me about that at aleksei.plate@jetbrains.com?

PR #1446, opened 2026-01-06
Fix(VIM-4104): Allow Tab to move to next component

citizenmatt:
When IdeaVim is not active, and the current editor is single line, or hosted in a dialog/tool window, or a readonly viewer (e.g. the Commit tool window editor), then hitting Tab will move focus to the next component rather than inserting a tab character. When IdeaVim is active, Tab does nothing (in Insert mode) because the `VimShortcutKeyAction` handles the keystroke, passing it through the key handler pipeline and then trying to invoke the (disabled) `TabAction` platform action to insert the tab character.

This PR fixes this behaviour by making `VimShortcutKeyAction` disabled if the above conditions are met. Since no actions are enabled, Swing will handle the keystroke. 

IdeaVim will not disable the action if the user has explicitly created an (Insert mode) map for Tab; if the user has explicitly set up a map, we should use it. The IdeaVim action is also only disabled in Insert mode, because Tab is a valid command in Normal. Admittedly, this command is navigating the jump list, and is unlikely in a single line or hosted editor. We could consider allowing this extra behaviour in Normal (or more) modes.

Fixes [VIM-4104](https://youtrack.jetbrains.com/issue/VIM-4104).

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/action/VimShortcutKeyAction.kt
+++ b/src/main/java/com/maddyhome/idea/vim/action/VimShortcutKeyAction.kt
@@ -19,6 +19,7 @@ import com.intellij.openapi.actionSystem.PlatformDataKeys
 import com.intellij.openapi.application.invokeLater
 import com.intellij.openapi.diagnostic.logger
 import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.editor.ex.EditorEx
 import com.intellij.openapi.editor.impl.EditorComponentImpl
 import com.intellij.openapi.progress.ProcessCanceledException
 import com.intellij.openapi.project.DumbAware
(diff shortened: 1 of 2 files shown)


src/main/java/com/maddyhome/idea/vim/action/VimShortcutKeyAction.kt
```diff
@@ -162,6 +165,24 @@ class VimShortcutKeyAction : AnAction(), DumbAware/*, LightEditCompatible*/ {
...
+    if (keyCode == KeyEvent.VK_TAB && editor.inInsertMode) {
+      // IdeaVim will handle Tab, and in Insert mode, will invoke the platform's TabAction to insert the tab character.
+      // (Other Tab features are handled by IDE actions called before VimShortcutKeyAction).
+      // Single line editors or editors hosted in a dialog/tool window do not accept/insert Tab, and TabAction is
+      // disabled. In fact, all actions are disabled and normal Swing handling will move the focus to the next
+      // component. Disable this action in the same circumstances, unless the user explicitly maps Tab.
+      // TODO: Should this also ignore Tab in Normal mode?
+      // This is a valid Vim action (navigating jump list), but unlikely in a single line or embedded editor.
+      if (injector.keyGroup.getKeyMapping(MappingMode.INSERT).get(listOf(keyStroke)) == null) {
+        if (editor.isOneLineMode || (editor as? EditorEx)?.isEmbeddedIntoDialogWrapper == true || editor.isViewer) {
```
1grzyb1: Just a nitpick maybe extract `injector.keyGroup.getKeyMapping(MappingMode.INSERT).get(listOf(keyStroke)) == null` to seperate method `hasInsertMapping` or something like that and join into second if so we won't need nested if statement.

Btw have you chacked how it behaves for other onle line editors?

citizenmatt: Yep, I'll get it updated, and I'll double check with other one-line editors

citizenmatt: Actually, I don't think we enable Vim features in any editors other than main and commit now, right?

citizenmatt: Added function to check for mappings. It's essentially the same as `maparg()`, so you can probably guess what my next PR will be 😁

citizenmatt:
@1grzyb1 @AlexPl292 Could we merge this PR, please? I have a few draft PRs queuing up behind it (see #1746 and friends)

PR #1466, opened 2026-01-12
Search fixes/improvements (show match count)

davidot:
## Content:
This PR fixes one small bug in highlighting and adds two new features.

- Fixes a case where highlighting does not match actual search.
  How to reproduce: Have a word with capital letters and one all lower case, use * on the capitalized word to search for that one. The search correctly ignores case, but if smartcase is enabled the highlighting incorrectly does not highlight the lower case word.
  Maybe related: [VIM-3459](https://youtrack.jetbrains.com/issue/VIM-3459) Regression: 'smartcase' should not affect star command

- Add an option to disable highlighting many matches during incsearch
  This otherwise can massively slow down searching as the IDE tries to highlight all 0s in a csv for example.
  This is protected behind an option ``, but for now I have set the default to 100, WOULD BE CHANGE IN DEFAULT BEHAVIOR (setting it to -1 gets old behavior)
  Related: 
  - [VIM-3581](https://youtrack.jetbrains.com/issue/VIM-3581) Disable search highlight outside editor to avoid search slowdown highlighted in other tool window (e.g. run/debug tool window)
  - [VIM-3093](https://youtrack.jetbrains.com/issue/VIM-3093) hlsearch+incsearch slow in large files

- Add option to show matches to search in file `showmatchcount` this will display a counter and index into the matches while searching, e.g. `[13/159]` means 159 matches and your at or before match 13.
  Related:
  - [VIM-2544](https://youtrack.jetbrains.com/issue/VIM-2544) [Feature Request]Show match count after searching
	 
Example:
image
Notice here that I'm searching for "0" in the left editor, the left file show matches because there are only 8 (see right part of command bar). The right file has 1000+ matches so shows nothing, meaning no hiccups while typing "01" for example.
Also visible in the middle bottom is the counter for matches in the files and the current index in the matches for that file.

Once you press enter the "protection" is gone, meaning to only a single slowdown in this case
image

After that I clicked on the right side and you can see what the search counter looks like when the entry is not active (maybe it should have a border on the left side?)

## Review
Please feel free to school me on anything Kotlin, this is my first time using the languages, just tried to match the patterns I saw in other files.

Also I really tried adding tests for option 2 & 3 but I had to mock so much I gave up, if there is a good/easy way to do it would love to hear it.

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/helper/SearchHighlightsHelper.kt
+++ b/src/main/java/com/maddyhome/idea/vim/helper/SearchHighlightsHelper.kt
@@ -107,6 +107,12 @@ private fun updateSearchHighlights(
       && (currentEditor == null || it.projectId == currentEditor.projectId)
   }
 
+  val shouldIgnoreCase = pattern == null || shouldIgnoreCase(pattern, shouldIgnoreSmartCase)
+
+  var maxhlduringincsearch = injector.globalOptions().maxhlduringincsearch
+  if (maxhlduringincsearch < 0)
+    maxhlduringincsearch = Int.MAX_VALUE
+
   editors.forEach {
     val editor = it.ij
     var currentMatchOffset = -1
@@ -136,13 +142,24 @@ private fun updateSearchHighlights(
             pattern,
             searchStartLine,
             searchEndLine,
-            shouldIgnoreCase(pattern, shouldIgnoreSmartCase)
+            shouldIgnoreCase
           )
         if (results.isNotEmpty()) {
-          if (editor === currentEditor?.ij) {
-            currentMatchOffset = findClosestMatch(results, initialOffset, count1, forwards)
+          // Only in incsearch is current editor not null, then check result size
+          val showHighlightsInEditor = currentEditor == null || results.size < maxhlduringincsearch
+          if (editor == currentEditor?.ij) {
+            val currentMatchIndex = findClosestMatch(results, initialOffset, count1, forwards)
+            currentMatchOffset = if (currentMatchIndex == -1) -1 else results[currentMatchIndex].startOffset
+
+            if (!showHighlightsInEditor) {
+              // Always highlight at least the "current" match in the active editor
+              highlightSearchResults(editor, pattern, listOf(results[currentMatchIndex]), currentMatchOffset)
+            }
           }
-          highlightSearchResults(editor, pattern, results, currentMatchOffset)
+
+          if (showHighlightsInEditor) {
+            highlightSearchResults(editor, pattern, results, currentMatchOffset)
+          }
         }
       }
       editor.vimLastSearch = pattern
@@ -237,7 +254,7 @@ private fun findClosestMatch(
     return -1
   }
 
-  return sortedResults[nextIndex % results.size].startOffset
+  return nextIndex % results.size
 }
 
 internal fun highlightSearchResults(
--- a/src/main/java/com/maddyhome/idea/vim/helper/SearchHighlightsHelper.kt
+++ b/src/main/java/com/maddyhome/idea/vim/helper/SearchHighlightsHelper.kt
@@ -27,6 +27,7 @@ import com.maddyhome.idea.vim.newapi.ij
 import com.maddyhome.idea.vim.newapi.vim
 import com.maddyhome.idea.vim.state.mode.inCommandLineModeWithVisual
 import com.maddyhome.idea.vim.state.mode.inVisualMode
+import com.maddyhome.idea.vim.ui.ex.ExEntryPanel
 import org.jetbrains.annotations.Contract
 import java.awt.Font
 import java.util.*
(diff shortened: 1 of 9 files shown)


src/test/java/org/jetbrains/plugins/ideavim/helper/SearchHelperTest.kt
```diff
@@ -90,6 +90,33 @@ class SearchHelperTest : VimTestCase() {
     kotlin.test.assertEquals(previousWordPosition, text.indexOf("second"))
   }
 
+
+  @TestWithoutNeovim(reason = SkipNeovimReason.NOT_VIM_TESTING)
+  @Test
+  fun testFindAllIgnoreCaseOverwritesSmartCase() {
```
1grzyb1, line 96: Can you also add a test in `SearchWholeWordForwardActionTest` for `*` action?

src/main/java/com/maddyhome/idea/vim/helper/SearchHighlightsHelper.kt
```diff
@@ -107,6 +107,12 @@ private fun updateSearchHighlights(
       && (currentEditor == null || it.projectId == currentEditor.projectId)
   }
 
+  val shouldIgnoreCase = pattern == null || shouldIgnoreCase(pattern, shouldIgnoreSmartCase)
+
+  var maxhlduringincsearch = injector.globalOptions().maxhlduringincsearch
```
1grzyb1, line 112: @AlexPl292 Please confirm we want to add this custom property.
`NeoVim` is using it like this:

```
- 'redrawtime' ('rdt', default 2000ms) — limits time spent on search highlighting. If highlighting all matches takes longer than this, highlighting is disabled for that pattern.
```

But his solution with just using count is much simpler imho

src/main/java/com/maddyhome/idea/vim/helper/SearchHighlightsHelper.kt
```diff
@@ -189,11 +202,44 @@ private fun updateSearchHighlights(
...
+    }
+
+    // If any of the following hold we are still searching:
+    // - We just highlighted some search results (countMatches is not MaybeClear
+    // - We did not clear highlights (
+    // - We are moving towards to a new position
+    if (countMatches == CountMatchesState.MaybeClearCount && clearHighlights && newCaretPosition == null) {
+      ExEntryPanel.getOrCreatePanelInstance().clearStatusText()
+      return@forEach
+    }
```
1grzyb1, line 231: Could you make this part about countMatches more self-explanatory?
It's hard to follow what `countMatches` is actually representing.
For example it could single method `shouldClearStatus` and inside `justHighlighted`, `highlightedNotCleared`, `MovingTowardsNextPositions`.

Also, not making `countMatches` mutable would help, as it's hard to follow what it actually represents

src/main/java/com/maddyhome/idea/vim/helper/SearchHighlightsHelper.kt
citizenmatt: I would also vote for using standard Vim options. I don't think IdeaVim should introduce its own (the custom ones we have start with "idea"). I think we end up in a good situation where googling a solution for a problem in Vim can also fix it in IdeaVim.

The same goes for the other option introduced in this PR.

davidot: Very good point it is good to be searchable.

So would it be okay if I implement the options like:
`set shortmess-=S` for showing search count (mention in docs that no other options are supported at the moment.

And `set redrawtime=200` -> But we don't actually do redraw time but use the count? Personally I would prefer to not freeze even for 2 seconds, unless I really want to search for something that occurs 1000s of times.
I had a quick try but stopping halfway (after 2000ms of adding highlights) feels very non intuitive to me.
Now you don't know if there are no other matches or it simply timed out.

Let me know your preference!

1grzyb1: We can pass the deadline to `injector.searchHelper.findAll`and check periodically if time elapsed every N matches found.
That would reproduce how it's implemented in nvim

AlexPl292:
Hi @davidot, gentle ping — it's been a while and I want to check in before this gets too stale.

Are you still interested in pushing this forward? If life got in the way, that's fine — let us know, and we could at least cherry-pick the smartcase fix on its own since it's independently valuable. If I don't hear back in ~2 weeks, I'll do that and close the rest.

Thanks again for the contribution!

AlexPl292:
Thanks again @davidot! I've cherry-picked the smartcase highlighting fix onto `master` (credited to you) — it'll ship in the next release and covers [VIM-3459](https://youtrack.jetbrains.com/issue/VIM-3459).

Closing the rest: per @citizenmatt's and @1grzyb1's review, we'd rather build the match-counter and incsearch limit on standard Vim mechanisms (`shortmess` `S`, a `redrawtime`-style budget) than custom options. If you'd like to rework them that way, a fresh PR would be very welcome. 🙏

PR #1534, opened 2026-01-27
VIM-4120 Display vim error messages in output panel

1grzyb1:
This PR introduces displaying vim error messages in output panel
To achive that I refactor `OutputPanel` architecture to support showing multiple lines in different styles so we can display errors in red alongside normal output messages.
I removed `ExOutputModel` with `ExOutputPanel` and replaced it with a single `OutputPanel`. This way, it's much simpler to understand what is going on. Also, there were a lot of strange ifs to unit test 
`isUnitTestMode` which were unnecessary actually and just making it more difficult to read.

https://github.com/user-attachments/assets/d5954925-af1b-4991-8974-72dd04616434


src/testFixtures/kotlin/org/jetbrains/plugins/ideavim/VimTestCase.kt
```diff
@@ -727,10 +727,10 @@ abstract class VimTestCase(private val defaultEditorText: String? = null) {
 
   fun assertCommandOutput(command: String, expected: String) {
     enterCommand(command)
-    assertExOutput(expected)
+    assertOutput(expected)
   }
 
-  fun assertExOutput(expected: String, clear: Boolean = true) {
+  fun assertOutput(expected: String, clear: Boolean = true) {
```
AlexPl292: TBH, I don't understand the reason for this rename. Now it got more "broad", is it intentional? Can you please specify the reason for the rename in the commit message and write the javadoc on those methods explaining what exact kind of the output it expected to be asserted.

PR #1617, opened 2026-03-20
Fix :edit to create file if it does not exist (VIM-266)

AndDe-gourav:
This change makes :edit behave like Vim by creating a new file
when the specified file does not exist.

Previously, :edit returned an error if the file was missing.

Changes:
- Create file inside project base path if not found
- Create parent directories when needed
- Refresh VirtualFileSystem after creation
- Open created file in editor

### DIFF
--- a/modules/ideavim-backend/src/main/java/com/maddyhome/idea/vim/group/file/FileRemoteApiImpl.kt
+++ b/modules/ideavim-backend/src/main/java/com/maddyhome/idea/vim/group/file/FileRemoteApiImpl.kt
@@ -12,6 +12,7 @@ import com.intellij.ide.vfs.VirtualFileId
 import com.intellij.ide.vfs.virtualFile
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.application.readAction
+import com.intellij.openapi.command.WriteCommandAction
 import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.diagnostic.debug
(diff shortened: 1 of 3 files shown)


1grzyb1:
Hi
I don't think we should create files using `java.io.File.createNewFile()` and `mkdirs()` as they are blocking I/O on the EDT.
There is `VfsUtil.createDirectoryIfMissing` built in in platform maybe try with this one.

AndDe-gourav:
I updated the code which uses the VfsUtil.createDirectoryIfMissing function.

1grzyb1:
Did you test that by running IdeaVim with split mode?
Also we should at least add standard tests for this behaviour and for split mode in `split-mode-tests`

AndDe-gourav:
I tested it but did'nt write tests,
should i write test cases for every single type of uses of command of :e

1grzyb1:
It would be useful to have at least one test for that, especially in split-mode tests, as it is easy to break stuff in split-mode scenarios

AndDe-gourav:
added normal and split mode tests

AndDe-gourav:
@1grzyb1 does all things are correct now.

1grzyb1 (approved):

1grzyb1:
@AndDe-gourav can you cleanup commit history? I mean we don't want to have merge commits, we prefer rebasing onto master

AndDe-gourav:
@1grzyb1 rebasing done.

PR #1695, opened 2026-04-17
VIM-1693 Implement autocmd

1grzyb1:
Implements Vim's `:autocmd` support, allowing users to register Ex commands to run on editor events like `BufRead`, `BufWrite`, `BufNewFile`, `FileType`, `InsertEnter/Leave`, `WinEnter/Leave`, `FocusGained/Lost`, and buffer enter/leave. Supports `augroup` blocks, multiple comma-separated events per command, and file/filetype pattern matching with Vim-style globs (*, **, ?, [abc], {foo,bar}). Includes grammar updates in Vimscript.g4, thread-safe listener handling, a new `AutoCmdService` API, extensive tests, and user-facing documentation in doc/autocmd.md.


src/test/java/org/jetbrains/plugins/ideavim/autocmd/AutoCmdTest.kt
```diff
@@ -0,0 +1,120 @@
...
+  @BeforeEach
+  override fun setUp(testInfo: TestInfo) {
+    super.setUp(testInfo)
+    configureByText("\n")
+    enterCommand("autocmd!")
+  }
+
+  @Test
+  fun `should execute command on InsertEnter`() {
+    enterCommand("autocmd InsertEnter * echo \"hi\"")
```
AlexPl292, line 30: Why changing that?

src/testFixtures/kotlin/org/jetbrains/plugins/ideavim/impl/TestInjector.kt
```diff
@@ -0,0 +1,50 @@
...
+import com.maddyhome.idea.vim.api.VimInjector
+import com.maddyhome.idea.vim.api.VimOptionGroup
+import com.maddyhome.idea.vim.group.IjVimOptionGroup
+import java.util.concurrent.atomic.AtomicBoolean
+
+/**
+ * Test injector that can wraps the original injector and add some functionality
+ *
+ * THIS IS NECESSARY to be careful with it!
+ * It's not supposed to be used to change the injector behaviour in tests. It only should cause some side effects
```
AlexPl292: Unfortunately, such messages bring no value but raise anxiety for the dev. Instead, it's better to write down the details of how exactly this class should be handled.

src/main/java/com/maddyhome/idea/vim/autocmd/AutoCmdImpl.kt
```diff
@@ -0,0 +1,60 @@
...
+  private var currentAugroup: String? = null
+
+  override fun registerEventCommand(command: String, event: AutoCmdEvent) {
+    eventHandlers.getOrPut(event) { mutableListOf() }.add(AuCommand(command, currentAugroup))
+  }
+
+  override fun clearEvents() {
+    if (currentAugroup != null) {
+      clearAuGroup(currentAugroup!!)
+      return
```
AlexPl292, line 31: Please try to avoid the code with `!!`. This raises anxiety in me :D

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/AutoCmdService.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import com.maddyhome.idea.vim.autocmd.AutoCmdEvent
+
+interface AutoCmdService {
+
+  fun handleEvent(event: AutoCmdEvent)
+  fun registerEventCommand(command: String, event: AutoCmdEvent)
+  fun clearEvents()
+  fun startAugroup(name: String)
+  fun endAuGroup()
```
AlexPl292: `Augroup` casing is inconsistent with the rest of functions

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/AutoCmdService.kt
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+package com.maddyhome.idea.vim.api
+
+import com.maddyhome.idea.vim.autocmd.AutoCmdEvent
+
+interface AutoCmdService {
+
+  fun handleEvent(event: AutoCmdEvent, filePath: String? = null)
+  fun registerEventCommand(command: String, event: AutoCmdEvent, pattern: String = "*")
```
AlexPl292, line 34: Please either use `Path` class or explain why the String should be here

src/main/java/com/maddyhome/idea/vim/autocmd/AutoCmdImpl.kt
```diff
@@ -0,0 +1,69 @@
...
+        if (handler.group == name) {
+          iterator.remove()
+        }
+      }
+    }
+  }
+
+  override fun handleEvent(event: AutoCmdEvent, filePath: String?, editor: VimEditor?) {
+    val editor = editor ?: injector.editorGroup.getFocusedEditor() ?: return
+    val path = filePath ?: editor.getPath()
```
AlexPl292: Please try using getSelectedEditor instead of getFocusedEditor. Focus is a part of "UI model" and there are always issues with that.

src/main/java/com/maddyhome/idea/vim/ui/ex/ExEntryPanel.kt
```diff
@@ -193,7 +194,8 @@ class ExEntryPanel private constructor() : JPanel(), VimCommandLine {
       }
       positionPanel()
       glassPane.isVisible = true
-      entry.requestFocusInWindow()
+      // Request focus after any pending output panel close (which uses invokeLater)
+      ApplicationManager.getApplication().invokeLater { entry.requestFocusInWindow() }
     }
```
AlexPl292, line 197: I hope, we'll remove this if we'll use getSelectedEditor

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/AutoCmdService.kt
```diff
@@ -0,0 +1,21 @@
...
+ * license that can be found in the LICENSE.txt file or at
+ * https://opensource.org/licenses/MIT.
+ */
+
+package com.maddyhome.idea.vim.api
+
+import com.maddyhome.idea.vim.autocmd.AutoCmdEvent
+
+interface AutoCmdService {
+
```
AlexPl292, line 24: Please put the docs for this service

src/main/java/com/maddyhome/idea/vim/autocmd/AutoCmdImpl.kt
```diff
@@ -0,0 +1,35 @@
...
+  override fun handleEvent(event: AutoCmdEvent) {
+    eventHandlers[event]?.forEach { executeCommand(it) }
+  }
+
+  private fun executeCommand(command: String) {
+    val editor = injector.editorGroup.getFocusedEditor() ?: return
+    val context = injector.executionContextManager.getEditorExecutionContext(editor)
+    injector.vimscriptExecutor.execute(command, editor, context, skipHistory = true)
+  }
+}
```
AlexPl292: There is a bunch of unnecessary changes in this PR

src/main/java/com/maddyhome/idea/vim/listener/VimListenerManager.kt
```diff
@@ -913,3 +923,19 @@ private object MouseEventsDataHolder {
   const val allowedSkippedDragEvents = 3
   var dragEventCount = allowedSkippedDragEvents
 }
+
+private class AutoCmdInsertEnterListener : ModeWillChangeListener {
+  override fun modeWillChange(editor: VimEditor, oldMode: Mode, newMode: Mode) {
+    if (oldMode != Mode.INSERT && newMode == Mode.INSERT) {
+      injector.autoCmd.handleEvent(AutoCmdEvent.InsertEnter)
```
AlexPl292: Is InsertEnter fired when working with replace mode? IIRC, in Vim it's almost the same.

AlexPl292 (changes requested):
Thank you!

src/test/java/org/jetbrains/plugins/ideavim/autocmd/AutoCmdTest.kt
1grzyb1: I had some issues with parsing " in autocmd in the past so it 's good to have echo with "" in test

src/main/java/com/maddyhome/idea/vim/autocmd/AutoCmdImpl.kt
1grzyb1: Oh right still some java muscle memory

src/main/java/com/maddyhome/idea/vim/listener/VimListenerManager.kt
1grzyb1: Nope added additional check and test to confirm it's working

PR #1746, opened 2026-05-05
Add mapping functions

citizenmatt:
This PR adds the mapping functions:

* `hasmapto()`
* `maparg()`
* `mapcheck()`
* `maplist()`
* `mapset()`

It fixes a couple of small bugs along the way: [VIM-4225](https://youtrack.jetbrains.com/issue/VIM-4225/Vim-sneak-maps-S-in-Op-pending-mode-instead-of-Z), which used the wrong mapping for vim-sneak in op-pending mode, and a bug in `:map` output where it would list all maps if there were no matching prefixes (e.g. typing `:map q23` will list all maps because there isn't a map that starts with `q23`).

~~This PR uses a function added as part of #1446, so is marked as draft until that is merged. Then I'll rebase and force push this one.~~


citizenmatt:
Rebased and force pushed. Ready for review @1grzyb1 @AlexPl292

vimscript-info/FUNCTIONS_INFO.MD
```diff
@@ -412,22 +453,26 @@ See `:help buffer-functions`, `:help window-functions` and `:help arg-functions`
...
+| ✅      | `getcmdtype()`        | return the current command-line type                                 |
+|        | `getcmdwintype()`     | return the current command-line window type                          |
+|        | `getcompletion()`     | list of command-line completion matches                              |
+|        | `getcompletiontype()` | get the type of the command-line completion for the specified string |
+|        | `fullcommand()`       | get full command name                                                |
+|        | `cmdcomplete_info()`  | get command-line completion information                              |
 
 ## Quickfix and location lists
 
+Note that IdeaVim does currently support the quickfix feature.
```
AlexPl292: Does or does not?

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimKeyGroup.kt
```diff
@@ -93,4 +95,94 @@ interface VimKeyGroup {
...
+
+      val originalModes = mappingInfo.originalModes
+      if (originalModes.size == 1) {
+        results.add(MappingInfoWithMode(mappingInfo, originalModes))
+      }
+      else {
+        entry.collectPath(fromKeysPool)
+        when (multiModeMappings[fromKeysPool]?.contains(originalModes)) {
+          null, false -> {
+            multiModeMappings.computeIfAbsent(fromKeysPool) { mutableSetOf() }.add(originalModes)
```
AlexPl292: `fromKeysPool` is mutable, we can't put it to the map

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimKeyGroup.kt
```diff
@@ -93,4 +95,94 @@ interface VimKeyGroup {
...
+      val entry = iterator.next()
+      val mappingInfo = entry.mappingInfo
+
+      val originalModes = mappingInfo.originalModes
+      if (originalModes.size == 1) {
+        results.add(MappingInfoWithMode(mappingInfo, originalModes))
+      }
+      else {
+        entry.collectPath(fromKeysPool)
+        when (multiModeMappings[fromKeysPool]?.contains(originalModes)) {
```
AlexPl292: That's an odd `!= true` :D

AlexPl292 (changes requested):
Thank you for the PR. Just a couple of notes.

vimscript-info/FUNCTIONS_INFO.MD
citizenmatt: Oops!

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimKeyGroup.kt
citizenmatt: Oh, yeah, great catch. That could have been a super subtle bug

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimKeyGroup.kt
citizenmatt: Ha ha! I'd forgotten how Kotlin handled comparing null to boolean and I just wanted to get the logic right...

citizenmatt:
@AlexPl292 All fixed. It was easier to fix the feedback in the original commits and force push

AlexPl292:
Thank you for the changes!

PR #1749, opened 2026-05-06
Add digraph Vim functions

citizenmatt:
This PR adds the digraph functions `digraph_get()`, `digraph_getlist()`, `digraph_set()` and `digraph_setlist()`.

It follows on from #1746, so contains those changes too. It's marked as draft until that PR is merged, then I'll rebase, force push and make ready for review.

PR #1775, opened 2026-05-15
Add history functions

citizenmatt:
This PR adds the history functions:

* `histadd()`
* `histdel()`
* `histget()`
* `histnr()`

It also does a little bit of refactoring and cleanup around the history implementation, encapsulating tracking the current history item in `HistoryBlock` rather than as part of the command line UI. It also fixes a couple of minor bugs related to navigation through the history, with additional tests.

~~It follows on from #1749, so contains those changes too. It's marked as draft until that PR is merged, then I'll rebase, force push and make ready for review.~~

### DIFF
--- a/src/main/java/com/maddyhome/idea/vim/group/HistoryGroup.java
+++ b/src/main/java/com/maddyhome/idea/vim/group/HistoryGroup.java
@@ -73,7 +73,7 @@ public class HistoryGroup extends VimHistoryBase
 
   private void readData(@NotNull Element element, String key) {
     final Type type = getTypeForString(key);
-    if (!isInitialised(type)) {
+    if (isInitialised(type)) {
       return;
     }
 
--- a/src/main/java/com/maddyhome/idea/vim/group/HistoryGroup.java
+++ b/src/main/java/com/maddyhome/idea/vim/group/HistoryGroup.java
@@ -13,7 +13,6 @@ import com.intellij.openapi.components.RoamingType;
 import com.intellij.openapi.components.State;
 import com.intellij.openapi.components.Storage;
 import com.intellij.openapi.diagnostic.Logger;
-import com.maddyhome.idea.vim.history.HistoryBlock;
 import com.maddyhome.idea.vim.history.HistoryEntry;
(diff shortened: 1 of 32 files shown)


citizenmatt:
@AlexPl292 Last one, ready for review!

src/main/java/com/maddyhome/idea/vim/group/HistoryGroup.java
```diff
@@ -74,21 +72,18 @@ public void readData(@NotNull Element element) {
   }
 
   private void readData(@NotNull Element element, String key) {
-    HistoryBlock block = getHistories().get(getTypeForString(key));
-    if (block != null) {
+    final Type type = getTypeForString(key);
+    if (!isInitialised(type)) {
```
AlexPl292: With `!` we will go into the code only if there IS an entry. However, it used to be that we go to further only if block is null, thus, no entry.

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/model/functions/handlers/historyFunctions/HistGetFunctionHandler.kt
```diff
@@ -0,0 +1,39 @@
...
+    vimContext: VimLContext,
+  ): VimString {
+    val history = arguments.getString(0).value
+    val index = arguments.getNumberOrNull(1)?.value
+
+    val historyType = VimHistory.Type.getTypeByString(history)
+      ?: injector.commandLine.getActiveCommandLine()?.historyType
+      ?: return VimString.EMPTY
+
+    val entries = injector.historyGroup.getEntries(historyType, index ?: 0, 0)
```
AlexPl292: If the index is missing, according to the documentation: `most recent item from the history is used`. But getEntries with both 0s will return ALL entries, instead of the latest one.

AlexPl292 (changes requested):
Hi, just two notes

citizenmatt:
Thanks, fixed!

AlexPl292:
Yay, thank you!

PR #1777, opened 2026-05-15
VIM-4234 Rollback to java 21

1grzyb1:
It turns out platform 2025.3 doesn't work with java 25


citizenmatt:
This is missing an update to `.idea/gradle.xml`, which is still trying to use `corretto-25`

PR #1781, opened 2026-05-15
Fix Incorrect behavior of C-u in insert mode (VIM-768)

AndDe-gourav:
now when using A in normal mode and then C-u do not move caret to wrong one past position


AlexPl292:
Thank you for the fix!

PR #1790, opened 2026-05-20
Add Abolish plugin

1grzyb1:
Partial implementaion of abolish plugin.
We support `:Subvret` method for subverting and searching using abolish coercions.
We cannot support now `:Abolish` as it requires implementing `:iabbrev` first.


citizenmatt:
The commits (and the changelog) have the wrong ticket - it should be [VIM-1326](https://youtrack.jetbrains.com/issue/VIM-1326), not [VIM-1362](https://youtrack.jetbrains.com/issue/VIM-1362).

PR #1809, opened 2026-05-28
Fix vim 315 incorrect insert normal mode behavior

AndDe-gourav:
while at the end of the line in insert mode pressing C-o , caret goes out of the end character leads to incorrect behavior, so i normalize it. and when returning to insert mode again restore the caret, so it match vim behavior when using C-o and then db from the end of the line

### DIFF
--- a/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/KeyHandler.kt
+++ b/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/KeyHandler.kt
@@ -260,6 +260,7 @@ class KeyHandler {
     val command = keyState.commandBuilder.buildCommand()
     val operatorArguments = OperatorArguments(command.rawCount, editorState.mode)
 
+    //true if goes form insert to OpPending
     val wasOpPendingFromInsert = (editor.mode as? Mode.OP_PENDING)?.let {
       it.returnTo is Mode.INSERT
     } ?: false
@@ -370,6 +371,7 @@ class KeyHandler {
       }
       injector.actionExecutor.executeVimAction(editor, cmd.action, context, operatorArguments)
 
+      // this restore the caret position when use C-o then db from the end of the line
       if (wasOpPendingFromInsert && editorState.mode is Mode.INSERT) {
         restoreAllowEndCaretPosition(editor, editorState)
       }
@@ -389,11 +391,7 @@ class KeyHandler {
       // mode commands. An exception is if this command should leave us in the temporary mode such as
       // "select register"
       if (editorState.mode is Mode.NORMAL && !cmd.flags.contains(CommandFlags.FLAG_EXPECT_MORE)) {
-        val oldMode = editorState.mode as Mode.NORMAL
-        editor.mode = oldMode.returnTo
-        if (oldMode.isInsertPending || oldMode.isReplacePending) {
-          restoreAllowEndCaretPosition(editor, editorState)
-        }
+        editor.mode = editorState.mode.returnTo
       }
 
       instance.reset(keyState, editorState.mode)
--- a/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/KeyHandler.kt
+++ b/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/KeyHandler.kt
@@ -370,16 +370,8 @@ class KeyHandler {
       }
       injector.actionExecutor.executeVimAction(editor, cmd.action, context, operatorArguments)
 
-      //C-o followed by db, this will move the caret one past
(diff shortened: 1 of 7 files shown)


1grzyb1:
Hi thanks for the PR!
Some tests are still failing can you take a look at that first?

AndDe-gourav:
@1grzyb1 i think this is done

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/KeyHandler.kt
```diff
@@ -259,6 +260,11 @@ class KeyHandler {
     val command = keyState.commandBuilder.buildCommand()
     val operatorArguments = OperatorArguments(command.rawCount, editorState.mode)
 
+    //true if goes form insert to OpPending
+    val wasOpPendingFromInsert = (editor.mode as? Mode.OP_PENDING)?.let {
+      it.returnTo is Mode.INSERT
+    } ?: false
```
1grzyb1: There is one issue with gating on OP_PENDING.
When we do `x` or D from `C-O` they are not using OP_PENDING so it won't fallback cursor to the end of line.

Try to do something like this instead:
```
    val isSingleCommandFromInsert = when (val m = editor.mode) {
      is Mode.NORMAL -> m.isInsertPending || m.isReplacePending
      is Mode.OP_PENDING -> m.returnTo is Mode.INSERT || m.returnTo is Mode.REPLACE
      else -> false
    }
```

And add tests for non op-pending cases

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/action/change/insert/InsertSingleCommandAction.kt
```diff
@@ -12,6 +12,7 @@ import com.intellij.vim.annotations.Mode
 import com.maddyhome.idea.vim.api.ExecutionContext
 import com.maddyhome.idea.vim.api.VimEditor
 import com.maddyhome.idea.vim.api.injector
+import com.maddyhome.idea.vim.api.normalizeOffset
```
1grzyb1: unused import

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/api/VimChangeGroupBase.kt
```diff
@@ -704,7 +704,22 @@ abstract class VimChangeGroupBase : VimChangeGroup {
    * @param editor The editor to put into NORMAL mode for one command
    */
   override fun processSingleCommand(editor: VimEditor) {
+
+    injector.vimState.wasCaretAtEndOfLineBeforeInsertNormal = editor.nativeCarets().any { caret ->
+      val line = caret.getBufferPosition().line
+      val lineEndAllowed = editor.getLineEndOffset(line, true)
+      val lineEndNotAllowed = editor.getLineEndOffset(line, false)
+      caret.offset == lineEndAllowed && lineEndAllowed != lineEndNotAllowed
```
1grzyb1: nitpick: extract this to a helper method `isCaretAtEnd` to improve readability

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/KeyHandler.kt
```diff
@@ -383,6 +396,19 @@ class KeyHandler {
...
+
+    private fun restoreAllowEndCaretPosition(editor: VimEditor, editorState: VimStateMachine) {
+      if (!editorState.wasCaretAtEndOfLineBeforeInsertNormal) return
+      for (caret in editor.nativeCarets()) {
+        val line = caret.getBufferPosition().line
+        val lineEndNotAllowed = editor.getLineEndOffset(line, false)
+        val lineEndAllowed = editor.getLineEndOffset(line, true)
+        if (caret.offset == lineEndNotAllowed && lineEndAllowed != lineEndNotAllowed) {
+          caret.moveToOffset(lineEndAllowed)
+        }
```
1grzyb1: Same logic as in `ChangeGroupBase` I assume we could extract this to same helper so we can do `caret.isAtEndLine()`

1grzyb1 (changes requested):

AndDe-gourav:
@1grzyb1 why some test fails when i run ideaVim tests in android studio even when they are correct

1grzyb1:
@AndDe-gourav Hard to tell. can you show me more details which tests are failing and how?

AndDe-gourav:
@1grzyb1 merge the PR if everything is correct and for my failing tests question i am opening a discussion

1grzyb1 (approved):

PR #1826, opened 2026-06-10
Update handling of output panel

citizenmatt:
This PR got a little large, sorry. However, each commit should be pretty small and self-contained.

This PR improves the handling of the output panel, mostly about how and when to dismiss the persistent "single-line" output panel that is shown without the "hit-enter" prompt and no clear way to dismiss it. 

Vim displays messages in the message area, which overlaps with the command line at the bottom of the screen. If the output is greater than the number of lines in the [`'cmdheight'` option](https://vimhelp.org/options.txt.html#%27cmdheight%27) (default 1), the message is displayed with the ["more-prompt" pager](https://vimhelp.org/message.txt.html#more-prompt), allowing scrolling, and ending up at the ["hit-enter" prompt](https://vimhelp.org/message.txt.html#hit-enter). If the message fits in `'cmdheight'`, it remains visible until the screen is redrawn.

IdeaVim doesn't have a permanent command line or message area. The command line is drawn as a popup text field at the bottom of the screen, and any messages are displayed in a similar scrollable popup text area, with more-prompt and hit-enter prompt. Until recently, single line messages were also displayed with a hit-enter prompt, but the current implementation treats a single line message like Vim - no hit-enter prompt and it remains visible. However, because it is a popup/overlay, it is more intrusive than Vim's (semi-)persistent message area, and it becomes distracting when it is still showing an error message that is no longer related to the most recent actions (see [VIM-4230](https://youtrack.jetbrains.com/issue/VIM-4230)).

With these changes, IdeaVim will:

- [x] Refactor and minor clean up off the output panel API and calling code
- [x] Add tests for the output panel's behaviour, both in single-line mode and hit-enter mode
- [x] Hide the single-line output panel when explicitly calling the redraw action (`<C-L>`)
- [x] Hide the single-line output panel when redrawing due to scrolling (e.g. `<C-E>` / `<C-Y>`, as well as caret motion)
- [x] Hide the single-line output panel with `<Esc>`. While this isn't Vim behaviour, it is an intuitive keystroke to try and dismiss something in the IDE, such as hints, popups and highlighting
- [x] Hide the single-line output panel when mode changes, like Vim
- [x] Hide the single-line output panel after a timeout. See below for more details
- [x] Prevent clearing the message at the same time as it's set. For example, we show the current search term when hitting `n`, but if the action shows a message and causes a scroll, it would be immediately removed, meaning sometimes we'd show the search term and sometimes we wouldn't
- [x] Fix the `'nomore'` scrolling implementation. If `'nomore'` is set to disable the more-prompt pager, IdeaVim would fail to scroll to the end of the output panel
- [x] Update the more-prompt message and implementation to match Vim. It now supports scrolling back as well as forward, and uses the same keys that Vim supports
- [x] Shortcut handling is bypassed when the more-prompt or hit-enter prompt is being shown. The output panel needs to be dismissed first (necessary to support e.g. `<C-C>`, `<C-F>`, `<C-B>` and `<C-H>` at the more-prompt)
- [x] The more-prompt and hit-enter prompts are also correctly updated when the output pane is scrolled with the mouse wheel
- [x] Lines are wrapped at character break rather than word break, which is very useful for long map output
- [x] Wrapped lines are included in scroll height calculations
- [x] The [`'cmdheight'` option](https://vimhelp.org/options.txt.html#%27cmdheight%27) is supported to set the height of the "single-line" output panel. Any messages that are less than or equal to the number of lines in `'cmdheight'` are shown in the "single-line" output panel
- [x] Supports [modeless selection](https://vimhelp.org/gui.txt.html#modeless-selection) with the mouse to highlight text in the output panel (and ex entry field) and uses [`<C-Y>` to copy to the clipboard](https://vimhelp.org/cmdline.txt.html#c_CTRL-Y)
- [x] Empty and blank text have special handling. If the single-line output is going to show an empty string, it's not visible (except in tests). This prevents showing a (persistent) empty output panel. If the text is several empty lines, it is only shown if it's greater than `'cmdheight'` - i.e. we only show it if it will appear in the more-prompt pager (see [VIM-4240](https://youtrack.jetbrains.com/issue/VIM-4240/Some-more-command-mode-odd-behaviours))
- [x] Typing `:` when showing the output panel will start a new command line, like Vim
- [x] External commands (e.g. `:!echo ''`) always use the hit-enter prompt, even if the output is empty, or less than `'cmdheight'`, and includes the typed command. This matches Vim behaviour, and avoids showing a (persistent) empty output panel, or showing nothing
- [x] Implement the [`g<` action](https://vimhelp.org/message.txt.html#g%3C) to show the last output

It also fixes some UI issues:

- [x] Check the height of the single-line output panel when scrolling, like we do with the command line height
- [x] Fix border insets, which fixes some layout and scrolling misalignment 
- [x] Add a small border to the content of the ex entry field and output panel, so text is not too close to the edge of the panel
- [x] Output scrolling is animated, like scrolling in the editor. It is also disabled when editor scrolling is disabled
- [x] The mouse cursor is properly shown while over the output panel (and ex entry field)

A couple of points of interest:

1. The single-line output panel is automatically hidden after a timeout. This timeout comes from the "wait" value in the [`'messagesopt'` option](https://vimhelp.org/options.txt.html#%27messagesopt%27). Vim uses the "wait" value to dismiss multiline output when the hit-enter prompt is disabled by removing the "hit-enter" flag from the option. IdeaVim extends this to also use the "wait" value to dismiss the single-line output. As such, IdeaVim includes "wait" in the default value (`messagesopt=hit-enter,history:500,wait:10000`), which Vim doesn't. The default value is set to `10000`, hiding the output after 10 seconds. It can be set to `0` to never hide the output (with hit-enter, `0` means to not wait)
    (The "history" value isn't implemented)
2. The `'nomore'` option in Vim disables the more-prompt pager. It automatically scrolls to the end of the output, and the only available action is then to dismiss the output. Since IdeaVim is a GUI application, we can also scroll the output with the mouse wheel. At this point, we've got a disabled pager, but scrolled content; we can scroll with the mouse but not the keyboard. The IdeaVim solution is for `'nomore'` to disable the pager when initially showing the output. It will automatically scroll to the end of the output, and then show the hit-enter prompt. At this point, the pager is available again. The user can scroll back with mouse or keyboard.
3. The "hit-enter" flag in `'messagesopt'` hides the hit-enter prompt at the end of output, and will dismiss the output after "wait" milliseconds. This is fairly horrible UX even in Vim - the user can do nothing until the timeout has elapsed. Again, because IdeaVim is a GUI app, we are a little more flexible. IdeaVim will dismiss the output after "wait" milliseconds, but during this time, the user can dismiss the output panel with `q` or `<Esc>`, or scroll back and abandon the timeout
4. The `set-commands.md` file has been updated and will need to be copied over to the wiki


1grzyb1 (approved):
Great work! Thanks!

PR #1871, opened 2026-06-25
Restore plugin compatibility for Argument.Type

citizenmatt:
The recent change to enable `'langmap'` support refactored argument types, including removing `Argument.Type.DIGRAPH`. Unfortunately, `DIGRAPH` is being used by [TheBlob42/idea-which-key](https://github.com/TheBlob42/idea-which-key) and I missed this break in compatibility. See TheBlob42/idea-which-key#113.

This PR adds back a static field to `Argument.Type` called `DIGRAPH`. It is set to the same value as `Argument.Type.CHARACTER` (which replaced `DIGRAPH`) and marked as deprecated.

Paging @1grzyb1 for a hopefully speedy point release. Apologies for missing this!


1grzyb1:
Cherry picked and released 2.40.1

citizenmatt:
Thank you!

PR #1889, opened 2026-07-03
Allow Setting Custom `.ideavimrc` Path

ameertaweel:
Currently, IdeaVIM looks for `.ideavimrc` in the user's home directory and in `$XDG_CONFIG_HOME`.

However, sometimes it's desirable to tell IdeaVIM to look for `.ideavimrc` in a specific custom path.

This change allows the user to specify a custom path in the `$IDEA_VIM_CUSTOM_VIMRC` environment variable.

My specific usecase for this:

I use the Nix package manager. Nix makes it easy to "wrap" programs along with their configuration. It basically creates a wrapper shell script that sets the needed options and environment variables then launches the software. I can already wrap IntelliJ by setting a custom `$XDG_CONFIG_HOME`, but then whenever anything in IntelliJ (including other plugins) looks for something in `$XDG_CONFIG_HOME`, it will look in the custom directory. This is a bit too much, I need a way to only change the location of the `.ideavimrc` file.

### DIFF
--- a/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/services/VimRcService.kt
+++ b/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/services/VimRcService.kt
@@ -66,6 +66,15 @@ object VimRcService {
   }
 
   private fun findIdeaVimRc(homeVimrcPaths: Array<String>, xdgVimrcPath: String): Path? {
+    val customVimrc = System.getenv("IDEA_VIM_CUSTOM_VIMRC")
+    if (!customVimrc.isNullOrEmpty()) {
+      val file = Path(customVimrc)
+      if (file.exists()) {
+        logger.debug { "Found ideavimrc file: $file" }
+        return file
+      }
+    }
+
     // Check whether file exists in home dir
     val homeDirName = System.getProperty("user.home")
     if (homeDirName != null) {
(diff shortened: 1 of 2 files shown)


vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/services/VimRcService.kt
```diff
@@ -66,6 +66,15 @@ object VimRcService {
   }
 
   private fun findIdeaVimRc(homeVimrcPaths: Array<String>, xdgVimrcPath: String): Path? {
+    val customVimrc = System.getenv("IDEA_VIM_CUSTOM_VIMRC")
+    if (customVimrc != null && customVimrc != "") {
```
1grzyb1: nit: You can use `isNullOrEmpty()`

1grzyb1 (changes requested):
It would be usefull to document this property in main README

vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/services/VimRcService.kt
ameertaweel: Done

ameertaweel:
Thanks for your review @1grzyb1. I updated the README to reflect the change. I also used `isNullOrEmpty()`.

1grzyb1 (approved):

PR #1903, opened 2026-07-09
fix Wrong 'motion last column action test' and 'D , d$ after i_Ctrl-o' behavior 

AndDe-gourav:
these behavior also require to refactor the caret position


1grzyb1:
Hi, thanks for the PR but some tests are failing now and there are conflicts with the master branch

AndDe-gourav:
@1grzyb1 it is done!

PR #1936, opened 2026-07-22
Fix vim 841 dollar motion does not include linebreak

AndDe-gourav:


AlexPl292 (changes requested):
Thanks — VIM-841 is real, but the condition is wrong.

Vim joins the next line only when `$` set `curswant = MAXCOL`. This patch triggers on "range ends at a line end":

```
v$x    → joins
v19lx  → no join   (same range)
```

`deletedToEndOfLine` can't tell those apart, so the join also fires on `d$`, `<C-W>` at EOL, `v$p`, and blockwise `D`.

Better place: extend the range past the newline when converting the charwise Visual selection to a `TextRange`, gated on `caret.vimLastColumn == LAST_COLUMN`. Fixes `d`/`c`/`y`/`p` together, leaves `d$` alone.

Also: Vim keeps the trailing newline in the register (`v$dp` still differs), and the test expects one `e` too many — real Vim gives `consectSed in orci mauris.`, cursor on `S`.

AndDe-gourav:
the fix for this is included in PR #2035

